PR #2, opened 2021-06-11
road/container bugs, terminal cost, extension size update on RCL

maconard:
Roads: can no longer be constructed on top of themselves
Containers: can no longer be constructed on top of themselves nor obstacles and construct with the correct hits
Terminals: correct construction cost
Extensions: correctly update capacity on change in RCL at levels 6/7/8

### DIFF
--- a/src/mods/controller/processor.ts
+++ b/src/mods/controller/processor.ts
@@ -18,7 +18,7 @@ export function claim(controller: StructureController, user: string) {
 }
 
 function updateExtensionCapacities(room: Room) {
-	// Extensions only change capacity a RCL 6/7/8
+	// Extensions only change capacity at RCL 6/7/8
 	if (room['#level'] > 5) {
 		const newCap = C.EXTENSION_ENERGY_CAPACITY[room['#level']];
 		for (const structure of room.find(C.FIND_STRUCTURES)) {
--- a/src/mods/controller/processor.ts
+++ b/src/mods/controller/processor.ts
@@ -1,3 +1,5 @@
+import type { Room } from 'xxscreeps/game/room';
+import type { Structure } from 'xxscreeps/mods/structure/structure';
 import * as C from 'xxscreeps/game/constants';
 import * as CreepLib from './creep';
 import { Game } from 'xxscreeps/game';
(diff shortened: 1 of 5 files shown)


maconard:
Obstacles can still be constructed on non-obstacles, such as containers. (Ex: extension auto-built on top of a container.)

I think we want a higher level solution for this one, maybe a change to checkPlacement in structure.ts, rather than fixing it in the checkPlacement local to each structure.

laverdet:
I appreciate you taking the time to look into these issues!

I took a peek at the rules for construction site placement in Screeps and actually the fix might be more simple:
https://github.com/screeps/engine/blob/78631905d975700d02786d9b666b9f97b1f6f8f9/src/utils.js#L173-L178

If I'm reading this correctly, you can't have any construction site on top of any other one. Not even a rampart and an extension at the same time. I would want to test this in the original engine, but if the behavior is confirmed we can just fix it with a simpler rule in checkCreateConstructionSite. And also a construction site may not be on top of a structure of the same `structureType`. That should fix the current big issues. Obstacle checking is going to be complicated to get right because we want to avoid special-casing between object types at all costs, which leads to the next point:

I think the extension energy updating mechanic needs to be rethought. The philosophy of mods is that each mod should be super general and avoid making assumptions about other game objects. Actually I've been declaring their dependencies upfront:
https://github.com/laverdet/xxscreeps/blob/main/src/mods/controller/index.ts#L4-L5

So based on this, a controller can only assume that Creep and Structure exists, it shouldn't assume anything about Extensions. This has two big advantages. The first is that obviously the code may be easier to reason about, and each mod ends up with less reliance on other mods. The second is more subtle though. By avoiding these kinds of special cases and generalizing the behavior in a mod-agnostic way we're strengthening the power of the API in the future.

The technique I had in mind for this is that `mods/controller` would define a method: `RoomObject.prototype['#roomLevelDidUpdate'](level: number, previousLevel: number | null)`.

When the room controller level changes the processor could iterate all objects and invoke that method. Then the extensions could just override ['#roomLevelDidUpdate'] to update themselves and the controller never has to know that extensions exist. This has the advantage that we can use the same method in the future to deactivate structures when the controller level drops, or implement whatever else room level affects.

Adding a property like this is a little tricky. You have to declare it [for TypeScript] and then define it [for JavaScript]. `combat.ts` kind of has an example of this: https://github.com/laverdet/xxscreeps/blob/main/src/mods/combat/creep.ts. There's also `harvestable` which defines ['#checkHarvest'] on RoomObject: https://github.com/laverdet/xxscreeps/blob/main/src/mods/harvestable/game.ts, but don't copy the `registerHarvestable` pattern here. Harvestable was the first mod developed so I didn't yet have the ['#semiPrivate'] pattern to build on and I was using a lot more annoying registration functions.

Thanks again for taking the time to submit this PR. If you want to keep working on this I don't think we'll be bumping into each other.. my active work right now is in the storage engine and pathing / terrain code.

maconard:
Thanks for the feedback! I had a feeling I wouldn't get your design intentions right on the first try :) I will work on this!

PR #3, opened 2021-06-13
Docker file and readme with example compose

n-kri:


n-kri:
build version from my branch https://hub.docker.com/r/nilskri/xxscreeps

laverdet:
Thank you for setting this up. Hopefully a Docker image will help users get xxscreeps running if they're having problems with nvm.

I think the build phase should be more like:
COPY package.json .
RUN npm install
COPY .npmrc bin src *.json .
RUN npm build

Then you'd be able to get rid of the unwieldy dockerignore file. Also it really speeds up the Docker build when the source is modified because you don't have to run npm each time. I'm not sure we need those paths in the gitignore either since they're kind of a user preference which could be set in the local exclude file.

Sorry for formatting, I'm on a phone but I wanted to get a response to you

n-kri:
After tinkering about a bit it seems that all files are required for the npm install stage
i tried it with different subset of those you mentioned above but i was unable to make it work i always resulted in errors
- without bit it complained on install that xxscreeps.js wasn't there
- after adding bin no complains during install but during runtime `/usr/bin/env: 'node\r': No such file or directory`
- after adding the /src folder pre npm install everything works

also it seemt that only he following paths are required for a successful build an run inside a container
/bin/**
/src/**
*.json

regarding the .dockerignore i simplified it a bit by inverting ii and only allowing the specific files required

laverdet:
Thanks!

PR #4, opened 2021-06-15
Container had invalid starting hitpoints.

BarryOEight:
Fixed it to start with the container constant amounts.


laverdet:
Thanks!

PR #5, opened 2021-06-18
Fixed Dockerfile

n-kri:
When using COPY with more than one source file, the destination must be a directory and end with a /


laverdet:
Spooky, it works for me on `Docker version 20.10.7, build f0df350`. Anyway sorry for breaking your build.

PR #8, opened 2021-07-30
Add friendly error message on run command that doesn't exist

samualtnorman:
I ran `npx xxscreeps export` instead of `import` and got a cryptic error message

a friendly error message would have saved me a little time, that's what this is

the message looks like this
```
➜  xxscreeps git:(main) ✗ npx xxscreeps foo
Invalid command or module "foo", built in commands are import, start, main, backend, processor, runner
```

the `import, start, main, backend, processor, runner` is not static, it grabs the keys from the object containing the command aliases to the module paths which I moved into a variable, so I could reference it multiple times

that variable is named `commands` but let me know if you want it changed to be more descriptive
I catch the error with a `.catch()` on the `import()` but let me know if you want that changed to a `try {} catch {}`

I also rethrow the error if the `error.code` is not `"ERR_MODULE_NOT_FOUND"` but let me know if there's something else preferred

thanks

### DIFF
--- a/src/config/entry.ts
+++ b/src/config/entry.ts
@@ -68,14 +68,25 @@ if (missingFlags.length) {
 	if (specifier && !specifier.startsWith('-')) {
 		// Run
 		const base = new URL('../..', import.meta.url);
-		await import(`${new URL({
+		const commands: Record<string, string | undefined> = {
 			import: './dist/scripts/scrape-world.js',
 			start: './dist/engine/service/launcher.js',
 			main: './dist/engine/service/main.js',
 			backend: './dist/backend/server.js',
 			processor: './dist/engine/service/processor.js',
 			runner: './dist/engine/service/runner.js',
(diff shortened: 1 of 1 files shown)


laverdet:
Thanks for the PR. I think this would be better as a two-step process, first with `import.resolve` wrapped in a try/catch to see if the import can be resolved. Then an unguarded `await import`, using the return value of the previous `import.resolve`. The v8 debugger tends to treat "caught" and "uncaught" exceptions differently so wrapping the entire program in a catch is undesirable.

samualtnorman:
wdym `import.resolve`?

laverdet:
Sorry, it's `import.meta.resolve`. It attempts to resolve a module and returns a either a `file:///` string or throws.

samualtnorman:
woops, it turns out I am bad at git
anyway, did you mean something more like this

(sorry about the history)

PR #34, opened 2021-08-26
[Test] added test for safe-mode scenario when 2 creeps can share/move…

Gadjung:
… to same room positon

sorry for formatting change :/


laverdet:
Actually I think we can improve this situation with some slight tweaks: 26151f65eac068bfac9c096e3c0c8b5ae0d82101

This fixes a bug which is still present in the vanilla implementation, where movements could be cancelled by a safe-mode obstruction. It also makes movement power more important than followers when it comes to who gets to move.

laverdet:
Thanks for the help!

PR #41, opened 2021-09-10
Steamless account creation

bastianh:


laverdet:
Thanks for the implementation! If you don't mind there's a few minor things that need to be updated:
- The email assignment should go through `context.authenticateForProvider` / `User.findProvidersForUser`. This allows the user to actually login via email instead of their account name. There's some sample code in the steam provider.
- I'd like to see a configure option to disable registration because I don't feel like dealing with a ton of fake accounts on xxworld

Let me know if you don't feel like updating the diff and I can take care of it after merging.

bastianh:
I have updated the branch, let me know what you think

laverdet:
Thank you!

PR #42, opened 2021-09-11
Added StructureObserver type

bastianh:
I have created a StructureObserver that works until the point where the intent is called in the processor. I have no idea how to get the target room visible for the user the next tick, maybe you have a hint ? 
I'll continue to look into it in the mean time .)

### DIFF
--- a/src/mods/observer/observer.ts
+++ b/src/mods/observer/observer.ts
@@ -12,12 +12,15 @@ export const format = () => compose(shape, StructureObserver);
 const shape = declare('Observer', struct(ownedStructureFormat, {
 	...variant('observer'),
 	hits: 'int32',
-	'#actionLog': RoomObject.actionLogFormat,
 }));
 
 export class StructureObserver extends withOverlay(OwnedStructure, shape) {
 	override get hitsMax() { return C.OBSERVER_HITS }
 	override get structureType() { return C.STRUCTURE_OBSERVER }
+
+	override isActive() {
+		return this.room.controller!.level === 8;
+	}
 
 	observeRoom(roomName: string) {
 		return chainIntentChecks(
@@ -52,7 +55,7 @@ registerBuildableStructure(C.STRUCTURE_OBSERVER, {
 function checkObserve(observer: StructureObserver, target: string) {
 	if (observer.room.controller!.level < 8) return C.ERR_RCL_NOT_ENOUGH;
 	const range = Game.map.getRoomLinearDistance(observer.room.name, target);
-	if (!(range < Infinity) || range === 0) {
+	if (!(range < Infinity)) {
 		return C.ERR_INVALID_ARGS;
 	} else if (range > C.OBSERVER_RANGE) {
 		return C.ERR_NOT_IN_RANGE;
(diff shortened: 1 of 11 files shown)


laverdet:
I sent this response via GitHub email yesterday but it doesn't seem like it went through. Copy/pasting below:
Thanks, it's exciting to see progress on this from the community. I haven't had a chance to look at the code yet but I can answer the question. I've actually thought about this in the past so I think I have a good idea. Basically you'll implement a new RoomObject which is invisible to the player and is responsible for providing vision. In the object's tick handler it will just remove itself from the room on the next tick.

You'll have to implement null LOOK and FIND constants for this, so that it's invisible to the player. The rest of it should be pretty easy to plug in to. Basically the object would provide vision, but not provide intents.

src/mods/observer/observer.ts
```diff
@@ -0,0 +1,59 @@
...
+	create(site) {
+		return create(site.pos, site['#user']);
+	},
+});
+
+//
+// Intent checks
+function checkObserveRange(observer: StructureObserver, target: string) {
+	const range = Game.map.getRoomLinearDistance(observer.room.name, target);
+	if (!(range < Infinity) || range === 0) {
```
laverdet, line 53: You can skip the negated comparison unless you're specifically dealing with NaN:
```
1/0 == 2/0
> true
```

laverdet:
Should `checkObserveRange` be called `checkObserveRoom`? Additionally you'd want a check of `() => checkMyStructure(this.spawn, StructureObserver),`. Otherwise the player could issue an observeRoom intent for non-observer structures.

src/mods/observer/observer.ts
bastianh: I think I got this check from the terminal structure and the `NaN` case happens if the user specifies an invalid room name, but I'll recheck.

bastianh:
> You'll have to implement null LOOK and FIND constants for this, so that it's invisible to the player. The rest of it should be pretty easy to plug in to. Basically the object would provide vision, but not provide intents.

wouldn't it make sense to not register it in the look index at all? from what I have seen in `lookAt` everything will be returned that is registered.
I guess I'm also doing something wrong with my format, since I get some errors I can't explain ;)

laverdet:
> > You'll have to implement null LOOK and FIND constants for this, so that it's invisible to the player. The rest of it should be pretty easy to plug in to. Basically the object would provide vision, but not provide intents.
> 
> wouldn't it make sense to not register it in the look index at all? from what I have seen in `lookAt` everything will be returned that is registered.
> I guess I'm also doing something wrong with my format, since I get some errors I can't explain ;)

Oh yeah definitely we don't want them in the look index. If I remember correctly the room code unconditionally adds all objects to LOOK and FIND so I was just saying you'd have to add a null check to skip that part.

bastianh:
I have updated the MR and according to the tests it's working .)

I had to add a few casts regarding the look types, not sure if that was the correct way to go

laverdet:
Thanks! Hopefully I can get this up and running on xxworld today

PR #43, opened 2021-09-12
Implemented MapVisuals

bastianh:


src/mods/visual/game.ts
```diff
@@ -37,3 +38,21 @@ extend(Room, {
...
+		 * circles, text labels).
+		 */
+		readonly visual: RoomVisual;
+	}
+}
+
+extend(GameMap, {
+	visual: {
+		get() {
+			return new RoomVisual('map');
```
laverdet, line 55: Is there a reason not to memoize the visual instance like `Room` does (above, line 34)

src/mods/visual/model.ts
```diff
@@ -24,7 +24,11 @@ export async function loadVisuals(shard: Shard, userId: string, roomName: string
 		return visualsString;
 	}
 	if (payload.time && typedArrayToString(payload.time) === `${shard.time}`) {
-		return stringify(payload['*']) + stringify(payload[roomName]);
+		if (roomName === 'map') {
```
laverdet, line 27: Can you modify the `hmget` above so that it doesn't fetch `*` in the case we're fetching `map`?

src/mods/visual/visual.ts
```diff
@@ -60,7 +64,9 @@ const rectSchema = struct({
 export type PolyStyle = Partial<TypeOf<typeof polySchema>['s']>;
 const polySchema = struct({
 	...variant('p'),
-	points: vector(array(2, 'double')),
+	// to allow room names for map visuals the points have to be saved in a less optimal way as objects.
```
laverdet, line 67: I'm not familiar enough with map visuals to know this, but would it not be possible to parse the room name as soon as the player gives it to us and store the result as raw coordinates? Basically when you add properties to a schema it increases the size of the data structure, even in the case of `optional`. Since these payloads are often pretty huge (around 500kb) I definitely don't want them to be any bigger than they have to be.

laverdet (changes requested):

src/mods/visual/visual.ts
bastianh: For the poly style there are two ways the coordinates can be send to the user ... as a simple array of coordinates like [[x,y],[x,y]...] or as room-positions [{x,y,n},{x,y,b},...]. I guess the second form was implemented when map visuals where added.
I now changed the database implementation to always use the second form because it works for room and for map visuals. 
I guess there is no easy way to allow both styles saved to the database? Maybe it makes sense to store it a different way and change the structure during the json serialization when it gets send over the socket?

src/mods/visual/game.ts
bastianh: I actually had it that way but I think it won't work because the map is not recreated every tick? I was getting the same RoomVisual Object and it grew with every tick.

src/mods/visual/visual.ts
laverdet: I took a quick peek at engine's implementation.. I see that they're storing `{ roomName, x, y}` but what we can do is something like this:
```js
const { rx, ry } = parseRoomName(roomName);
return { x: rx + x / 50, y: ry + y };
```

Then on the way to the client we can convert it into the format expected.

src/mods/visual/game.ts
bastianh: I did not really found a solution with that... it works with getting a new instance on every call... To optimize it we would need to memoize the instance, but get a new one on first call every tick.
I looked through the code and thought about creating one with the `gameInitializer` hook but not sure if that makes sense... or maybe store it in a different object but I did not really find out where to store it then.

src/mods/visual/visual.ts
bastianh: I have implemented the room storage in doubles for all the visual types... so no need to change any schemas

PR #45, opened 2021-11-17
Fix RoomVisual.poly() when called with a list of points in array format

bastianh:
from the documentation https://docs.screeps.com/api/#RoomVisual.poly

poly(points, [style])
===============

points : An array of points. Every item should be either an array with 2 numbers (i.e. [10,15]), or a RoomPosition object.


laverdet:
Thanks!

Issue #36, opened 2021-09-03
Server crash 1-2 ticks after start


Gadjung:
win10, Node 16.6.1, running via VS 2019 terminal
with and without default mods, tick speed does not change anything either:
```
mods:
  - xxscreeps/mods/classic
  - xxscreeps/mods/backend/cookie
  - xxscreeps/mods/backend/password
  - xxscreeps/mods/backend/steam
```
 
on clean copy of xxscreeps,
after
- npm install
- npm run build
- npx xxscreeps import

running `npx xxscreeps start` starts server, but after 1-3 long ticks (configured to 1s in config)
an error is thrown:
```
PS K:\xxscreeps_fork> npx xxscreeps start
🌎 Listening
Abandoning intents in rooms [W1N1, W1N9, W9N1, W9N9] for tick 1
Tick 1 ran in 5026ms; avg: 5025.55ms
Abandoning intents in rooms [W1N1, W1N9, W9N1, W9N9] for tick 2
Tick 2 ran in 5010ms; avg: 5010.11ms
node:internal/process/esm_loader:74
    internalBinding('errors').triggerUncaughtException(
                              ^
Error [ERR_UNHANDLED_ERROR]: Unhandled error. ("export 'install' (imported as 'SourceMap') was not found in 'source-map-support' (module has no exports)")
    at new NodeError (node:internal/errors:371:5)
    at Worker.emit (node:events:383:17)
    at Worker.[kOnErrorMessage] (node:internal/worker:298:10)
    at Worker.[kOnMessage] (node:internal/worker:309:37)
    at MessagePort.<anonymous> (node:internal/worker:210:57)
    at MessagePort.[nodejs.internal.kHybridDispatch] (node:internal/event_target:460:20)
    at MessagePort.exports.emitMessage (node:internal/per_context/messageport:23:28) {
  code: 'ERR_UNHANDLED_ERROR',
  context: "export 'install' (imported as 'SourceMap') was not found in 'source-map-support' (module has no exports)"
}
```

on subsequent tries of `npx xxscreeps start` the error pops immediately (no ticks processed) 
problem seems to steam from `source-map.ts` 

i tried installing it globally but nothing changed

issue started around 30 Aug so i tried some older commits `bae0f5ea00d557c49161a8fb9cc416d4c1a566f8` and `b104d94c85f7f8976a86568d1dfea4db1d0972d7 ` but nothing changed

laverdet:
[...]

PR #49, opened 2021-12-12
force git to load files with lf instead of crlf on windows

thmsndk:
I am on Windows and According to this https://eslint.org/docs/rules/linebreak-style#using-this-rule-with-version-control-systems
forcing LF should fix this, and it does look like it is fixed when I make a change to the file and discard it so it is loaded from git again.
image

Alternatively we could disable the rule?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.ts text eol=lf


laverdet:
I think what's happened here is you enabled `autoclrf` in your gitconfig, which I'd definitely recommend disabling since you're using VS Code. Anyway, `* -text` would be the correct fix in gitattributes if this is something we want to explicitly disable. The configuration you have there of course will fail on non-ts files.

thmsndk:
@laverdet it's been a while since i've set up my dev environment, but I'm pretty sure it's a default thing on windows, as per the rule description I [linked](https://eslint.org/docs/rules/linebreak-style#using-this-rule-with-version-control-systems)

> For example, the default behavior of git on Windows systems is to convert LF linebreaks to CRLF when checking out files, but to store the linebreaks as LF when committing a change. This will cause the linebreak-style rule to report errors if configured with the "unix" setting, because the files that ESLint sees will have CRLF linebreaks. If you use git, you may want to add a line to your .gitattributes file to prevent git from converting linebreaks in .js files:

they suggest to do this, I just modified it to .ts files.
```*.js text eol=lf```

I did however end up using a linux dev container with vscode instead due to not being able to start the server on windows as per #36 

Do we want to change it to `* text eol=lf` or just close this PR?

PR #53, opened 2021-12-14
Feature/power

thmsndk:
I intend to work on fletching out more power things
* moves `RESOURCE_OPS` into a `power` mod and registers it to `RESOURCES_ALL`
  * Fixes an issue with overmind when it reaches RCL 6 and expects `RESOURCE_OPS` to be part of `RESOURCES_ALL`

### DIFF
--- a/src/config/mods/import/constants.ts
+++ b/src/config/mods/import/constants.ts
@@ -20,3 +20,4 @@ export * from 'xxscreeps/mods/road/constants';
 export * from 'xxscreeps/mods/source/constants';
 export * from 'xxscreeps/mods/spawn/constants';
 export * from 'xxscreeps/mods/structure/constants';
+export * from 'xxscreeps/mods/power/constants';
--- a/src/game/constants/resource.ts
+++ b/src/game/constants/resource.ts
@@ -5,7 +5,7 @@ export const RESOURCE_METAL = 'metal' as const;
 export const RESOURCE_BIOMASS = 'biomass' as const;
 export const RESOURCE_MIST = 'mist' as const;
 
-export const RESOURCE_OPS = 'ops' as const;
+/*export const RESOURCE_OPS = 'ops' as const;*/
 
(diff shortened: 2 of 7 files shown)


laverdet:
Power is probably the hardest feature to work on because the effects of power creeps infiltrate every entity in the game in very particular ways. I think the "final boss" of xxscreeps mods is to publish an external mod which implements the Operator.

thmsndk:
@laverdet That is a very good point about operator touching many things. While it is a great exercise, making a core feature an "external" mod might give some issues? But I definitely think doing an external mod would make sense. should the external mods be some of the earlier season mods instead perhaps?

Do we want to merge this fix in, or perhaps fix it in another way?

laverdet:
Well this patch on its own doesn't actually do anything. I wouldn't mind Power living in core xxscreeps but given the scope of work that needs to be done for power creeps I've always planned for those to live outside the core repository. That's why I consider them the "final boss" of the mods system, because of how many features they end up touching.

Also, personally, I find power creeps to be a not very good feature so I'm not even excited about it.

thmsndk:
@laverdet What this patch does is add/register `RESOURCE_OPS` to `RESOURCES_ALL` overmind expects it to be in there, if it aint there the code crashes because it ends up receiving undefined instead of actual data.

Is there another way we could add it instead of with the power mod structure? the reason I did it this way was because I saw the chemistry mod doing it, and I felt that was flexible e.g. this part, and I also felt that the constant should live in a mod in the spirit of the mod architecture, cause mods needs to be able to extend all aspects of the core gameplay.

```typescript
const resourceSchema = registerEnumerated('ResourceType', ...resources);
C.RESOURCES_ALL.push(...resources);
```

I mean I'm fine with power creeps being an external mod, I haven no idea where to even start writing something as an external mod though, lot's of advanced concepts happening in xxscreeps, at least for me :)

I am just trying to contribute where I can, and perhaps give an overview of the roadmap so others can contribute as well

PR #62, opened 2023-04-29
Bring repository into working order

TylerSeanRau:
This commit brings this repo back into functioning order previously `npm install` would fail for a number of reasons two of which were related to c++17 not being used and another for updates to v8 (which updating nan solves).

For this to make this repo function
https://github.com/laverdet/ivm-inspect/pull/1 also needs to be accepted.


laverdet:
Thanks for taking a look at this. It looks like nodejs v20.x now requires c++17, which is fine, but will require a semver major update on isolated-vm. There's also some other complications with isolated-vm on newer versions of nodejs [> v19.x] which need to be resolved. I'll try to take a closer look at this soon to get these repositories working on newer versions.

TylerSeanRau:
@laverdet no prob! I've been wanting to use this repo again for a while now and figured I'd try to get it working.

I found that with some tweaks I could get through all the install steps but ultimately I gave up because `npx xxscreeps start` would throw errors from isolated-vm that I didn't know how to solve. (and they looked a lot less like configuration problems :))

The teaks I tried were:

* With nodejs v20.0.0 I got stuck when I found that I couldn't use `--experimental-specifier-resolution=node`, people say the fix for this is to write an experimental-loader but the documentation for [experimental loader](https://nodejs.org/api/esm.html#loaders) doesn't give any insight on how to make them. If someone could direct me to some docs for it then I'd give it a shot.
    * resolution: go down to nodejs LTS 18.16.0
* at some point I encounted `import jsonSchema from './badge.schema.json';` from `dist/engine/db/user/badge.js` needing to be changed to `import jsonSchema from './badge.schema.json' assert {"type":"json"};`, I believe I manually was fixing this in both node v20.0.0 and v 18.16.0
* with node v18.16.0 I encountered issues with packages during `npx xxscreeps start` as mentioned above

To get things working for the time being I installed node v16.13.2 and everything functions perfectly

TylerSeanRau:
@laverdet I don't want this particular repo to become a burden for you so feel free to describe what needs to be done here and I'll try to take it from there.  (If you could also note external dependencies I'd need to wait for that'd be great)

PR #69, opened 2024-05-27
Remove MOVE from drop intent check

### LINKED ISSUE #68: Fix drop requiring MOVE body part
Do you read this the same way I do: drop shouldn't require MOVE body part?

xxscreeps: `https://github.com/laverdet/xxscreeps/blob/main/src/mods/creep/creep.ts#L470`

screeps: `https://github.com/screeps/engine/blob/master/src/game/creeps.js#L404`

Change:
```ts
-         () => checkCommon(creep, C.MOVE),
+         () => checkCommon(creep),
```

Will try to raise a PR :).

blbu:
`MOVE` should not be required to successfully `drop`.

Fixes #68


laverdet:
Good catch!

PR #79, opened 2026-03-28
feature: Add CLI mod with admin endpoint for server management

misterwise:
## Summary

Adds an initial CLI mod that provides admin/management endpoints for xxscreeps, equivalent to the traditional Screeps private server CLI. This is an early implementation covering the core sandbox and foundational helpers — submitted now to get feedback on style and approach before building out the more significant features (room mutations, map/bot helpers, etc.).

Evaluates JavaScript in a per-request VM sandbox with access to the database, admin helpers, and server controls.

- `GET /cli` — greeting/health check
- `POST /cli` — execute JavaScript expression, return result as text

## What's implemented

- **Sandbox objects:** `print`, `console.log/warn/error`, `db`, `shard`, `storage.db`, `storage.shard`, `storage.pubsub`, `users.findByName/info`, `rooms.list/load`, `system.*`, `help()`
- **System controls:** `getTickDuration`, `setTickDuration`, `pauseSimulation`, `resumeSimulation`, `sendServerMessage`, `resetAllData` (stub)
- **Top-level `await`** via async IIFE wrapping — not available in the traditional CLI
- **5-second execution timeout** — not present in the traditional CLI
- **24 tests** covering expression evaluation, output capture, error handling, top-level await, sandbox isolation, all helpers, and edge cases
- **Tracking doc** at `docs/cli-mod.md` with full feature surface comparison

## What remains

- `rooms.save(name, room)` — serialize and write room state back to storage (needed for live room mutations)
- `system.save()` — flush DB/shard after mutations
- Room object inspection helpers
- `map.*`, `bots.*` helpers (deferred — require engine-level integration)
- Verify against a running server (tested via unit tests only so far)

## Intentional divergences from the traditional CLI

| Area | Traditional | This mod | Reason |
|------|------------|----------|--------|
| Port | Separate port (port+1) | Same port as game server | xxscreeps uses a single Koa server with hooks; a second port adds config/firewall surface for no benefit |
| Greeting | `GET /greeting` | `GET /cli` | Consistency — both endpoints share the `/cli` path prefix |
| Sandbox lifetime | Per-connection persistent (backend-local) / per-request (screeps-launcher) | Per-request (fresh) | HTTP has no connection state; matches screeps-launcher. Variables do not persist between commands |
| `storage.db` | MongoDB collections API | KeyValProvider (Redis-like) | xxscreeps uses binary keyval, not document collections. No compatible API without a full query translation layer. Thin helpers (`users`, `rooms`) cover common operations |
| `storage.env` | Redis client | Not provided | xxscreeps has no separate env store; equivalent data accessible via `shard` and `db` |
| Top-level `await` | Not supported | Supported | Strict improvement — traditional CLI required `.then()` chains for async operations |
| Timeout | None | 5 seconds | Strict improvement — prevents `while(true){}` from hanging the server |

## Design questions

**Core engine modifications for pause/resume.** This mod adds `pauseSimulation` and `resumeSimulation` message types directly to the service channel's Message union in `src/engine/service/index.ts`, and adds listener/pause logic to the main game loop in `src/engine/service/main.ts`. These are the only changes outside `src/mods/backend/cli/`.

An alternative would be a hook system for extending service channel message types, allowing mods to register custom messages without editing core types. Open to guidance on the preferred approach.


laverdet (commented):
Structurally I'm not sure an ACE (arbitrary code evaluation) endpoint is the way to go about this. Even if you did some kind of localhost-host only check it's still very sus and insecure even for a video game.

There are bones of a CLI in the primary launcher `src/config/entry.ts`. It's called "repl" here. With something like this `xxscreeps start` would function as a cli / repl.

src/mods/backend/cli/sandbox.ts
```diff
@@ -0,0 +1,239 @@
...
+
+	// Execute
+	const vmOptions = { filename: 'cli', timeout: 5000 };
+	try {
+		let result;
+		if (expression.includes('await')) {
+			// Top-level await: wrap in async IIFE
+			let fn;
+			try {
+				fn = vm.runInContext(`(async()=>(${expression}))`, sandbox, vmOptions);
```
laverdet: This still fails on: `doSomeMutation(); JSON.parse("}");`. It will cause `doSomeMutation()` to run twice. Drop the expression into a function and then invoke the function:
```ts
const AsyncFunction = (async function(){}).constructor;
const userFunction = function() {
  try {
    new AsyncFunction(`return (${code});`);
  } catch {
    new AsyncFunction(`${code}`);
  }
}();
userFunction(whatever);
```

misterwise:
> The redis storage provider is here by the way https://github.com/laverdet/xxscreeps-redis
> 
> With an actual storage provider you can run main loop, processor(s), runner(s), and backend(s) on different machines.

This repo is actually private it appears.

PR #84, opened 2026-03-28
fix: store capacity rendering for empty SingleStore and RestrictedStore

misterwise:
This became apparent when a room went from RCL7 to RCL8, the existing extensions failed to render properly when empty showing 0/0 energy and displaying oddly. Implemented the fix to mirror official behaviour wherever possible.

## Summary
- Add `#capacityEntries()` method to Store hierarchy so `renderStore` emits `storeCapacityResource` independently of store contents, fixing extensions (and other SingleStore/RestrictedStore users) showing 0/0 capacity when empty (e.g. after RCL upgrade)
- Add renderStore tests for empty and populated Open, Single, and Restricted stores

## Test plan
- `npm test` — all Store tests pass (6 new renderStore tests), existing tests unaffected

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -27,6 +27,7 @@ const javaScriptRules = {
 	eqeqeq: [ 'warn', 'smart' ],
 	'func-name-matching': 'warn',
 	'id-denylist': [ 'warn', 'xxx', 'foo', 'bar' ],
+	'id-length': [ 'warn' ],
 	'logical-assignment-operators': [ 'warn', 'always', { enforceForIfStatements: true } ],
 	'no-array-constructor': 'warn',
 	'no-caller': 'warn',
--- a/src/mods/chemistry/store.ts
+++ b/src/mods/chemistry/store.ts
@@ -28,6 +28,16 @@ export class LabStore extends withOverlay(Store, shape) {
 		}
 	}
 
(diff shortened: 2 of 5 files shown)


laverdet (commented):
It took me a little while to follow along. It seems like "Single store with resources renders capacity" is the main case here that is changed.

The result of the new `store['#capacityEntries']` function is always immediately passed to `Fn.fromEntries`, which makes me think it's a bit too vague. `store['#entries']` returns an entries iterable because usually we always want to walk over it. Really, the main reason `#entries` exists at all is for better TypeScript types, since otherwise it's just a wrapper around `Object.entries`. Anyway, in the capacity case, we never want to walk over it so the entry pairs don't actually matter.

I think what might be better is a new `#storeCapacityResource` function which just returns whatever is needed to the renderer and to that pretty print util function.

Additionally, there are three implementations of `#capacityEntries` (which would become `#storeCapacityResource`), and there also three concrete Store classes. That means that actually the base Store method should be abstract and each concrete class should override. With the way it's structured now the base `Store` implementation is only used in `OpenStore`.

This isn't super clear from my code but `RestrictedStore` corresponds to Nuker & maybe PowerSpawn. The tests were mostly conceptualized from here https://github.com/screeps/engine/blob/master/spec/engine/game/storeSpec.js.

PR #90, opened 2026-03-30
fix: tombstones persist forever in idle rooms after creep death

### LINKED ISSUE #58: [Bug] Tombstones with negative decay are not destroyed
These images of Source Keeper Tombstones with negative decay were taken in room E4S6 of [xxworld](https://xxworld.net).

image
image
image
image

Unfortunately, not really sure how or why this happened, but, likely its as simple as an `=== 0` instead of a `<= 0` in Tombstone tick logic.

misterwise:
## Summary
- Fixes #58 — tombstones with negative decay accumulating in source keeper rooms
- When a creep dies, `buryCreep()` inserts a tombstone via deferred `#insertObject`. The Tick loop captures the array length before iterating, so the tombstone's Tick processor never runs on the death tick — no `wakeAt(decayTime)` is registered. If nothing else schedules a room wake, the room sleeps indefinitely and the tombstone is never cleaned up
- In source keeper rooms this also prevents the keeper lair from detecting the missing keeper, so keepers stop respawning
- This isn't an issue in the official engine because rooms process every tick — there's no sleep/wake system, so the tombstone is always visible to the next tick's iteration loop
- The fix changes `context.didUpdate()` to `context.setActive()` at both `buryCreep` call sites, scheduling the room to process next tick

## Verification
- Added 5 tests covering keeper age death lifecycle, tombstone decay cleanup, keeper lair respawn detection, and a first-tick death edge case
- The first-tick death tests fail without the fix and pass with it
- Full existing test suite passes (exit 0)


laverdet (commented):
Can the test be simplified a lot? I don't think we should be mucking about in the scratch storage in the tests since that's really internal.

Does a keeper lair change anything about the simulation, or can this just test the tombstone in isolation?

Comments like "(it returns before line 317 on death)" aren't helpful because that will be out of date very fast.

PR #92, opened 2026-03-30
fix: construction sites should not block creep movement

### LINKED ISSUE #50: [Bug] [Pathfinding] unbuilt csites (ramparts) seem to block movement
I have an overmind bot where creeps are getting stuck trying to path trough a csite. 
As soon as I remove the csite where their orange path indicates it wants to path trough, they start moving.
image
image

misterwise:
## Summary

Construction sites of obstacle types (spawn, extension, wall, etc.) incorrectly block creep movement. In official Screeps, construction sites never block movement — they only influence pathfinding cost matrices.

The root cause is that the refactored obstacle checker system uses `makeObstacleChecker()` for both pathfinding and movement resolution, with no way to distinguish the two contexts. The construction site checker unconditionally treats the owner's obstacle-type sites as obstacles, but should only do so during pathfinding.

The original design (visible in the commented-out code in `obstacle.ts`) had a `pathing` flag for exactly this distinction — construction sites were only obstacles when `pathing` was true.

## Changes

- Add `pathing` flag to `MovementParameters` (defaults to `undefined`, so all existing callers are unaffected)
- Gate the construction site obstacle checker on `params.pathing` — returns `null` when not pathfinding
- Pass `pathing: true` only from the pathfinding cost matrix builder in `roomSearch()`

This also subsumes the earlier rampart-specific fix (f92550c), which changed rampart's `obstacle: undefined` to `false` as a narrower workaround for the same underlying issue.

Closes #50

## Verification

- Added `move()` and `moveTo()` tests for creep movement through a spawn construction site (obstacle type)
- Tests confirmed to fail without the fix (creep blocked at x=24) and pass with it (creep reaches x=25)
- All existing tests pass, including rampart csite movement tests from #51

### DIFF
--- a/src/game/path-finder/index.ts
+++ b/src/game/path-finder/index.ts
@@ -69,6 +69,7 @@ export function roomSearch(origin: RoomPosition, goals: RoomPosition[], options:
 				const check = makeObstacleChecker({
 					ignoreCreeps,
 					ignoreDestructibleStructures,
+					pathing: true,
 					room,
 					user: me,
 				});
--- a/src/game/path-finder/obstacle.ts
+++ b/src/game/path-finder/obstacle.ts
@@ -7,6 +7,7 @@ type MovementParameters = {
 	checkTerrain?: boolean | undefined;
 	ignoreCreeps?: boolean | undefined;
 	ignoreDestructibleStructures?: boolean | undefined;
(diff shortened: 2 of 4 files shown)


misterwise:
You're 100% right, thank you for correcting me.

PR #97, opened 2026-03-30
Fix map visual rendering, implement getSize and size limits, add MapVisual class

### LINKED ISSUE #17: [Bug] Missing Game.map.visual functions
trying to call any of the https://docs.screeps.com/api/#Game-map-visual
throws error that .visual is undefined

misterwise:
Closes #17

## Summary
Went in to look at the bugfix for the issue, but discovered opportunity to make visuals feature complete (as far as I can tell) without too much extra work. Also love that Claude decided to use the term "footgun" a new one for me.

- **Fix map visual coordinate decode**: The `loadVisuals` decode switch only handled lines and polys — circles, rects, and text fell through the `default` branch and were sent to the client with raw world coordinates instead of room-relative coordinates with room names. Added the three missing cases.
- **Implement `getSize()`**: Was a stub returning 0. Now tracks cumulative serialized bytes via a per-room `RoomVisualState` shared across all `RoomVisual` instances for the same room.
- **Enforce size limits**: 500 KB per room, 1000 KB for map — matching the official engine. Throws with the same error messages as official when exceeded.
- **Add `MapVisual` class**: Separate class registered as a global, returned by `Game.map.visual`. Methods require `RoomPosition`-style arguments (with `roomName`), providing compile-time safety against the previous footgun where `Game.map.visual.circle(25, 25)` would silently produce corrupt world coordinates. Matches the official API's separate `MapVisual` type.
- **Extract `decodeMapVisuals` from `loadVisuals`**: The decode logic was a closure inside `loadVisuals` that required Redis to test. Extracted into a standalone exported function for testability. `loadVisuals` is simplified — it calls `decodeMapVisuals` instead of inlining the switch.

## Why this wasn't caught earlier

PR #43 added map visual support by reusing `RoomVisual` with `roomName='map'`. This triggers an encode/decode cycle through world coordinates that the official engine doesn't use — official stores `{x, y, n: roomName}` directly without encoding. The decode step only handled lines and polys, silently passing other types through with raw world coordinates.

## Architectural note

The official engine's map visuals store room-relative coordinates with room names inline — no encoding. xxscreeps encodes to world coordinates on the game side, serializes to a binary schema, then decodes back on the backend. This round-trip is unnecessary but harmless now that all five visual types are decoded correctly. Eliminating it would require changes to the shared binary schema and is left as a future simplification.

## Verification

- 19 tests across 4 groups: coordinate decode (5), getSize (7), size limits (4), MapVisual class (3)
- Tests were written first against the buggy code and confirmed to fail before fixes were applied
- Full existing test suite passes with no regressions


src/mods/visual/visual.ts
```diff
@@ -237,21 +258,84 @@ export class RoomVisual {
...
+ * Map visuals provide a way to show various visual debug info on the game map. You can use the
+ * `MapVisual` object to draw simple shapes that are visible only to you.
+ *
+ * Map visuals are not stored in the database, their only purpose is to display something in your
+ * browser. All drawings will persist for one tick and will disappear if not updated. All
+ * `MapVisual` API calls have no added CPU cost (their cost is natural and mostly related to simple
+ * `JSON.serialize` calls). However, there is a usage limit: you cannot post more than 1000 KB of
+ * serialized data (see `getSize` method).
+ */
+export class MapVisual {
```
laverdet: This is the craziest pants-on-head abstraction I've seen in a while.

Try:
```ts
class VisualOf<Point extends unknown[]> {
	#description;

	constructor(description: string) {
		// use this in your error message
		this.#description = description;
	}

	line(...args: [ ...pos1: Point, ...pos2: Point, style?: LineStyle ]) {
		// ...
	}
}

export class RoomVisual extends VisualOf<[ x: number, y: number ] | [ pos: LocalPoint ] | [ pos: RoomPoint ]> {
	constructor(roomName: string) {
		super(`RoomVisual in room ${roomName}`);
	}
}

export class MapVisual extends VisualOf<[ pos: RoomPoint ]> {
	constructor() {
		super('MapVisual');
	}
}
```

laverdet (changes requested):

src/mods/visual/visual.ts
misterwise: I'm a bit taken aback it was possible to write that so badly. I think I've tidied up the inheritance cleanly now.

PR #101, opened 2026-03-31
fix: enforce MAX_CONSTRUCTION_SITES limit

misterwise:
## Summary
- `MAX_CONSTRUCTION_SITES` (100) was defined as a constant but never enforced
- Add global site count check in `Room.createConstructionSite` using `userGame.constructionSites`, returning `ERR_FULL` when the limit is reached
- Check is in the runner sandbox only, matching official engine behavior

## Verification
- Test creates 100 construction sites, verifies all succeed
- Test verifies the 101st returns `ERR_FULL`
- Test removes a site and verifies creation succeeds again at 99
- All existing tests continue to pass

### DIFF
--- a/src/mods/construction/test.ts
+++ b/src/mods/construction/test.ts
@@ -21,9 +21,10 @@ describe('Construction', () => {
 		});
 	}));
 	test('max construction sites', () => construction(async ({ player, tick }) => {
-		// Place 90 sites in tick 1
+		// Place most sites in tick 1
+		const firstBatch = C.MAX_CONSTRUCTION_SITES - 10;
 		await player('100', Game => {
-			for (const pos of Fn.range(90)) {
+			for (const pos of Fn.range(firstBatch)) {
 				const xx = 1 + (pos % 48);
 				const yy = 1 + Math.floor(pos / 48);
 				assert.strictEqual(Game.rooms.W1N1.createConstructionSite(xx, yy, 'road'), C.OK);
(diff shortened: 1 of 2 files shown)


src/mods/construction/test.ts
```diff
@@ -18,6 +19,34 @@ describe('Construction', () => {
 			assert(Object.values(Game.constructionSites).length === 1);
 		});
 	}));
+	test('max construction sites', () => construction(async ({ player, tick }) => {
+		// Place MAX_CONSTRUCTION_SITES roads in one tick
+		await player('100', Game => {
+			for (let ii = 0; ii < C.MAX_CONSTRUCTION_SITES; ++ii) {
+				const x = 1 + (ii % 48);
```
laverdet: Identifiers should always be longer than one length. Try `pos = whatever` or `xx = 123`

laverdet (changes requested):
This one doesn't work. `Game` is frozen in a tick and doesn't reflect construction sites which will be added next tick.

The test should include one where the player issues like 90 in one tick and then 20 in the next tick. Then you can remove one and try again like it's there now.

Vanilla Screeps just uses a global counter in the player runtime (see `createdConstructionSites`) which works fine I guess. My threat model assumes that players can issue any arbitrary intents, which is why I don't have that same client-side check. But, since the processor works per room there is no way to enforce this shard-level constraint. I suppose Vanilla has the same problem given a theoretical intents-breakage.

I'd say add a similar check in the player runtime with a comment noting that it is not enforced correctly in the processor.

PR #109, opened 2026-03-31
feature: Implement TOUGH body part damage reduction for boosted creeps

misterwise:
## Summary

Implements the missing TOUGH boost damage reduction, noted as a known gap in #82.

Boosted TOUGH parts (GO ×0.7, GHO2 ×0.5, XGHO2 ×0.3) now reduce incoming damage. The reduction is calculated by walking body parts front-to-back: each boosted TOUGH part absorbs `part.hits / multiplier` incoming damage while only losing `absorbed * multiplier` HP. Overflow past exhausted TOUGH parts hits remaining parts at 1:1. Reduction applies to gross incoming damage independently of same-tick healing.

### Changes

- **`src/mods/creep/creep.ts`** — Add `tickDamageTaken` field to track raw incoming damage separately from net `tickHitsDelta`; export `BoostsLookup` type
- **`src/mods/creep/processor.ts`** — `calculateEffectiveDamage()` function + tick processor integration with guard for zero-net-delta edge case
- **`src/mods/combat/test.ts`** — 8 tests covering all three boost tiers, ranged attack, same-tick healing, overflow past destroyed TOUGH, and balanced damage/healing edge case
- **`src/mods/chemistry/test.ts`** — Export `createLabWithResources` helper for reuse

### Known limitations

- Floating point imprecision possible in extreme cases (large damage with ×0.7 or ×0.3 multipliers) — matches JavaScript behavior in the official engine

## Verification

- 8 new tests, all confirmed to fail before implementation and pass after
- Zero regressions across existing test suite


laverdet (changes requested):

misterwise:
Force push to rebase and bring in fix from upstream/main regarding duplicate player(100) in test without tick in between.

PR #112, opened 2026-04-03
docs: Add architecture overview for new contributors

misterwise:
I have been working on consolidating my understanding of how xxscreeps fits together architecturally. I didn't actually have a strong understanding of how vanilla screeps did this it turns out. I find that visuals help a lot and I wonder if this would be a valuable documentation piece to help future contributors? I recognize much of this is likely taken as "known fact" for many in the screeps dev world, but if it was new to me, it may also be new to someone else. Most importantly, if I have left out a key aspect, or misrepresented, please correct my understanding!

- Adds `docs/architecture.md` — a visual, diagram-first guide to how xxscreeps works
- Links from the Contributing section in README
- Covers: runner/processor/backend split, binary blobs + overlay system, intents, schemas, mod system, `#`/`declare`/`'#quoted'` conventions, room lifecycle
- Aimed at lowering the barrier for new contributors to build a mental model before diving into source

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -63,6 +63,11 @@ open the source files listed at the end.
   supports `--no-backend` for headless operation.
 - All services can run on different machines. All except the main loop
   can have multiple instances.
+- The **launcher** (`npx xxscreeps start`) bootstraps everything: reads
+  the config, sets up the database, and spawns the services. It is not
+  a service itself — once started, the main loop takes over tick
+  orchestration. Flags like `--no-backend` control which services the
+  launcher starts.
 
 ---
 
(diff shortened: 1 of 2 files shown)


docs/architecture.md
```diff
@@ -0,0 +1,433 @@
...
+without requiring a manual migration step.
+
+The archive also contains [Kaitai](https://kaitai.io/) (`.ksy`) format
+descriptors. You can take a blob and its `.ksy` file to
+https://ide.kaitai.io and visually inspect every field in the binary
+layout — a very useful debugging tool.
+
+---
+
+## Overlay-Backed Fields Need `declare`
```
laverdet, line 295: This part should be removed. `withOverlay` automatically declares blob-backed fields. `declare` is actually seldomly-used.

PR #113, opened 2026-04-04
refactor: Move CLI mod out of backend, extract render system

misterwise:
I believe this to be a much cleaner implementation of the CLI now and better aligned with the architecture of xxscreeps more broadly. It does touch on some core systems though, and although I've reviewed it carefully I wouldn't be surprised if one or more of my assumptions is flawed. The biggest change was refactoring out rendering, which touched a lot of files, but I think makes for much cleaner dependency and import overall. If this looks good, I will start looking at some of the harder CLI features that involve editing state (bots, changing rooms, etc.)

## Summary

The CLI mod (`src/mods/backend/cli/`) loaded all backend infrastructure — Koa, HTTP routes, websocket handlers — just to get room rendering for `rooms.load()`. This happened because renderers were registered as side effects in each mod's `backend.ts`, and the only way to trigger them was `importMods('backend')`.

This PR separates rendering from the backend so the CLI can render room snapshots without pulling in HTTP infrastructure:

- **Extract render system** — `rooms.load()` needs renderer registrations (`bindRenderer` calls) to turn game objects into JSON. These were side effects in each mod's `backend.ts`, mixed with Koa routes and websocket handlers. ES modules are all-or-nothing — importing a file runs all its top-level code — so there's no way to load just the renderers without also executing route registrations. The fix: split renderer registrations into `render.ts` files with a new `render` provide slot. Mods that only had renderers in their `backend.ts` now have a `render.ts` instead (11 mods); mods with real backend code (controller, spawn) keep their `backend.ts` with routes only. `server.ts` calls `importMods('render')` before `importMods('backend')`. The CLI loads `importMods('render')` and gets rendering without HTTP infrastructure.

- **Move CLI mod** — `src/mods/backend/cli/` → `src/mods/cli/`. Provides `launcher` instead of `backend`. Loads renderers via `importMods('render')`.

- **Share launcher connections** — The CLI mod gets DB/Shard from the launcher via `launcher-context.ts` instead of creating duplicate connections.

- **Launcher shutdown coordination** — The launcher now catches SIGINT and publishes `{type: 'shutdown'}` on the service channel. `main.ts` listens for it instead of handling SIGINT directly. The tick loop always finishes the current tick before shutting down, matching the existing behavior.

Current CLI commands are read-only — `rooms.load()` renders snapshots, `users.*` and `shards.*` query the database, `system.*` manages server config and pause state. The raw `db`/`shard` keyval providers are also exposed as an admin escape hatch. This already goes beyond the official screeps-launcher CLI, which only works as a socket client to a running server — the offline mode (`xxscreeps cli`) connects directly to persistent storage without a server. The `render` provide slot and shared launcher context lay the groundwork for write operations (bot management, room editing) in a follow-up, where the overlay system and `runOneShot()` can provide typed, schema-validated mutations rather than the raw database writes that the official CLI relies on.

One thing not addressed here is mod extensibility — the sandbox currently hardcodes all available commands. Third-party mods (like admin-utils or screepsmod-portals in the vanilla ecosystem) have no way to register their own CLI helpers without modifying the CLI mod directly. A hook registration pattern (similar to `backend/symbols.ts` hooks for routes and middleware) would let mods provide a `cli` slot and inject commands into the sandbox context. The current design doesn't block this — `createSandbox()` is centralized and easy to extend — but it's worth considering before this ships.

## New functionality

**Offline CLI** (`xxscreeps cli`) — Connects directly to the database without a running server. Works with any persistent storage provider (Redis, file-backed). Useful for inspecting or managing server state without starting the full server. Warns when the provider is `local://` since that data only exists in-memory inside the server process.

## Bug fixes

- **EPIPE crash on client disconnect** — The socket server had no error handler on client connections. If a client disconnected while the server was writing a response, the unhandled `EPIPE` crashed the process.

- **Readline crash on Node 24** — The REPL client called `rl.prompt()` after stdin EOF closed the readline interface. Node 24 throws `ERR_USE_AFTER_CLOSE` where older versions silently ignored it.

- **Concurrent storage saves** — `BlobStorage.save()` and `LocalKeyValResponder.save()` could overlap (periodic save fire-and-forget + shutdown save). Two concurrent saves write to the same `.swp` temp file. Save calls are now chained — each waits for the previous to finish.

- **Backend data loss on shutdown** — `BackendContext.disconnect()` dropped the connection without saving. Added `BackendContext.close()` which saves before disconnecting, used on graceful shutdown.

## Verification

- Stress tested the live socket server: 10x abrupt disconnects, 50x rapid-fire command ordering, 20 concurrent connections, malformed input, half-close, 100KB responses, rapid pause/resume cycling — all passed
- Discovered and fixed EPIPE crash, readline crash, and concurrent save race during stress testing
- Full code review against xxscreeps coding style and architecture guidelines — eliminated dead `.d.ts` files, backward-compat shims, stub `backend.ts` files, duplicate socket path derivation, leaked timers, and a mid-tick shutdown behavior change
- 80 tests pass (55 in-process CLI + 25 socket integration including 6 subprocess smoke tests)
- Zero lint errors across all changed files


laverdet:
I don't think it makes sense to split out 'render' from backend. The backend render format is an internal and undocumented format used only by the Screeps Steam & web clients. If I were to write my own frontend client then all this code would be thrown away and I would send raw blobs, the same as the runtime engine. Basically, it serves no purpose besides compatibility with the existing frontend client. Without the client, the backend is pointless, and without the backend the renderers are meaningless. They are explicitly the same thing.

I actually don't even think the cli should have this rooms.load() method. There is nothing you can do with the data except look at it. The cli should just be using the same runtime environment that the game uses. That format would also give you the option of mutating game state and saving it back to the shard.

misterwise:
Ok, so the approach with CLI that I missed was my focus on alignment with existing CLI for screeps private servers. I can see why you would want to diverge from that. Render is a good way to return similar data structures to say screeps-launcher, but it does limit us. I can rewrite to leave render alone and focus on interacting with the objects themselves which could be much more interesting. Is there anything else about the CLI that you think is an important divergence from current CLI implementations I should consider?

laverdet:
I see no reason for any compatibility with existing CLI snippets people may be using. The xxscreeps keyval & blob architecture is too significant of a departure from the mongo-based backend. Like ok sure there's no reason to change something like `system.sendServerMessage`, but--

https://github.com/screeps/screeps/blob/master/README.md#command-line-interface-cli
```js
// Show all creeps in a room
storage.db['rooms.objects'].find({$and: [{room: 'W1N1'}, {type: 'creep'}]});
```

This is so far removed from the xxscreeps architecture that you could write a 10k LOC compatibility layer and never achieve anything meaningful.

laverdet:
As for what the cli should even do it'd be worth doing community research on that. Personally I just generate a world and then spawn my player into it to see what happens. I don't have any need to, say, pick up a mineral and move it but maybe others would.

misterwise:
I can work with that, and it makes sense to me to align more closely with the actual implementation of xxscreeps. I'll see what I can come up with. Pretty excited to have an acceptable working CLI, I think it will be a big selling point for many people.

misterwise:
Closing this in favor of a rewrite on `feature/cli-commands`.

Your feedback was the right call — the render split is gone, and commands now run inside the game runtime via `runOneShot` so they can mutate state and save back to the shard. Also added a `sandbox` hook for mod extensibility.

New PR coming once it's rebased onto the monorepo layout. Thanks for the steer.

PR #114, opened 2026-04-05
Implement StructureFactory with produce intent and commodity recipes

misterwise:
## Summary
- Adds a `factory` mod (`src/mods/factory/`) with schema, game bindings, `produce` intent, processor, backend renderer, and tests.
- Extracts commodity recipes and factory constants out of `src/game/constants/resource.ts` into `src/mods/factory/constants.ts`.

## Verification
- `npm run build`
- `npm test` — 168/168 pass, including the new factory suite
- `npx eslint` on modified files — 0 errors (4 warnings match baseline patterns in `mods/chemistry/lab.ts` and `mods/chemistry/backend.ts`)

## Notes
- `getEffectiveLevel` / `checkRecipeLevel` intentionally only consider the permanent factory level. The `PWR_OPERATE_FACTORY` bars-block rule is documented inline and left for when the effects substrate lands.


src/game/constants/resource.ts
```diff
@@ -47,552 +47,10 @@ export const RESOURCE_SPIRIT = 'spirit' as const;
...
-			[RESOURCE_SPIRIT]: 3,
-			[RESOURCE_CRYSTAL]: 110,
-			[RESOURCE_GHODIUM_MELT]: 150,
-			[RESOURCE_ENERGY]: 64,
-		},
-	},
-};
-*/
+// Populated by factory mod via Object.assign
+export const COMMODITIES: Partial<Record<string, {
```
laverdet: I think this should be moved too?

laverdet (changes requested):

misterwise:
Thanks for the review! All items addressed in 467c376:

- **Constants moved** — RESOURCE_* and COMMODITIES moved from `game/constants/resource.ts` into `mods/factory/constants.ts`, added to the mod import chain. Follows the chemistry mod pattern (mod owns its constants, directly exports the lookup table instead of Object.assign).
- **Formatting** — Chain recipes reformatted to multi-line.
- **`room.factory` removed** — Confirmed vanilla only assigns `room.controller`, `room.storage`, and `room.terminal` in the Structure constructor. Removed the accessor, Room interface extension, and associated tests. (this was an artifact of how I wanted things to be, rather than how they are :( )
- **Schema key** — Renamed `factory` → `commodity` to match the chemistry mod convention.
- **No-op `if`** — Simplified `getCommodityRecipe` to a single cast return.
- **Check ordering** — Swapped ERR_NOT_ENOUGH_RESOURCES before ERR_FULL to match vanilla.
- **`Fn.accumulate`** — Rather than `.length`, inlined the sum into the existing `Object.entries` loop (one pass, no `Fn` import needed). `.length` would have given component count instead of amount sum, breaking the net-capacity check.
- **`ResourceType`** — Changed processor parameter from `string` to `ResourceType`, removed casts. Kept the `as ResourceType` on `Object.entries` keys since those are inherently `string`.
- **Test cleanup** — Fixed line breaks, removed recipe snapshot test and `room.factory` tests.

PR #120, opened 2026-04-17
Short-circuit Room.findPath when origin === goal

misterwise:
## Summary

`Room.findPath` at `packages/xxscreeps/game/room/path.ts:127` delegates to `PathFinder.roomSearch`, which correctly returns an empty path for a same-position query. But the auto-append branch at lines 132-140 falls through to `origin.isNearTo(goal)` — which is `true` for range 0 — and pushes `goal` onto the empty path, yielding a spurious 1-step result with `dx = dy = 0`.

## Vanilla behavior

Both pathfinder paths in vanilla short-circuit before any search:

- Legacy pathfinder: [`@screeps/engine/src/game/rooms.js:914-916`](https://github.com/screeps/engine/blob/master/src/game/rooms.js#L914-L916)
  ```js
  if (fromX == toX && fromY == toY) {
      return opts.serialize ? '' : [];
  }
  ```

- New pathfinder (`_findPath2`): [`@screeps/engine/src/game/rooms.js:227-229`](https://github.com/screeps/engine/blob/master/src/game/rooms.js#L227-L229)
  ```js
  if (fromPos.isEqualTo(toPos)) {
      return opts.serialize ? '' : [];
  }
  ```

xxscreeps' `findPath` has no upfront cross-room guard and delegates straight to the pathfinder, structurally mirroring `_findPath2`, so `isEqualTo` is the right match.

## Fix

```diff
 findPath(origin: RoomPosition, goal: RoomPosition, options: ... = {}) {

+    if (origin.isEqualTo(goal)) {
+        return options.serialize ? '' : [];
+    }
+
     // Delegate to `PathFinder` for main search
     const result = PathFinder.roomSearch(origin, [ goal ], options);
```

## Testing

Verified against ok-screeps: `creep.pos.findPathTo(creep.pos)` now returns `[]` (or `''` when `serialize: true`), matching vanilla. All other pathfinding tests still pass.

### DIFF
--- a/packages/xxscreeps/game/room/path.ts
+++ b/packages/xxscreeps/game/room/path.ts
@@ -126,6 +126,10 @@ extend(Room, {
 
 	findPath(origin: RoomPosition, goal: RoomPosition, options: FindPathOptions & { serialize?: boolean } = {}) {
 
+		if (origin.isEqualTo(goal)) {
+			return options.serialize ? '' : [];
+		}
+
 		// Delegate to `PathFinder` for main search
 		const result = PathFinder.roomSearch(origin, [ goal ], options);
 


laverdet:
I'll take a closer look at this next week if I can.

Looks like I already made two attempts at getting this right:
https://github.com/laverdet/xxscreeps/commit/f58ab89a9b99d86d764b8387edcf5bbb4e84c532
https://github.com/laverdet/xxscreeps/commit/c3b734397b18f613d3ab4ca41231a483ad43ccd1

You have to be careful because doing `creep.moveTo(source.pos)` will route to a source, which is on a wall (not pathable). If you route to the wall position with `range:0` it will waste a ton of CPU but still return the correct path (1 away from the source).

PR #122, opened 2026-04-18
Scale ConstructionSite.progressTotal for roads by terrain ratio

misterwise:
## Summary

`ConstructionSite.progressTotal` returns the base `CONSTRUCTION_COST[structureType]` for every site, so a road placed on a swamp or wall tile reports the plain-terrain cost instead of the terrain-scaled cost the engine actually charges.

**Vanilla** ([`@screeps/engine/src/processor/intents/room/create-construction-site.js:35-44`](https://github.com/screeps/engine/blob/master/src/processor/intents/room/create-construction-site.js#L35-L44)) scales the stored `progressTotal` by `CONSTRUCTION_COST_ROAD_SWAMP_RATIO` (5x) on swamp and `CONSTRUCTION_COST_ROAD_WALL_RATIO` (150x) on wall. A road site on swamp reports `300 x 5 = 1500`; on wall, `300 x 150 = 45000`.

**xxscreeps (current)** at `packages/xxscreeps/mods/construction/construction-site.ts:34` returns the bare cost from the getter regardless of terrain.

The two adjacent road code paths already scale correctly — RCL energy in `mods/road/road.ts` `checkPlacement` (wall/swamp ratios) and post-build hits in the shared `build.js` completion branch. Only the observable getter was missing the multiplier.

## Fix

Apply the terrain multiplier in the getter for `STRUCTURE_ROAD` only; every other structure type short-circuits through the same single-cost path as before.

## Testing

Verified against ok-screeps.

### DIFF
--- a/packages/xxscreeps/mods/construction/construction-site.ts
+++ b/packages/xxscreeps/mods/construction/construction-site.ts
@@ -17,6 +17,7 @@ const shape = () => struct(RoomObject.format, {
 	...variant('constructionSite'),
 	name: 'string',
 	progress: 'int32',
+	progressTotal: 'int32',
 	structureType: enumerated(...structureFactories.keys() as never as ConstructibleStructureType[]),
 	'#user': Id.format,
 });
@@ -31,7 +32,6 @@ export class ConstructionSite extends withOverlay(RoomObject.RoomObject, shape) 
 	override get '#lookType'() { return C.LOOK_CONSTRUCTION_SITES; }
 	@enumerable override get my() { return this['#user'] === me; }
 	@enumerable get owner() { return userInfo.get(this['#user']); }
-	@enumerable get progressTotal() { return C.CONSTRUCTION_COST[this.structureType]; }
 
 	override '#addToMyGame'(game: GameConstructor) {
 		game.constructionSites[this.id] = this;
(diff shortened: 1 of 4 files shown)


laverdet:
The `construction` mod should not know anything about the `road` mod. This is a bad special case.

Furthermore, there is logic in place already for this:
https://github.com/laverdet/xxscreeps/blob/819a8632ff5f0bb9c9a042993a7897cec1b8a5c1/packages/xxscreeps/mods/road/road.ts#L47-L59

I think the intention was that `checkPlacement` would return the construction cost, or `null` on placement failure, but that never got implemented. `progressTotal` should just be stored directly on the construction site's blob overlay.

misterwise:
> The `construction` mod should not know anything about the `road` mod. This is a bad special case.
> 
> Furthermore, there is logic in place already for this:
> 
> https://github.com/laverdet/xxscreeps/blob/819a8632ff5f0bb9c9a042993a7897cec1b8a5c1/packages/xxscreeps/mods/road/road.ts#L47-L59
> 
> I think the intention was that `checkPlacement` would return the construction cost, or `null` on placement failure, but that never got implemented. `progressTotal` should just be stored directly on the construction site's blob overlay.

Got it, added it to the blob, much cleaner.

PR #126, opened 2026-04-20
Fix controller safe-mode guards and generateSafeMode intent

misterwise:
Three related bugs in `mods/controller/` surface when a player drives a
safe-mode lifecycle through a CARRY/MOVE ghodium-ferry creep.

## `checkActivateSafeMode` (`controller.ts:103-115`)

Vanilla's `StructureController.activateSafeMode` (see
[`@screeps/engine/src/game/structures.js:211-234`](https://github.com/screeps/engine/blob/master/src/game/structures.js#L211-L234))
returns:

- `ERR_TIRED` when `ticksToDowngrade < CONTROLLER_DOWNGRADE[level]/2 - CONTROLLER_DOWNGRADE_SAFEMODE_THRESHOLD`
  **or** when the controller is `upgradeBlocked`
- `ERR_BUSY` when any of the player's owned controllers already has an active safe mode

xxscreeps was missing the downgrade-threshold and `upgradeBlocked`
guards, and the `safeMode → ERR_BUSY` check was scoped to the single
target controller rather than the player's whole portfolio. Add all
three. The cross-room scan subsumes the old single-target `safeMode`
branch — safe mode on the target room satisfies the scan.

## `checkGenerateSafeMode` (`creep.ts:153`)

Required a `WORK` body part via `checkCommon(creep, C.WORK)`. Vanilla's
[`Creep.prototype.generateSafeMode`](https://github.com/screeps/engine/blob/master/src/game/creeps.js#L1049-L1070)
requires only `my`, `!spawning`, `SAFE_MODE_COST` ghodium, and range —
no body-part requirement, consistent with the pattern that
resource-sacrifice intents (transfer/withdraw/pickup/drop) don't need
WORK. Drop the `WORK` argument.

## `generateSafeMode` processor (`processor.ts:126`)

Once the user-side WORK guard is gone and the processor actually runs,
it surfaces a latent bug: the ghodium decrement was written as
`creep.store[C.RESOURCE_GHODIUM] -= C.SAFE_MODE_COST`, which is a
no-op on the schema-backed Store proxy. `safeModeAvailable` incremented
but the creep kept its ghodium — free safe-mode charges. Use
`store['#subtract']`, matching the pattern in `upgradeController` a
few lines down.

Verified against ok-screeps.

### DIFF
--- a/packages/xxscreeps/mods/controller/controller.ts
+++ b/packages/xxscreeps/mods/controller/controller.ts
@@ -1,7 +1,7 @@
 import type { Room } from 'xxscreeps/game/room/index.js';
 import { chainIntentChecks } from 'xxscreeps/game/checks.js';
 import * as C from 'xxscreeps/game/constants/index.js';
-import { Game, intents, userInfo } from 'xxscreeps/game/index.js';
+import { Game, hooks, intents, userInfo } from 'xxscreeps/game/index.js';
 import { OwnedStructure, checkMyStructure, ownedStructureFormat } from 'xxscreeps/mods/structure/structure.js';
 import { compose, declare, struct, variant, withOverlay } from 'xxscreeps/schema/index.js';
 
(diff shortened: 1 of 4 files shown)


packages/xxscreeps/mods/controller/controller.ts
```diff
@@ -106,10 +106,21 @@ export function checkActivateSafeMode(controller: StructureController) {
 		() => {
 			if (controller.safeModeAvailable <= 0) {
 				return C.ERR_NOT_ENOUGH_RESOURCES;
-			} else if (controller.safeModeCooldown) {
+			}
+			const downgradeThreshold = C.CONTROLLER_DOWNGRADE[controller.level]! / 2
+				- C.CONTROLLER_DOWNGRADE_SAFEMODE_THRESHOLD;
+			if (
+				controller.safeModeCooldown ||
+				(controller.upgradeBlocked ?? 0) > 0 ||
```
laverdet: `controller.upgradeBlocked !== undefined`

packages/xxscreeps/mods/controller/controller.ts
```diff
@@ -106,10 +106,21 @@ export function checkActivateSafeMode(controller: StructureController) {
...
+				(controller.ticksToDowngrade ?? Infinity) < downgradeThreshold
+			) {
 				return C.ERR_TIRED;
-			} else if (controller.safeMode) {
-				return C.ERR_BUSY;
+			}
+			for (const room of Object.values(Game.rooms)) {
```
laverdet: Problems:
- Game.rooms doesn't work in the processor. It will just have the one room. So this check would only run in the runtime
- What if the user invokes `activateSafeMode` on more than one controller in the same tick? `safeMode` is not updated until the tick processes, so you'd allow them to activate their whole colony.

Interestingly, vanilla also only does the world-wide check in their runtime. So with intent injection you could activate safe mode on more than one controller. It's probably a vulnerability on their side as well.

I would say lock this down in the same manner that vanilla does. You'd have a `lastActivateSafeMode` flag, similar to `createConstructionSite`'s `createdNames`. Remove the old intent if they issue two safe modes in the same tick.

PR #128, opened 2026-04-20
Fix structure ownership precedence in checkMyStructure and destroy

misterwise:
Two bugs in `mods/structure/structure.ts`:

**1. `checkMyStructure` precedence** (`structure.ts:203`). The `!structure.my && !structure.room.controller?.my` fallback lets another player's structure succeed when the caller owns the room controller. Vanilla's per-intent methods gate on `!this.my` alone — `StructureLab.prototype.runReaction` (`@screeps/engine/src/game/structures.js:318`), `StructureObserver.prototype.observeRoom` (line 549), `StructureFactory.prototype.produce` (line 1093). Require `structure.my`.

**2. `Structure.prototype.destroy`** (vanilla `structures.js:72-78`) checks only `room.controller.my`, never `this.my`. xxscreeps's `checkDestroy` delegated to `checkMyStructure(structure, Structure)`, which accepted self-ownership — a player could destroy their own rampart placed in another player's room. Inline `controller.my` in `checkDestroy`; keep `checkMyStructure` as the structure-ownership path only.

**Intentional divergence on `StructureController.my` for never-claimed controllers.** Vanilla distinguishes never-claimed (`user` field absent → `my === undefined`) from post-unclaim (`user === null` → `my === false`). xxscreeps serializes `#user` as `string | null` with no "absent" representation, so a single-branch getter must collapse one case onto the other. `OwnedStructure.my` is simplified to `this['#user'] === me`, consistent with `Creep.my` and `ConstructionSite.my`, and the never-claimed case returns `false` instead of `undefined`. Post-unclaim (the case player code actually branches on) now matches vanilla.

Verified against ok-screeps.

### DIFF
--- a/packages/xxscreeps/mods/structure/structure.ts
+++ b/packages/xxscreeps/mods/structure/structure.ts
@@ -203,7 +203,7 @@ export function checkWall(pos: RoomPosition) {
 export function checkMyStructure(structure: Structure, constructor: abstract new(...args: any[]) => any) {
 	if (!(structure instanceof constructor)) {
 		return C.ERR_INVALID_ARGS;
-	} else if (!structure.my && !structure.room.controller?.my) {
+	} else if (!structure.my) {
 		return C.ERR_NOT_OWNER;
 	}
 	return C.OK;
@@ -218,7 +218,8 @@ export function checkIsActive(structure: Structure) {
 
 export function checkDestroy(structure: Structure) {
 	return chainIntentChecks(
-		() => checkMyStructure(structure, Structure),
+		() => structure instanceof Structure ? C.OK : C.ERR_INVALID_ARGS,
+		() => structure.room.controller?.my ? C.OK : C.ERR_NOT_OWNER,
 		() => {
 			if ((structure.hits ?? 0) <= 0) {
 				return C.ERR_INVALID_TARGET;
--- a/packages/xxscreeps/mods/structure/structure.ts
+++ b/packages/xxscreeps/mods/structure/structure.ts
@@ -105,10 +105,7 @@ export class OwnedStructure extends withOverlay(Structure, ownedShape) {
 	/**
 	 * Whether this is your own structure.
 	 */
-	@enumerable override get my() {
-		const user = this['#user'];
-		return user === null ? undefined : user === me;
-	}
+	@enumerable override get my() { return this['#user'] === me; }
 
 	// TODO: This may be invoked each tick until the processor calls isActive. The cache
 	// does not persist from runner into processor, only from processor into runtime.


packages/xxscreeps/mods/structure/structure.ts
```diff
@@ -107,7 +107,7 @@ export class OwnedStructure extends withOverlay(Structure, ownedShape) {
 	 */
 	@enumerable override get my() {
 		const user = this['#user'];
-		return user === null ? undefined : user === me;
+		return user === null ? false : user === me;
```
laverdet: This could just be `user === me` without the ternary?

misterwise: Yes indeed.

laverdet:
Could you verify one more time that this matches vanilla? Consider the case that a lab is constructed in a room, and then another player invades and takes control of the room, but leaves the previous player's lab. Should that lab be allowed to produce? I think vanilla returns `ERR_RCL_NOT_ENOUGH` in this case.

misterwise:
Looking closely again, I verified the behaviour for the case you specified: `ERR_RCL_NOT_ENOUGH` for the owner of the lab, and `ERR_NOT_OWNER` for the owner of the room — matches vanilla. But it highlighted that the change to `my()` was not sufficient. What it was trying to fix is that in vanilla `controller.my` is `undefined` for never-claimed rooms and `false` for post-unclaim rooms. I dropped the changes to my() since we're replacing one incorrect behaviour for a separate incorrect behaviour at the moment. See the updated PR body for analysis we can close separately if it's worth it. For strict behaviour matching we can, but in most cases I can imagine this never actually impacts player code.

laverdet:
I'm ok dropping `controller.my === undefined` for never-claimed controllers. That's a dumb quirk.

I think the previous `my` change you had (always returning true/false for OwnedStructure) seemed correct.

I'm not sure the intent changes you have here in the force push are correct.

misterwise:
Added back the `OwnedStructure.my` simplification (`user === me`) in the latest push — the `null → undefined` branch is gone. I agree that if we pick a place to diverge, false is much better than undefined.

On the intent changes: both were derived from reading every vanilla call site that `checkMyStructure` and `Structure.destroy` stand in for.

**1. `checkMyStructure`: strip the `controller.my` disjunction.**

Old: `!structure.my && !structure.room.controller?.my` — pass if the caller owns the structure OR the room.

Every vanilla method that ends up funnelling through `checkMyStructure` in xxscreeps gates on `!this.my` alone, with no controller-ownership fallback:

| Method | Vanilla line (`engine/src/game/structures.js`) |
|---|---|
| `StructureLab.runReaction` / `reverseReaction` / `boostCreep` | 319, 362, 414 |
| `StructureObserver.observeRoom` | 550 |
| `StructurePowerSpawn.processPower` | 615 |
| `StructureTerminal.send` | 716 |
| `StructureTower.attack` / `heal` / `repair` | 768, 788, 808 |
| `StructureSpawn.canCreateCreep` / `notifyWhenAttacked` | 899, 1223 |
| `StructureFactory.produce` | 1093 |
| `StructureController.unclaim` / `activateSafeMode` | 204, 214 |

The invaded-lab scenario makes the divergence concrete:

| Actor | `structure.my` | `controller.my` | Old xxscreeps | Vanilla | New xxscreeps |
|---|---|---|---|---|---|
| Lab owner (room now owned by another player) | true | false | pass → `ERR_RCL_NOT_ENOUGH` | pass → `ERR_RCL_NOT_ENOUGH` | pass → `ERR_RCL_NOT_ENOUGH` |
| Room owner (lab still owned by previous player) | false | true | **pass → runs the foreign lab** | `ERR_NOT_OWNER` | `ERR_NOT_OWNER` |

The first row (the `ERR_RCL_NOT_ENOUGH` path) is unchanged. The second row is the bug: the `&& !controller.my` clause silently granted the new room-owner control of every owned structure the previous player left behind — lab, terminal, tower, factory, observer, spawn, etc.

**2. `checkDestroy`: inline `controller.my` instead of delegating.**

Vanilla `Structure.prototype.destroy` (structures.js:72–78) gates *only* on `!this.room.controller || !this.room.controller.my`. It never reads `this.my`.

Old xxscreeps routed `checkDestroy → checkMyStructure`, so the disjunction applied:

| Target | `structure.my` | `controller.my` | Old xxscreeps | Vanilla |
|---|---|---|---|---|
| Own rampart in another player's room | true | false | pass → **rampart destroyed** | `ERR_NOT_OWNER` |
| Foreign structure in own room | false | true | pass (coincidentally) | pass |

The coincidental case keeps working; the buggy case (destroying an owned structure inside a room owned by someone else) is what the inlined check fixes.

**Summary.** Every call site that went through `checkMyStructure` in xxscreeps maps to a vanilla method gated on `!this.my` alone; `destroy` is the one exception, and it's been inlined. The cases where the old disjunction happened to match vanilla still match; the cases it got wrong are the ones this PR fixes.

PR #129, opened 2026-04-20
Fix cooldown anchoring for lab, factory, and extractor intents

misterwise:
xxscreeps runs intent processing and the cooldown getter at the same `Game.time`, while vanilla writes `cooldownTime` in the intent processor at `gameTime = T` and reads the getter in user code at `runtimeData.time = T+1`. Vanilla's observable cooldown is therefore implicitly `T + DURATION - (T+1) = DURATION - 1`; xxscreeps sees the full `DURATION` unless the write compensates.

Three sites drifted from vanilla's observable:

1. `mods/chemistry/processor.ts` runReaction/reverseReaction/unboostCreep (lines 32, 92, 139) wrote `Game.time + X`, reporting `REACTION_TIME` instead of `REACTION_TIME - 1`. Subtract 1.

2. `mods/factory/processor.ts` produce (line 22) had the same bug, reporting the full recipe cooldown. Subtract 1.

3. `mods/mineral/processor.ts` harvest (line 21) wrote `Game.time + EXTRACTOR_COOLDOWN - 1`, but vanilla's extractor uses a counter + tick-processor decrement (`@screeps/engine/src/processor/intents/extractors/tick.js`), so it never gets the implicit `-1` that cooldownTime-based intents do. The `-1` over-corrected; drop it so the raw `EXTRACTOR_COOLDOWN` is visible at the first observer tick.

The same pattern was applied in `mods/logistics/processor.ts:17` (merged as #127): the `-1` is the engine-level compensation for xxscreeps's single-Game.time execution model.

Verified against ok-screeps.

### DIFF
--- a/packages/xxscreeps/mods/chemistry/processor.ts
+++ b/packages/xxscreeps/mods/chemistry/processor.ts
@@ -29,11 +29,6 @@ const intents = [
 		lab.store['#add'](product, C.LAB_REACTION_AMOUNT);
 		left.store['#subtract'](left.mineralType!, C.LAB_REACTION_AMOUNT);
 		right.store['#subtract'](right.mineralType!, C.LAB_REACTION_AMOUNT);
-		// `-1` compensates for xxscreeps running intent processing and the cooldown
-		// getter at the same Game.time, which drops the implicit decrement vanilla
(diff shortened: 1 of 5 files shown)


laverdet:
The code is good but please delete these comments which don't add anything of value

PR #130, opened 2026-04-20
Fix Store.getFreeCapacity null propagation and OpenStore pool semantics

misterwise:
Three connected bugs in `mods/resource/store.ts`, all keyed off the same invariant: `Store.getCapacity(rt) === null` is the canonical "this store cannot hold this resource type" signal. Three call paths must respect it.

1. **Base `Store.getFreeCapacity`** did `getCapacity(rt)! - getUsedCapacity(rt)!` with non-null assertions. For `LabStore` (and any restricted store) queried with a wrong resource type both operands are `null`; `null - null` coerces to `0`, losing the null signal downstream callers rely on. Short-circuit to `null` when either operand is `null`.

2. **`OpenStore`** inherited the per-type subtraction formula, which is wrong for shared-capacity stores: `#capacity - this[rt]` returns full capacity for any empty slot regardless of what else fills the pool. A creep with `{energy: 45}` in a 50-cap store reported `getFreeCapacity('H') === 50` instead of `5`, so e.g. the mineral overflow check at `mods/mineral/processor.ts:14` absorbed all harvested H instead of dropping it. Override `OpenStore.getFreeCapacity` to return `#capacity - _sum` regardless of `resourceType`.

3. **`checkHasResource` / `checkHasCapacity`** bypassed the store API and tested raw amounts, so a wrong-resource transfer/withdraw to an energy-only structure returned `ERR_FULL` or `ERR_NOT_ENOUGH_RESOURCES` instead of `ERR_INVALID_TARGET`. Gate both helpers on `getCapacity(rt) === null` first.

Verified against ok-screeps.

### DIFF
--- a/packages/xxscreeps/mods/resource/store.ts
+++ b/packages/xxscreeps/mods/resource/store.ts
@@ -339,9 +339,10 @@ export function calculateChecked(object1: WithStore | undefined, object2: WithSt
 }
 
 export function checkHasCapacity(target: WithStore, resourceType: ResourceType, amount: number) {
-	if (target.store.getCapacity(resourceType) === null) {
+	const free = target.store.getFreeCapacity(resourceType);
+	if (free === null) {
 		return C.ERR_INVALID_TARGET;
-	} else if (target.store.getFreeCapacity(resourceType)! >= Math.max(1, amount)) {
+	} else if (free >= Math.max(1, amount)) {
 		return C.OK;
 	} else {
 		return C.ERR_FULL;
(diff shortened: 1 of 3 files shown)


packages/xxscreeps/mods/resource/store.ts
```diff
@@ -329,7 +339,9 @@ export function calculateChecked(object1: WithStore | undefined, object2: WithSt
 }
 
 export function checkHasCapacity(target: WithStore, resourceType: ResourceType, amount: number) {
-	if (target.store.getFreeCapacity(resourceType) >= Math.max(1, amount)) {
+	if (target.store.getCapacity(resourceType) === null) {
+		return C.ERR_INVALID_TARGET;
+	} else if (target.store.getFreeCapacity(resourceType)! >= Math.max(1, amount)) {
```
laverdet: Assign `target.store.getFreeCapacity(resourceType)` to a temporary variable. First, check null, then check against the value. That way you don't invoke the same methods twice and also you don't need the type assertion.

PR #131, opened 2026-04-21
Fix RawMemory.set guards and implement foreign/public memory segments

misterwise:
Implements foreign and public memory segments. The `RawMemory.set` guards and `_parsed` handling have been split into #140; this PR now carries only the foreign-segment round-trip and the public-segment plumbing. Depends on / stacks naturally with #140 but the two don't overlap so either can land first.

## What changed

- **Capture + flush** follows the existing `activeSegments` pattern: module-local pending state, `flushPublicSegments` / `flushDefaultPublicSegment` / `flushForeignSegment` helpers consumed by `game.ts send`, persisted by `driver.ts save` via new `model.ts` helpers.
- **Tri-state request payloads.** `undefined` = no call this tick, `null` = explicit clear, value = new request. Applies uniformly to `foreignSegmentRequest`, `defaultPublicSegmentUpdate`, and the foreign payload itself (`ForeignSegmentPayload | null | undefined`), so the runtime stops recreating the `foreignSegment` lazy-getter each tick when nothing changed.
- **Shard-scoped persistence.**
  - `user/${userId}` hash gains a `defaultPublicSegment` field (mirrors vanilla's users-doc).
  - `user/${userId}/activeForeignSegment` hash stores `{username, userId, segmentId}` — `hgetall`-friendly so reads don't go through `JSON.parse`.
  - `user/${userId}/publicSegments` is a redis set (its own key, not a hash field — mirrors vanilla's env semantics).
  - All helpers take `Shard`, not `Database`.
- **Delivery.** `driver.ts refresh` `Promise.all`s `isPublicSegment` + `loadMemorySegmentBlob`, packs the raw bytes into `payload.foreignSegment`. Runtime installs a lazy `data` getter on `RawMemory.foreignSegment` so `typedArrayToString` runs at most once and only if the script reads `.data`.
- **Pubsub instead of per-tick refetch.** Publishers send `{type: 'segment', id}` per segment write and `{type: 'publicSet'}` when the public set changes. Each runner subscribes to its current foreign target's channel; the cached foreign payload is refreshed only on notification (or on target change), not every tick. Subscription lifecycle is tracked via an `Effect | undefined` disconnect handler.
- **Resolution.** `saveActiveForeignSegment` folds the username→userId lookup and `defaultPublicSegment` fallback. Failed resolution deletes the stored key and returns `null` rather than carrying a prior request forward silently.
- **Fan-out.** `driver.ts save` runs every independent await through `Promise.all` (isPublic + loadSegment on refresh, del + sadd in `savePublicSegments`, and the whole save fan-out including the foreign-request resolve + re-subscribe in an IIFE). Pubsub publishes fire alongside their writes — reads are tick-synchronized so there's no race.

## Test plan
- [x] Verified full round-trip against ok-screeps (setPublicSegments, setDefaultPublicSegment, setActiveForeignSegment with and without explicit id, clear path)
- [x] Zero new lint warnings on the four changed files

### DIFF
--- a/packages/xxscreeps/mods/memory/driver.ts
+++ b/packages/xxscreeps/mods/memory/driver.ts
@@ -1,26 +1,27 @@
-import type { SegmentPayload, flush } from './memory.js';
-import type { TickResult } from 'xxscreeps/engine/runner/index.js';
+import type { ForeignSegmentPayload, SegmentPayload, flush } from './memory.js';
+import type { ForeignSegmentRequest, StoredForeignSegmentRequest } from './model.js';
+import type { Effect } from 'xxscreeps/utility/types.js';
 import { hooks } from 'xxscreeps/engine/runner/index.js';
 import { Fn } from 'xxscreeps/functional/fn.js';
 import { kMaxActiveSegments } from './memory.js';
-import { loadMemorySegmentBlob, loadUserMemoryBlob, saveMemoryBlob, saveMemorySegmentBlob } from './model.js';
+import { getPublicSegmentChannel, isPublicSegment, loadActiveForeignSegment, loadMemorySegmentBlob, loadUserMemoryBlob, saveActiveForeignSegment, saveDefaultPublicSegment, saveMemoryBlob, saveMemorySegmentBlob, savePublicSegments } from './model.js';
 
-// Receive and send memory payloads from driver
 declare module 'xxscreeps/engine/runner/index.js' {
 	interface InitializationPayload {
 		memoryBlob: Readonly<Uint8Array> | null;
 	}
 	interface TickPayload {
 		memorySegments?: SegmentPayload[];
+		// Tri-state: `undefined` = no change, `null` = clear, object = install
+		foreignSegment?: ForeignSegmentPayload | null;
 	}
 	interface TickResult {
 		activeSegmentsRequest: number[] | null;
-		foreignSegmentRequest: null | {
-			id: number | undefined;
-			username: string;
-		};
+		foreignSegmentRequest: ForeignSegmentRequest | null | undefined;
 		memorySegmentsUpdated: SegmentPayload[] | null;
 		memoryUpdated: ReturnType<typeof flush>;
+		defaultPublicSegmentUpdate: number | null | undefined;
+		publicSegmentsUpdate: number[] | undefined;
 	}
 }
 
@@ -28,14 +29,37 @@ hooks.register('runnerConnector', player => {
 	const { shard, userId } = player;
 	let activeSegments: Set<number>;
 	let nextSegments: Set<number> | undefined;
-	let foreignSegmentRequest: TickResult['foreignSegmentRequest'] = null;
-	return [ undefined, {
+	let activeForeignSegment: StoredForeignSegmentRequest | null = null;
+	let subscribedUserId: string | null = null;
+	let subscriptionEffect: Effect | undefined;
+	let foreignDirty = false;
+
+	async function syncForeignSubscription() {
+		const target = activeForeignSegment?.userId ?? null;
+		if (target === subscribedUserId) {
+			return;
+		}
+		subscriptionEffect?.();
+		subscriptionEffect = undefined;
+		subscribedUserId = target;
+		foreignDirty = true;
+		if (target !== null) {
+			const subscription = await getPublicSegmentChannel(shard, target).subscribe();
+			subscription.listen(message => {
+				if (message.type === 'publicSet' || message.id === activeForeignSegment?.segmentId) {
+					foreignDirty = true;
+				}
+			});
+			subscriptionEffect = () => subscription.disconnect();
+		}
+	}
+
+	return [ () => subscriptionEffect?.(), {
 		async initialize(payload) {
-			// Reset state that mirrors the runtime sandbox
 			activeSegments = new Set();
 			nextSegments = undefined;
-			foreignSegmentRequest = null;
-			// Get current memory payload
+			activeForeignSegment = await loadActiveForeignSegment(shard, userId);
+			await syncForeignSubscription();
 			payload.memoryBlob = await loadUserMemoryBlob(shard, userId);
 		},
 
@@ -52,24 +76,59 @@ hooks.register('runnerConnector', player => {
 				activeSegments = nextSegments;
 				nextSegments = undefined;
 			}
-			if (foreignSegmentRequest) {
-				// TODO
+			// Only refetch and push to the runtime when the publisher notified us (or on target
+			// change). Otherwise the runtime holds last tick's payload and doesn't recreate the
+			// lazy-getter object.
+			if (foreignDirty) {
+				foreignDirty = false;
+				if (activeForeignSegment) {
+					const { username, userId: targetUserId, segmentId } = activeForeignSegment;
+					const [ isPublic, blob ] = await Promise.all([
+						isPublicSegment(shard, targetUserId, segmentId),
+						loadMemorySegmentBlob(shard, targetUserId, segmentId),
+					]);
+					payload.foreignSegment = isPublic && blob !== null
+						? { username, id: segmentId, bytes: blob }
+						: null;
+				} else {
+					payload.foreignSegment = null;
+				}
 			}
 		},
 
 		async save(payload) {
-			// Update activate segments
+			// Update active segments
 			if (payload.activeSegmentsRequest) {
 				nextSegments = new Set(Fn.take(payload.activeSegmentsRequest, kMaxActiveSegments));
 			}
-			foreignSegmentRequest = payload.foreignSegmentRequest;
+
+			const foreignRequest = payload.foreignSegmentRequest;
+			const segmentWrites = payload.memorySegmentsUpdated
+				? [ ...Fn.take(payload.memorySegmentsUpdated, kMaxActiveSegments) ]
+				: [];
+			const publicSegmentsChanged = payload.publicSegmentsUpdate !== undefined;
+			const channel = getPublicSegmentChannel(shard, userId);
+
 			await Promise.all([
 				// Save primary memory blob
 				payload.memoryUpdated.payload && saveMemoryBlob(shard, userId, payload.memoryUpdated.payload),
 				// Save memory segments
-				payload.memorySegmentsUpdated &&
-					Promise.all(Fn.map(Fn.take(payload.memorySegmentsUpdated, kMaxActiveSegments),
-						segment => saveMemorySegmentBlob(shard, userId, segment.id, segment.payload))),
+				...Fn.map(segmentWrites, segment => saveMemorySegmentBlob(shard, userId, segment.id, segment.payload)),
+				// Save default public segment
+				payload.defaultPublicSegmentUpdate !== undefined
+					&& saveDefaultPublicSegment(shard, userId, payload.defaultPublicSegmentUpdate),
+				// Save public segment set
+				publicSegmentsChanged && savePublicSegments(shard, userId, payload.publicSegmentsUpdate!),
+				// Notify foreign readers alongside the writes — reads are tick-synchronized, so no
+				// race. Segment updates fire per-id; public-set changes fire once.
+				...Fn.map(segmentWrites, segment => channel.publish({ type: 'segment', id: segment.id })),
+				publicSegmentsChanged && channel.publish({ type: 'publicSet' }),
+				// Resolve + persist the foreign-segment request, then re-subscribe. These are
+				// serial; wrap in an IIFE so the rest of the Promise.all runs in parallel.
+				foreignRequest !== undefined && async function() {
+					activeForeignSegment = await saveActiveForeignSegment(shard, userId, activeForeignSegment, foreignRequest);
+					await syncForeignSubscription();
+				}(),
 			]);
 		},
 	} ];
--- a/packages/xxscreeps/mods/memory/memory.ts
+++ b/packages/xxscreeps/mods/memory/memory.ts
@@ -9,10 +9,11 @@ export const kMaxMemorySegmentLength = 100 * 1024;
 
 let activeSegments = new Map<number, string>();
 let didUpdateSegments = false;
-let requestedForeignSegment: null | {
-	id: number | undefined;
-	username: string;
-};
+// `undefined` = no call this tick, `null` = explicit clear, object = new request. Matches the
+// tri-state convention on the driver payload.
+let requestedForeignSegment: { id: number | undefined; username: string } | null | undefined;
+let requestedDefaultPublicSegment: number | null | undefined;
+let requestedPublicSegments: number[] | undefined;
 let memory: Uint16Array;
 let memoryLength = 0;
 let string: string | undefined;
@@ -66,6 +67,14 @@ export const RawMemory = {
 	segments: {} as Record<string, string>,
 
 	/**
+	 * An object with a memory segment of another user available on this tick. Use
+	 * `setActiveForeignSegment` to fetch this. The object contains the following keys: `username`,
+	 * `id`, and `data`. The segment is only available if the target user has marked it public via
+	 * `setPublicSegments`.
+	 */
+	foreignSegment: undefined as ForeignSegment | undefined,
+
+	/**
 	 * Get a raw string representation of the `Memory` object.
 	 */
 	get() {
@@ -116,7 +125,14 @@ export const RawMemory = {
 	 * @param id The ID of the requested segment from 0 to 99. If undefined, the user's default public
 	 * segment is requested as set by setDefaultPublicSegment.
 	 */
-	setActiveForeignSegment(username: string, id?: number) {
+	setActiveForeignSegment(username: string | null, id?: number) {
+		if (username === null) {
+			requestedForeignSegment = null;
+			return;
+		}
+		if (id !== undefined && !isValidSegmentId(id)) {
+			throw new Error(`"${id}" is not a valid segment ID`);
+		}
 		requestedForeignSegment = { id, username };
 	},
 
@@ -126,14 +142,29 @@ export const RawMemory = {
 	 * @param id The ID of the memory segment from 0 to 99. Pass `null` to remove your default public
 	 * segment.
 	 */
-	setDefaultPublicSegment(_id: number) { console.error('TODO: setDefaultPublicSegment'); },
+	setDefaultPublicSegment(id: number | null) {
+		if (id !== null && !isValidSegmentId(id)) {
+			throw new Error(`"${id}" is not a valid segment ID`);
+		}
+		requestedDefaultPublicSegment = id;
+	},
 
 	/**
 	 * Set specified segments as public. Other users will be able to request access to them using `setActiveForeignSegment`.
 	 * @param ids An array of segment IDs. Each ID should be a number from 0 to 99. Subsequent calls
 	 * of `setPublicSegments` override previous ones.
 	 */
-	setPublicSegments(_ids: number[]) { /*console.error('TODO: setPublicSegments')*/ },
+	setPublicSegments(ids: number[]) {
+		if (!Array.isArray(ids)) {
+			throw new TypeError(`"${ids}" is not an array`);
+		}
+		for (const id of ids) {
+			if (!isValidSegmentId(id)) {
+				throw new Error(`"${id}" is not a valid segment ID`);
+			}
+		}
+		requestedPublicSegments = [ ...ids ];
+	},
 };
 
 /**
@@ -237,6 +268,21 @@ export type SegmentPayload = {
 	payload: Readonly<Uint8Array> | null;
 };
 
+// Wire shape: bytes from the driver. The player-visible `RawMemory.foreignSegment.data` lives
+// behind a lazy getter installed by `loadForeignSegment` so the UTF-16 decode only runs if the
+// script actually reads the string.
+export type ForeignSegmentPayload = {
+	username: string;
+	id: number;
+	bytes: Readonly<Uint8Array>;
+};
+
+type ForeignSegment = {
+	username: string;
+	id: number;
+	data: string;
+};
+
 export function isValidSegmentId(id: number) {
 	return Number.isInteger(id) && id >= 0 && id < kMaxMemorySegmentId;
 }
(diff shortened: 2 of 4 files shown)


packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -1,9 +1,26 @@
...
+import type { StoredForeignSegmentRequest } from './model.js';
+import type { Database } from 'xxscreeps/engine/db/index.js';
 import type { TickResult } from 'xxscreeps/engine/runner/index.js';
+import * as User from 'xxscreeps/engine/db/user/index.js';
 import { hooks } from 'xxscreeps/engine/runner/index.js';
 import { Fn } from 'xxscreeps/functional/fn.js';
+import { typedArrayToString } from 'xxscreeps/utility/string.js';
 import { kMaxActiveSegments } from './memory.js';
-import { loadMemorySegmentBlob, loadUserMemoryBlob, saveMemoryBlob, saveMemorySegmentBlob } from './model.js';
+import {
```
laverdet: No need to break these lines. Imports are usually managed automatically.

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -12,29 +29,54 @@ declare module 'xxscreeps/engine/runner/index.js' {
...
 	}
 	interface TickResult {
 		activeSegmentsRequest: number[] | null;
-		foreignSegmentRequest: null | {
-			id: number | undefined;
-			username: string;
-		};
+		// `request: null` distinguishes an explicit `setActiveForeignSegment(null)`
+		// (clear the stored request) from an absent call (no change).
+		foreignSegmentRequest: null | { request: null | ForeignSegmentRequest };
```
laverdet: Use null = clear, undefined = no change, object = change

With this format you don't need to nest the payload structure any further than it already is.

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -12,29 +29,54 @@ declare module 'xxscreeps/engine/runner/index.js' {
...
 		memorySegmentsUpdated: SegmentPayload[] | null;
 		memoryUpdated: ReturnType<typeof flush>;
+		defaultPublicSegmentUpdate: { id: number | null } | null;
```
laverdet: Same thing: number | null | undefined

packages/xxscreeps/mods/memory/model.ts
```diff
@@ -34,3 +41,55 @@ export async function saveMemorySegmentBlob(shard: Shard, userId: string, segmen
 		}
 	}
 }
+
+export async function loadDefaultPublicSegment(db: Database, userId: string): Promise<number | null> {
```
laverdet: All of these functions should accept `Shard`, not `Database`. Memory and RawMemory are not shared between shards.

packages/xxscreeps/mods/memory/model.ts
```diff
@@ -34,3 +41,55 @@ export async function saveMemorySegmentBlob(shard: Shard, userId: string, segmen
...
+	const raw = await db.data.hget(User.infoKey(userId), 'defaultPublicSegment');
```
laverdet: This should be its own key, not a key on an existing hash.

packages/xxscreeps/mods/memory/model.ts
```diff
@@ -34,3 +41,55 @@ export async function saveMemorySegmentBlob(shard: Shard, userId: string, segmen
...
+	if (raw === null || raw === '') {
```
laverdet: Why would `raw` be an empty string? You should not allow invalid data to be stored in the model.

packages/xxscreeps/mods/memory/model.ts
```diff
@@ -34,3 +41,55 @@ export async function saveMemorySegmentBlob(shard: Shard, userId: string, segmen
...
+		return null;
+	}
+	const id = Number(raw);
+	return isValidSegmentId(id) ? id : null;
```
laverdet: Sloppy

packages/xxscreeps/mods/memory/model.ts
```diff
@@ -34,3 +41,55 @@ export async function saveMemorySegmentBlob(shard: Shard, userId: string, segmen
...
+}
+
+// `KeyValProvider` lacks `hdel`, so clear-writes store an empty string; `load*` treats empty as "unset".
```
laverdet: Don't workaround missing database operations with slop like this. It'd be a oneliner to add hdel to the default and redis providers. But also, by making public segment its own key you sidestep the issue entirely. redis hash fields are frequently misused in this manner.

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -12,29 +29,54 @@ declare module 'xxscreeps/engine/runner/index.js' {
...
 	let nextSegments: Set<number> | undefined;
-	let foreignSegmentRequest: TickResult['foreignSegmentRequest'] = null;
+	let activeForeignSegment: StoredForeignSegmentRequest | null = null;
 	return [ undefined, {
 		async initialize(payload) {
 			// Reset state that mirrors the runtime sandbox
 			activeSegments = new Set();
 			nextSegments = undefined;
-			foreignSegmentRequest = null;
+			activeForeignSegment = await loadActiveForeignSegment(db, userId);
```
laverdet: This is a waterfall. Use Promise.all

packages/xxscreeps/mods/memory/model.ts
```diff
@@ -34,3 +41,55 @@ export async function saveMemorySegmentBlob(shard: Shard, userId: string, segmen
...
+	await db.data.hset(User.infoKey(userId), 'defaultPublicSegment', id === null ? '' : String(id));
+}
+
+export async function loadActiveForeignSegment(db: Database, userId: string): Promise<StoredForeignSegmentRequest | null> {
+	const raw = await db.data.hget(User.infoKey(userId), 'activeForeignSegment');
+	if (raw === null || raw === '') {
+		return null;
+	}
+	let parsed: unknown;
+	try { parsed = JSON.parse(raw); } catch { return null; }
```
laverdet: Store `segmentId` and `userId` in the same redis hash value. Then use hgetall to get the object, without this JSON.parse. For the most part we don't store JSON values in the database.

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -52,24 +94,46 @@ hooks.register('runnerConnector', player => {
 				activeSegments = nextSegments;
 				nextSegments = undefined;
 			}
-			if (foreignSegmentRequest) {
-				// TODO
+			if (activeForeignSegment !== null) {
```
laverdet: `if (activeForeignSegment) {`

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -52,24 +94,46 @@ hooks.register('runnerConnector', player => {
...
+				const { username, userId: targetUserId, id } = activeForeignSegment;
+				if (await isPublicSegment(shard, targetUserId, id)) {
```
laverdet: This is another waterfall. You should grab both keys unconditionally and then only assign if it's public.

The bigger problem is that you're fetching the memory each tick. It needs to use pubsub to avoid refetching each and every tick. This is very wasteful.

packages/xxscreeps/mods/memory/memory.ts
```diff
@@ -258,8 +303,30 @@ export function flushActiveSegments() {
...
+ * Returns the update from `RawMemory.setPublicSegments`
+ */
+export function flushPublicSegments() {
+	const tmp = requestedPublicSegments;
+	requestedPublicSegments = undefined;
+	return tmp ?? null;
+}
+
+export function loadForeignSegment(payload?: ForeignSegmentPayload) {
+	RawMemory.foreignSegment = payload;
```
laverdet: It shouldn't be updated every tick. Only when the foreign player updates it.

packages/xxscreeps/mods/memory/model.ts
```diff
@@ -34,3 +41,55 @@ export async function saveMemorySegmentBlob(shard: Shard, userId: string, segmen
...
+}
+
+function publicSegmentsKey(userId: string) {
+	return `user/${userId}/publicSegments`;
+}
+
+export async function savePublicSegments(shard: Shard, userId: string, ids: number[]) {
+	const key = publicSegmentsKey(userId);
+	const members = [ ...new Set(ids.filter(isValidSegmentId)) ].map(String);
+	await shard.data.del(key);
```
laverdet: Waterfall. `del` and `sadd` will be serialized on the redis end anyway. So you can just run both at the same time.

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -52,24 +94,46 @@ hooks.register('runnerConnector', player => {
...
+					const blob = await loadMemorySegmentBlob(shard, targetUserId, id);
+					if (blob !== null) {
+						const uint16 = new Uint16Array(blob.buffer, blob.byteOffset, blob.length >>> 1);
+						payload.foreignSegment = {
+							username,
+							id,
+							data: typedArrayToString(uint16),
```
laverdet: The memory code that this was copied from runs in the player's runtime, but you're running this in the driver. When at all possible work should be offloaded to the player's runtime.

The point is that you can make `RawMemory.foreignSegment` a lazy getter. The getter will only perform `typedArrayToString` if the user accesses it. Otherwise you just have a typed array sitting around which doesn't live in v8's heap.

packages/xxscreeps/mods/memory/model.ts
```diff
@@ -34,3 +41,55 @@ export async function saveMemorySegmentBlob(shard: Shard, userId: string, segmen
...
+	}
+	const { username, userId: storedUserId, id } = parsed as Record<string, unknown>;
+	if (typeof username === 'string' && typeof storedUserId === 'string' && typeof id === 'number') {
+		return { username, userId: storedUserId, id };
+	}
+	return null;
+}
+
+export async function saveActiveForeignSegment(db: Database, userId: string, request: StoredForeignSegmentRequest | null) {
+	await db.data.hset(User.infoKey(userId), 'activeForeignSegment', request === null ? '' : JSON.stringify(request));
```
laverdet: `resolveForeignRequest` should maybe just be rolled into this function.

laverdet (changes requested):
significant changes needed. pubsub is main thing that needs to be changed. you may want to split off the non-foreign related changes into another pr.

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -1,41 +1,66 @@
...
 		activeSegmentsRequest: number[] | null;
-		foreignSegmentRequest: null | {
-			id: number | undefined;
-			username: string;
-		};
+		foreignSegmentRequest: ForeignSegmentRequest | null | undefined;
 		memorySegmentsUpdated: SegmentPayload[] | null;
 		memoryUpdated: ReturnType<typeof flush>;
+		defaultPublicSegmentUpdate: number | null | undefined;
+		publicSegmentsUpdate: number[] | null;
```
laverdet: Here `null` means "no change" which is different than the other ones you've added.

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -1,41 +1,66 @@
...
 
 hooks.register('runnerConnector', player => {
 	const { shard, userId } = player;
 	let activeSegments: Set<number>;
 	let nextSegments: Set<number> | undefined;
-	let foreignSegmentRequest: TickResult['foreignSegmentRequest'] = null;
-	return [ undefined, {
+	let activeForeignSegment: StoredForeignSegmentRequest | null = null;
+	let subscribedUserId: string | null = null;
+	let subscription: SubscriptionFor<typeof getPublicSegmentChannel> | undefined;
```
laverdet: Actually, you want `let effect: Effect | undefined`. You don't need to store the current subscription, you need to store the disconnection handler. This pattern is used elsewhere in the codebase.

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -52,25 +77,57 @@ hooks.register('runnerConnector', player => {
...
+			// Refresh the foreign-segment cache only when the publisher notified us (or on first
+			// refresh after subscribe). Otherwise reuse the cached payload — no per-tick storage reads.
+			if (activeForeignSegment && foreignDirty) {
+				foreignDirty = false;
+				const { username, userId: targetUserId, segmentId } = activeForeignSegment;
+				const [ isPublic, blob ] = await Promise.all([
+					isPublicSegment(shard, targetUserId, segmentId),
+					loadMemorySegmentBlob(shard, targetUserId, segmentId),
+				]);
+				cachedForeign = isPublic && blob !== null
```
laverdet: You should assign it directly to `payload.foreignSegment` here. There's no need to store the blob in the driver once it has been transmitted to the player runtime.

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -52,25 +77,57 @@ hooks.register('runnerConnector', player => {
...
+				await syncForeignSubscription();
+			}
+
+			const segmentWrites = payload.memorySegmentsUpdated
+				? [ ...Fn.take(payload.memorySegmentsUpdated, kMaxActiveSegments) ]
+				: [];
+			const publicSegmentsChanged = payload.publicSegmentsUpdate !== null;
+
 			await Promise.all([
-				// Save primary memory blob
```
laverdet: Why delete comment?

packages/xxscreeps/mods/memory/model.ts
```diff
@@ -34,3 +55,93 @@ export async function saveMemorySegmentBlob(shard: Shard, userId: string, segmen
...
+		await shard.data.del(key);
+		return null;
+	}
+	let resolved: StoredForeignSegmentRequest;
+	if (prior?.username === request.username && request.id !== undefined) {
+		resolved = { username: request.username, userId: prior.userId, segmentId: request.id };
+	} else {
+		const targetUserId = await User.findUserByName(shard.db, request.username);
+		if (targetUserId === null) {
+			return prior;
```
laverdet: This should be `return null` right?

packages/xxscreeps/mods/memory/model.ts
```diff
@@ -34,3 +55,93 @@ export async function saveMemorySegmentBlob(shard: Shard, userId: string, segmen
...
+		}
+		const segmentId = request.id ?? await loadDefaultPublicSegment(shard, targetUserId);
+		if (segmentId === null) {
+			return prior;
```
laverdet: Same thing

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -52,25 +77,57 @@ hooks.register('runnerConnector', player => {
...
-			// Update activate segments
+			// Update active segments
 			if (payload.activeSegmentsRequest) {
 				nextSegments = new Set(Fn.take(payload.activeSegmentsRequest, kMaxActiveSegments));
 			}
-			foreignSegmentRequest = payload.foreignSegmentRequest;
+
+			const foreignRequest = payload.foreignSegmentRequest;
+			if (foreignRequest !== undefined) {
+				activeForeignSegment = await saveActiveForeignSegment(shard, userId, activeForeignSegment, foreignRequest);
```
laverdet: This should be dispatched with the Promise.all down below. Otherwise you have made another waterfall.

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -52,25 +77,57 @@ hooks.register('runnerConnector', player => {
...
+				...Fn.map(segmentWrites, segment => saveMemorySegmentBlob(shard, userId, segment.id, segment.payload)),
+				payload.defaultPublicSegmentUpdate !== undefined && saveDefaultPublicSegment(shard, userId, payload.defaultPublicSegmentUpdate),
+				publicSegmentsChanged && savePublicSegments(shard, userId, payload.publicSegmentsUpdate!),
 			]);
+
+			// Notify any foreign readers. Segment writes fire per-id so subscribers can filter;
+			// public-set changes fire once (subscribers must always re-check).
+			if (segmentWrites.length !== 0 || publicSegmentsChanged) {
+				const channel = getPublicSegmentChannel(shard, userId);
+				await Promise.all([
```
laverdet: Putting this outside the Promise.all causes another waterfall. It's fine to fire the update events at the same time as the updates themselves because reads are synchronized on tick. There is not a race.

laverdet: You can use the technique here if inline promises are becoming too unwieldy.

https://github.com/laverdet/xxscreeps/blob/dbbccf1bb2b70f8a079010f5b6c73fb509a7f34e/packages/xxscreeps/engine/db/storage/local/keyval.ts#L639-L645

packages/xxscreeps/mods/memory/memory.ts
```diff
@@ -83,7 +92,12 @@ export const RawMemory = {
 		if (typeof value !== 'string') {
 			throw new TypeError('Memory value must be a string');
 		}
-		RawMemory._parsed = json = undefined;
+		if (value.length > kMaxMemoryLength) {
+			throw new Error(`Reached maximum \`Memory\` limit. Requested: ${value.length} out of ${kMaxMemoryLength}`);
+		}
+		if (!accessedJson) {
+			RawMemory._parsed = json = undefined;
```
laverdet: I see `ok-screeps` doesn't have any memhack related commits. Well, the commit messages mention it but I don't see anything about it in the tests. Users will muck with `RawMemory._parsed` and expect certain behavior to follow. Deleting it means you don't care to save `Game.memory` this tick (unless RawMemory.set was called). Assigning `RawMemory._parsed` means you want that object to be serialized at the end of the tick.

Anyway, what about this (no conditional):
```
RawMemory._parsed = undefined;
```

The parsed `json` object would remain which maybe fixes `MEMORY-002`

packages/xxscreeps/mods/memory/memory.ts
```diff
@@ -254,14 +305,54 @@ export function flushActiveSegments() {
...
+}
+
+export function loadForeignSegment(payload?: ForeignSegmentPayload) {
+	if (!payload) {
+		RawMemory.foreignSegment = undefined;
+		return;
+	}
+	const { username, id, bytes } = payload;
+	let decoded: string | undefined;
+	RawMemory.foreignSegment = {
```
laverdet, line 342: Yeah you don't want to recreate this each tick. You're reseting the parsed data each tick even if the driver sent the same information as last tick.

packages/xxscreeps/mods/memory/model.ts
```diff
@@ -34,3 +55,93 @@ export async function saveMemorySegmentBlob(shard: Shard, userId: string, segmen
...
+	return resolved;
+}
+
+function publicSegmentsKey(userId: string) {
+	return `user/${userId}/publicSegments`;
+}
+
+export async function savePublicSegments(shard: Shard, userId: string, ids: number[]) {
+	const key = publicSegmentsKey(userId);
+	const members = [ ...new Set(ids.filter(isValidSegmentId)) ].map(String);
```
laverdet:
```ts
const members = Fn.pipe(
ids,
$$=>Fn.filter($$, isValidSegmentId),
$$=>new Set($$)
$$=>Fn.map($$, String)
$$=>[...$$]
```

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -52,24 +76,63 @@ hooks.register('runnerConnector', player => {
...
 				nextSegments = undefined;
 			}
-			if (foreignSegmentRequest) {
-				// TODO
+			// Only refetch and push to the runtime when the publisher notified us (or on target
+			// change). Otherwise the runtime holds last tick's payload and doesn't recreate the
+			// lazy-getter object.
+			if (foreignDirty) {
+				foreignDirty = false;
+				if (!activeForeignSegment) {
```
laverdet: prefer non-negated ifs

packages/xxscreeps/mods/memory/driver.ts
```diff
@@ -52,24 +76,63 @@ hooks.register('runnerConnector', player => {
...
 			}
-			foreignSegmentRequest = payload.foreignSegmentRequest;
+
+			const foreignRequest = payload.foreignSegmentRequest;
+			const segmentWrites = payload.memorySegmentsUpdated
+				? [ ...Fn.take(payload.memorySegmentsUpdated, kMaxActiveSegments) ]
+				: [];
+			const publicSegmentsChanged = payload.publicSegmentsUpdate !== undefined;
+			const channel = segmentWrites.length !== 0 || publicSegmentsChanged
+				? getPublicSegmentChannel(shard, userId)
```
laverdet: Acquiring a channel descriptor is pretty much free, there's no need to conditionalize it.

packages/xxscreeps/mods/memory/memory.ts
```diff
@@ -167,17 +201,20 @@ export function flush() {
 		return { size: 0 };
 	}
 
-	// Check for JSON-based `Memory` object
-	if (json) {
+	if (setCalled) {
+		// `RawMemory.set()` was called this tick — `string` is authoritative. Drop any cached
+		// `json` so the next tick's `Memory` access re-parses the new value rather than serving
+		// stale state.
+		setCalled = false;
```
laverdet: It's not correct. `RawMemory._parsed` overrides `RawMemory.set()`. It's basically the last thing the vanilla runtime does:

https://github.com/screeps/driver/blob/cf63d8adf902663e2ebddd7f8c5b7baa425dc928/lib/runtime/runtime.js#L246-L248

I know it sounds like I'm being picky but this codepath was the first thing I had to fix in order to get the open source bots working. The observable behavior should be as exact as possible. Note that doesn't mean copying line for line because we can avoid `JSON.parse(JSON.stringify())` passes in cases where vanilla doesn't bother.

laverdet (commented):
If possible I'd recommend splitting the `Memory` changes from the foreign segment stuff. Only minor issues remain with foreign segments but the undocumented `RawMemory._parsed` field is super sensitive and lots of players rely on very specific behavior.

misterwise:
Split per review — the RawMemory.set / `_parsed` changes now live in #140. This PR is force-pushed to a single commit with only the foreign and public memory segment work.

Also addressed the two remaining nits in `driver.ts`:
- `if (activeForeignSegment)` instead of `!activeForeignSegment` at the foreign-fetch branch.
- Acquire the `publicSegmentChannel` unconditionally and drop the `!`-asserts downstream.

PR #132, opened 2026-04-21
Add notifyWhenAttacked and missing structure surface getters

misterwise:
Four additive fixes bringing the public structure surface in line with vanilla. No behavior changes to existing paths.

1. **`Structure.prototype.notifyWhenAttacked(enabled)`** did not exist. Added on the `Structure` base with a `checkNotifyWhenAttacked` helper mirroring `@screeps/engine/src/game/structures.js:89` — reject if `my === false`, or if the room controller has an owner who isn't the caller, then require a boolean `enabled`. Persists via a new `'#noAttackNotify': 'bool'` field on the base structure shape; negated so the default-zero value matches vanilla's "notifications on" default. The intent processor is registered on `Structure` alongside `destroyStructure` and mirrors the `setPublic` pattern at `mods/defense/processor.ts:83`. Note: the flag is persisted but no consumer reads it yet — damage processors would need to be extended to emit attack notifications. A TODO at the write site flags this for a follow-up.

2. **`StructureWall.ticksToLive`** was missing. Vanilla exposes it via `defineGameObjectProperties` at `@screeps/engine/src/game/structures.js:840` — `undefined` for persistent walls, a number for NPC-spawned temporary ones. Added an `@enumerable` getter returning `undefined`; NPC-spawned temporary walls are not modeled here, so there's nothing to decrement.

3. **`Ruin.structureType`** was only reachable via the inner `ruin.structure` sub-object (`Object.defineProperty` at `ruin.ts:60`). Vanilla's canonical shape puts it on the ruin itself. Added a top-level `@enumerable get structureType()` delegating to `this['#structure'].type`.

4. **`StructureStorage.storeCapacity`** legacy compatibility getter was missing. Added `@deprecated @enumerable get storeCapacity()` returning `this.store.getCapacity()`, mirroring the `@deprecated energy`/`energyCapacity` pair on `StructureLab` at `mods/chemistry/lab.ts:36-38`.

Verified against ok-screeps.


packages/xxscreeps/mods/structure/structure.ts
```diff
@@ -75,6 +77,16 @@ export class Structure extends withOverlay(RoomObject, shape) {
...
 	}
 
+	/**
+	 * Toggle notifications for when this structure is attacked.
+	 * @param enabled Whether to receive email notifications on attack.
+	 */
+	notifyWhenAttacked(this: Structure, enabled: boolean) {
+		return chainIntentChecks(
+			() => checkNotifyWhenAttacked(this, enabled),
+			() => intents.save(this, 'notifyWhenAttacked', Boolean(enabled)));
```
laverdet: Vanilla has a check where they don't send the intent if the boolean already matches. It seems like a good one because I could imagine a player just invoking it every tick and sending junk intent each tick.

misterwise:
Addressed: `Structure.notifyWhenAttacked` now skips `intents.save` when the requested value already matches `#noAttackNotify` (opposite-polarity so the diff is a direct `===`). Also renamed the param from `enabled` to `notifyWhenAttacked` across the runner method, intent check, and processor for clarity. Merged `main` to resolve the `ruin.ts` conflict (placed `structureType` adjacent to the relocated `ticksToDecay` getter).

PR #133, opened 2026-04-21
Add Game.cpuLimit, Game.powerCreeps, Room.survivalInfo

misterwise:
xxscreeps was missing three data-properties vanilla exposes on `Game` and `Room`:

- **`Game.cpuLimit`** — now a getter deriving from `this.cpu.limit`.
- **`Game.powerCreeps`** — always `{}` and marked `@deprecated` in line with xxscreeps' existing deprecation markers on other legacy-compat surfaces.
- **`Room.survivalInfo`** — always `null` (xxscreeps has no survival mode).

Scope trimmed from the earlier version of this PR:

- The `id`-on-Flag restructure is deferred pending a separate discussion (per your note that flags are already architecturally separate from room state, so the id special-case should stay narrow to Flag).
- The `hits`/`hitsMax`/`my` base-leak cleanup is split into its own focused PR, which removes the getters rather than widening the base type via interface merge.

Verified against ok-screeps.

### DIFF
--- a/packages/xxscreeps/game/game.ts
+++ b/packages/xxscreeps/game/game.ts
@@ -75,6 +75,18 @@ export class Game extends GameBase {
 	declare cpu: CPU;
 
 	/**
+	 * Your assigned CPU limit for the current shard.
+	 * @deprecated
+	 */
+	get cpuLimit(): number { return this.cpu.limit; }
+
+	/**
+	 * A hash containing all your power creeps with their names as hash keys. xxscreeps does not
+	 * implement power creeps yet, so this is always an empty object.
+	 */
+	powerCreeps: Record<string, unknown> = Object.create(null);
+
+	/**
 	 * An object describing the world shard where your script is currently being executed in.
 	 */
 	shard: { name: string; type: string; ptr: boolean };
--- a/packages/xxscreeps/game/room/room.ts
+++ b/packages/xxscreeps/game/room/room.ts
@@ -31,6 +31,13 @@ export class Room extends withOverlay(BufferObject, shape) {
 	#removeObjects = new Set<RoomObject>();
 
 	/**
+	 * A [survival game](https://docs.screeps.com/survival.html) info, if the current game is
+	 * running in the Survival Mode. xxscreeps has no survival mode, so this is always `null`.
+	 */
+	// eslint-disable-next-line @typescript-eslint/class-literal-property-style
+	@enumerable get survivalInfo(): null { return null; }
+
+	/**
 	 * Find all objects of the specified type in the room. Results are cached automatically for the
 	 * specified room and type before applying any custom filters. This automatic cache lasts until
 	 * the end of the tick.


packages/xxscreeps/game/game.ts
```diff
@@ -74,6 +77,17 @@ export class Game extends GameBase {
 	 */
 	declare cpu: CPU;
 
+	/**
+	 * Your assigned CPU limit for the current shard.
+	 */
+	cpuLimit: number;
```
laverdet: I believe this is `@deprecated`

packages/xxscreeps/game/game.ts
```diff
@@ -84,6 +98,8 @@ export class Game extends GameBase {
 	constructor(state: GameState, data?: TickPayload) {
 		super(state);
 
+		this.cpuLimit = data?.cpu.limit ?? 0;
```
laverdet: This should just be a getter that returns `this.cpu.limit` or whatever

packages/xxscreeps/game/object.ts
```diff
@@ -97,6 +92,20 @@ export abstract class RoomObject extends withOverlay(BufferObject.BufferObject,
 	}
 }
 
+// Keep `hits`/`hitsMax`/`my`/`id` typable on the base (used by shared methods like
+// `#applyDamage`) without emitting prototype getters — subclasses opt in via their own schema.
+export interface RoomObject {
```
laverdet: If this is what we wanted to do, and I'm not sure it is, you would use `declare` in the class body above.

laverdet:
I'm not sure what this is trying to fix (the `id` stuff, which is unrelated to `survivalInfo` and `cpuLimit`). Flags are the only object in the game which don't get id's, so it's probably ok to special case them.

A getter returning `undefined` is more or less indistinguishable from an undefined property so idk what moving all that stuff around accomplished.

packages/xxscreeps/game/game.ts
misterwise: For deprecated we have two approaches, we could include and mark deprecated for now, until vanilla properly deprecates, or we could proactively just not support all deprecations. I personally lean towards removing deprecated functionality proactively but it would be good to be consistent there I think so we can mark the gaps as intentional.

misterwise:
> I'm not sure what this is trying to fix (the `id` stuff, which is unrelated to `survivalInfo` and `cpuLimit`). Flags are the only object in the game which don't get id's, so it's probably ok to special case them.
> 
> A getter returning `undefined` is more or less indistinguishable from an undefined property so idk what moving all that stuff around accomplished.

The goal was to clean up flag and remove properties that don't match vanilla, these got lumped together just because they are looking at the shape of objects in comparison to vanilla. I looked at base inheritance as well, but tend towards just making Flag a special case, it is a bit of an odd special case in Screeps as it is in rooms, but differs slightly from other RoomObjects. I looked at using declare but couldn't get it to work on the objects that needed them, but will take another look.

laverdet:
I think the missing piece needed to explain this is that flags do not affect game state. xxscreeps takes advantage of this and does not store them in the room. In vanilla flags are treated the same as any structure, which is of course crazy. It would be like having user code store in the room, doesn't make sense.

misterwise:
Rescoped this PR to just the Game/Room surface additions (`cpuLimit`, `powerCreeps`, `survivalInfo`). Incorporated your inline feedback on both cpuLimit points:

- `cpuLimit` — now an `@deprecated` getter deriving from `this.cpu.limit`.
- `powerCreeps` — descriptive comment noting xxscreeps doesn't implement power creeps yet
- `survivalInfo` — returns `null` since xxscreeps has no survival mode.

The `hits`/`hitsMax`/`my` work is split to #134, which takes a different approach than the interface-merge version you questioned — no base type widening at all, just a `this:` constraint on `#applyDamage`. The `id`-on-Flag work is deferred pending a separate conversation on special-casing Flag given its per-user-blob architecture.

packages/xxscreeps/game/game.ts
laverdet: Proactively dropping is the wrong call. `creep.carry` has been deprecated for like 10 years. They're not going to remove it. We need to support it and mark as deprecated internally.

Not only do we need to support deprecated behaviors, we need to support a lot of undocumented behaviors too. Off the top of my head there's RoomPosition: `__packedPos`, Store's `_sum` and CostMatrix's `_bits`. I didn't have to search long to find code that depends on all these features. Not supporting them means people won't use xxscreeps.

PR #134, opened 2026-04-22
Remove hits/hitsMax/my leak from RoomObject base

misterwise:
xxscreeps' `RoomObject` base defines `hits`/`hitsMax`/`my` as getters returning `undefined`. Vanilla installs these selectively on `Creep`/`Structure`/`ConstructionSite` only. The result is a prototype-shape divergence: `Source`, `Mineral`, `Resource`, `Tombstone`, `Ruin`, and `Flag` all expose those three entries on their prototypes in xxscreeps but not in vanilla. Invisible to value reads, observable via `'hits' in obj`, prototype walking, and any player-side type guards that rely on the property being genuinely absent.

The base getters existed only to let shared code like `#applyDamage` type-check `this.hits`. Rather than leave them on the prototype or introduce runtime scaffolding to compensate for their removal, this PR keeps them visible at the *type* level via a declaration-merged `export declare interface RoomObject` block placed after the class. The merge has zero runtime footprint, so consumer sites (`object.hits === undefined`, `structure.my`) keep their natural form — property access on non-destructible / non-ownable objects resolves to `undefined` naturally.

**Change:** `packages/xxscreeps/game/object.ts` only.

- Delete the base `hits`/`hitsMax`/`my` getters and the `hits` setter.
- Add a type-only `export declare interface RoomObject { get hits(); set hits(); get hitsMax(); get my(); }` merge after the class. No runtime effect; destructible/ownable subclasses continue to supply the real getters.

Split out from #133. Verified against ok-screeps.

### DIFF
--- a/packages/xxscreeps/game/object.ts
+++ b/packages/xxscreeps/game/object.ts
@@ -57,13 +57,9 @@ export abstract class RoomObject extends withOverlay(BufferObject.BufferObject, 
 	// eslint-disable-next-line @typescript-eslint/class-literal-property-style
 	get '#providesVision'() { return false; }
 	get '#user'(): string | null { return null; }
-	get hits(): number | undefined { return undefined; }
-	get hitsMax(): number | undefined { return undefined; }
-	get my(): boolean | undefined { return undefined; }
 	abstract get ['#lookType'](): string | null;
 
 	set '#user'(_user: string | null) { throw new Error('Setting `#user` on unownable object'); }
-	set hits(_hits: number | undefined) { throw new Error('Setting `hits` on indestructible object'); }
 
 	'#addToMyGame'(_game: GameConstructor) {}
 	'#afterInsert'(room: Room) {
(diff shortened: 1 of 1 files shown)


packages/xxscreeps/game/object.ts
```diff
@@ -74,8 +70,11 @@ export abstract class RoomObject extends withOverlay(BufferObject.BufferObject,
 		this.room = undefined as never;
 	}
 
-	'#applyDamage'(power: number, _type: number, _source?: RoomObject) {
-		if ((this.hits! -= power) <= 0) {
+	'#applyDamage'(
+		this: RoomObject & { hits: number; '#destroy'(): unknown },
```
laverdet: Break this inline type out into something like `interface DestructibleRoomObject extends RoomObject { hits: number; hitsMax: number }`. (or the interface probably won't be needed anyway per my comment above)

laverdet (changes requested):
I don't like that TypeScript is "wagging the dog" in this case. What you can do is add this to object.ts after RoomObject:

```ts
export declare interface RoomObject {
	get hits(): number | undefined;
	set hits(hits: number);
	get hitsMax(): number | undefined;
}
```

Yes, you export it twice don't worry about it.

Then you don't need the type guard in TypeScript and those properties don't show up in JavaScript.

misterwise:
Came full circle, this was also just a simple declare to resolve it in the end.

PR #135, opened 2026-04-23
Fix renewCreep busy/boost guards and spawnCreep check order

misterwise:
Three independent mismatches in mods/spawn/ with the canonical engine.

checkRenewCreep was missing ERR_BUSY. Vanilla's user-side renewCreep short-circuits on this.spawning before any target validation. xxscreeps went straight to checkIsActive and target checks, so renewing through a mid-spawn returned whatever downstream check produced first instead of ERR_BUSY.

checkRenewCreep rejected boosted creeps with ERR_NO_BODYPART. Vanilla accepts them and the processor strips the boosts. Removed the boost-presence branch from the check and moved the strip into the renewCreep processor: clear each part.boost and call the new shared helper (see below) to reconcile store capacity.

checkSpawnCreep ran checkIsActive before the name-collision check. With a duplicate name against an RCL-inactive spawn, vanilla returns ERR_NAME_EXISTS (in structures.js the name branch sits above the off branch). xxscreeps was returning ERR_RCL_NOT_ENOUGH. Moved checkIsActive after the name / directions / busy closure.

While here, extracted the overflow-drop half of recalculateBody (mods/creep/processor.ts) as a new exported helper dropOverflowResources. It already did exactly what renewCreep needed after a boost strip — capacity recompute via calculateCarry and drop excess resources at the creep's position — and was duplicated inline in the first pass of this PR. Both call sites now share it. Stripping boosts off of a creep on renew is deprecated but still functional behaviour in vanilla so left a deprecation note as well.

Verified against ok-screeps.

### DIFF
--- a/packages/xxscreeps/mods/spawn/spawn.ts
+++ b/packages/xxscreeps/mods/spawn/spawn.ts
@@ -144,8 +144,10 @@ export class StructureSpawn extends withOverlay(OwnedStructure, shape) {
 	 * square. The target should not have CLAIM body parts. The spawn should not be busy with the
 	 * spawning process. Each execution increases the creep's timer by amount of ticks according to
 	 * this formula: `floor(600 / body.length)`. Energy required for each execution is determined
-	 * using this formula: `ceil(creepCost / 2.5 / body.length)`. Renewing a creep removes all of its
-	 * boosts.
+	 * using this formula: `ceil(creepCost / 2.5 / body.length)`.
+	 *
+	 * Renewing a creep removes all of its boosts. Calling this on a boosted creep is deprecated
+	 * in the official API — prefer `StructureLab.unboostCreep` first.
 	 * @param creep The target creep object.
 	 */
 	renewCreep(creep: Creep) {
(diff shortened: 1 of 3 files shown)


packages/xxscreeps/mods/spawn/spawn.ts
```diff
@@ -144,8 +144,10 @@ export class StructureSpawn extends withOverlay(OwnedStructure, shape) {
 	 * square. The target should not have CLAIM body parts. The spawn should not be busy with the
 	 * spawning process. Each execution increases the creep's timer by amount of ticks according to
 	 * this formula: `floor(600 / body.length)`. Energy required for each execution is determined
-	 * using this formula: `ceil(creepCost / 2.5 / body.length)`. Renewing a creep removes all of its
-	 * boosts.
+	 * using this formula: `ceil(creepCost / 2.5 / body.length)`.
+	 *
+	 * Renewing a creep removes all of its boosts. Calling this on a boosted creep is deprecated
```
laverdet, line 149: That message has been there since 2019. I wonder if they'll ever remove it =o

misterwise: Yeah that struck me as well, it is written so "imminently" lol

PR #136, opened 2026-04-23
Reclaim body energy and divert into containers on creep death

misterwise:
buryCreep previously scaled the creep's carried store by dropRate and reclaimed no body energy, diverging from vanilla _die. Rewrite to mirror vanilla: body energy at rate * ticksToLive / lifeTime (CLAIM-aware, CREEP_PART_MAX_ENERGY-capped) plus LAB_BOOST_MINERAL and LAB_BOOST_ENERGY refunds per boosted part, carried store at full amount, and route every deposit through a same-tile container before tombstone remainder. The chemistry constants are read through the ambient C aggregator (same pattern as the existing C.BOOSTS consumption in creep.ts), so no new mod import edge.

Also fix a race in recalculateBody: when a creep's hits reach zero this tick, skip the capacity-shrink overflow drop so the store survives for buryCreep to transfer.

Adds a Tombstone-store test exercising suicide with a boosted body part.

Verified against ok-screeps.

### DIFF
--- a/packages/xxscreeps/mods/creep/tombstone.ts
+++ b/packages/xxscreeps/mods/creep/tombstone.ts
@@ -1,8 +1,10 @@
+import type { ResourceType } from 'xxscreeps/mods/resource/index.js';
 import * as Id from 'xxscreeps/engine/schema/id.js';
 import * as C from 'xxscreeps/game/constants/index.js';
 import { Game } from 'xxscreeps/game/index.js';
 import { RoomObject, create as createObject, getById, format as objectFormat } from 'xxscreeps/game/object.js';
 import { OpenStore, openStoreFormat } from 'xxscreeps/mods/resource/store.js';
+import { lookForStructureAt } from 'xxscreeps/mods/structure/structure.js';
 import { compose, declare, enumerated, optional, struct, variant, vector, withOverlay } from 'xxscreeps/schema/index.js';
 import { assign } from 'xxscreeps/utility/utility.js';
 import { Creep } from './creep.js';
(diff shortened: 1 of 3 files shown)


packages/xxscreeps/mods/creep/tombstone.ts
```diff
@@ -69,9 +71,41 @@ export function buryCreep(creep: Creep, rate = C.CREEP_CORPSE_RATE) {
...
 	tombstone.deathTime = Game.time;
 	tombstone.store = new OpenStore();
-	for (const [ resourceType, amount ] of creep.store['#entries']()) {
-		tombstone.store['#add'](resourceType, Math.floor(amount * rate));
+
+	if (rate > 0) {
+		const lifeTime = creep.body.some(part => part.type === C.CLAIM)
+			? C.CREEP_CLAIM_LIFE_TIME : C.CREEP_LIFE_TIME;
+		const lifeRate = rate * (creep.ticksToLive ?? 0) / lifeTime;
+		const bodyResources = new Map<ResourceType, number>([ [ C.RESOURCE_ENERGY, 0 ] ]);
```
laverdet: is there a point to initialize energy to 0? you should either remove this or assert the get(C.energy) calls later.

packages/xxscreeps/mods/creep/tombstone.ts
```diff
@@ -69,9 +71,41 @@ export function buryCreep(creep: Creep, rate = C.CREEP_CORPSE_RATE) {
...
+					(bodyResources.get(C.RESOURCE_ENERGY) ?? 0) + C.LAB_BOOST_ENERGY * lifeRate);
+			}
+			bodyResources.set(C.RESOURCE_ENERGY, (bodyResources.get(C.RESOURCE_ENERGY) ?? 0) +
+				Math.min(C.CREEP_PART_MAX_ENERGY, C.BODYPART_COST[part.type] * lifeRate));
+		}
+		// Same-tile container absorbs resources before the tombstone, matching vanilla.
+		const container = lookForStructureAt(creep.room, creep.pos, C.STRUCTURE_CONTAINER);
+		const deposit = (type: ResourceType, amount: number) => {
+			if (amount <= 0) return;
+			if (container !== undefined && container.hits > 0) {
```
laverdet, line 93: do we need to check container.hits?

misterwise:
Restructured to make the deposit order explicit instead of relying on Map iteration:

- Split the accumulator into `bodyEnergy` (scalar) + `bodyBoosts` (Map), then deposit energy first, then iterate boosts. Vanilla `_die.js:42,56` gets the same ordering via `{energy:0}` + object enumeration, but here it's load-bearing at the call site and needs no comment. CREEP-DEATH-012 exercises the ordering (body energy must fill the container before any boost mineral does).

- Left `container.hits > 0`. Vanilla `_die.js:59,77` keeps it to handle the same-tick destruction race: damage intents run before the structure-cleanup pass, so a container attacked/dismantled to 0 hits this tick is still in `roomObjects` when the creep dying on the same tile reaches `_die.js`. Depositing into it would feed a structure that's about to be swept. xxscreeps has the same tick shape (damage processor → death → cleanup), so the guard carries over.

packages/xxscreeps/mods/creep/tombstone.ts
misterwise: Yes, for ordering of what goes into the container, this matches Vanilla.

laverdet:
Ah I guess it is the difference of the resource ending up in a ruin vs a tombstone.

PR #137, opened 2026-04-23
Feature: Add admin CLI for server operations

misterwise:
This is a fairly major update to the CLI, avoiding the need for render, and I want to stop and submit what I have before going too much further. I have tested the surface of it manually and like where it is right now. There is more that could be added and some next items would be exportWorld, random room generation, proper invalidation of stale-sessions in the client, support for writing to memory segments, more fine-tuned tick control (advance N ticks). I want to keep the impact on core systems limited in each PR because of the risk of my misunderstanding or poorly implementing internal support.

## Summary
- Ships a mod that exposes a structured command schema over a Unix domain socket, reachable via three client modes: `xxscreeps` (interactive JavaScript REPL), `xxscreeps admin <group> <command>` (one-shot, flag-based, script-friendly), and `xxscreeps offline` (direct DB access, no server).
- Adds command groups for the common server tasks: `users`, `bots`, `map`, `rooms`, `shards`, `system`, plus `auth.setPassword` from the password mod.
- Adds a cross-process cache-invalidation channel so out-of-band writes (CLI, scripts) don't leave stale caches in the processor, runner, and backend.

## Feature — new code

All of this is additive under `packages/xxscreeps/mods/cli/*`:

| Area | Files |
|---|---|
| Socket server + per-connection sandbox | `cli.ts`, `commands.ts`, `sandbox.ts`, `socket.ts`, `launcher.ts`, `symbols.ts` |
| Admin CLI client (flag translation + bash completion) | `admin/{main,argv,help,socket}.ts` |
| Offline standalone client | `standalone.ts` |
| REPL client | `packages/xxscreeps/config/repl.ts` (was previously empty) |
| `auth.setPassword` | `packages/xxscreeps/mods/backend/password/{cli,model}.ts`, refactored `backend.ts`/`index.ts` |
| `importWorld` wrapper callable from the sandbox | `packages/xxscreeps/scripts/import.ts` |
| Tests | `mods/cli/{test,test-socket,test-setup}.ts` |

Socket handshake protocol: first message accepts optional `{ shard }`; subsequent messages are `{ expression }`. Sandbox lifetime equals socket lifetime; variables persist between commands on the same connection. Destructive commands gate on `--force`; pause-required commands auto-wrap in `pauseSimulation`/`resumeSimulation` inside a single admin socket lifetime.

## Changes to existing runner/processor/engine code

Each item is called out separately so reviewers can evaluate them independently of the CLI feature.

### 1. Graceful shutdown contract — `engine/service/{launcher,main}.ts`
Previously `main.ts` handled `SIGINT` and published `shutdown` on the service channel. That races with the launcher (which owns the CLI socket and needs to unlock the game mutex before main exits). SIGINT ownership moved to `launcher.ts`; `main.ts` now only *listens* for `shutdown` on the service channel. `main.ts` also gained three bail-outs so a shutdown arriving mid-tick doesn't advance `shard.time` past unrun work:
- break out of the init loop on `shutdown`
- re-check `halted` inside the mutex scope (shutdown can land while blocked on mutex acquire — e.g., CLI `importWorld` wipes scratch under a pause)
- wrap the tick-wait `for await` in `Async.breakable` so a mid-tick shutdown interrupts the `intentAbandonTimeout` wait

### 2. `launcher` hook — `engine/service/{launcher.ts,symbols.ts}` (new)
New hook kind `launcher(db, shard) => void | Promise<void>`. Mods registering under `provides: ['launcher']` run inside the launcher process with shared db/shard connections. The CLI socket server is the only consumer today. Added empty-world startup warning (detects terrain-without-rooms as a failed previous import).

### 3. Cross-process cache invalidation — `engine/service/invalidation.ts` (new) + consumers
New `channel/invalidation` with three message types: `room`, `world`, `accessibleRooms`. Each long-lived service drains at a safe point:

- **`engine/service/processor.ts`** — queue invalidations and drain at the top of the main loop (inline with processor-channel messages to preserve FIFO); forward as `invalidate` requests to room workers.
- **`engine/processor/worker.ts`** — new `ProcessorRequest` case `invalidate`: `target: 'room'` drops the cached `Room`; `target: 'world'` clears caches *and* reloads the terrain blob since `World` has it baked in.
- **`engine/service/runner.ts`** — on `world` invalidation, rebind `world`, reload terrain, disconnect all `PlayerInstance`s (their sandboxes bake in `terrainBlob`).
- **`backend/server.ts`** — drop per-request terrain cache + `reloadWorld()` on `world`/`accessibleRooms`; evict a single room on `room`.

Intent-pipeline writers don't publish here — they mutate the cached `Room` in place.

### 4. Room-status carries the active-rooms set — `engine/db/shard.ts`, `game/map.ts`, `backend/context.ts` + endpoint consumers
`Shard.loadWorld` now also fetches `rooms` smembers and passes it as a `ReadonlySet<string>` to `new World(...)`, which forwards it to `GameMap`. `GameMap.getRoomStatus(roomName)` returns `'out of borders'` when a terrain-backed room isn't in the active set (was hard-coded `'normal'`).

Backward compatibility: callers that build `World` without the set (sandbox init, processor worker) keep seeing `'normal'` — the set is optional.

`BackendContext` no longer holds a separate `accessibleRooms` field; `world` is now mutable with a `reloadWorld()` helper. Consumers updated:
- `backend/endpoints/game/map-stats.ts` — skip blob load for non-`normal` rooms; return `{ status }` only
- `backend/endpoints/game/room.ts` — `/api/game/room-status` consults `getRoomStatus` instead of returning static `'normal'`
- `backend/sockets/map.ts` — subscription filter uses `world.accessibleRooms`

### 5. Backend terrain cache eviction — `backend/endpoints/game/terrain.ts`
Exposed two eviction functions (`evictTerrainCacheEntry`, `clearTerrainCache`) called by `backend/server.ts`'s invalidation subscriber. Added `Cache-Control: no-cache` on the terrain response so clients revalidate via ETag after a terrain mutation.

### 6. Concurrent-save serialization — `engine/db/storage/local/{blob,keyval}.ts`
Added a `saveChain: Promise<void>` that serializes `save()` callers. Necessary because the CLI triggers two concurrent saves on shutdown: one from the CLI's `importWorld` and one from `backendContext.close()`. Previously both would race on the same tmp-file, and the loser's `rename()` could replace a freshly written file with stale content. The chain `.catch`es internally so a transient failure doesn't poison subsequent saves; the caller-facing promise still rejects normally.

### 7. `hdel` + expires-entry fix — `engine/db/storage/local/keyval.ts`, `engine/db/storage/provider.ts`, `packages/redis/keyval.ts`
Added `hdel(key, fields) → Promise<number>` to the `KeyValProvider` interface (CLI needs it for user removal). Local implementation removes the empty hash when the last field is dropped — fixes a pre-existing bug where drained hashes left orphan entries in the expires set. Also tightened `hgetall`'s return type to `Record<string, string | undefined>`.

### 8. `sleepingRoomsKey` exported — `engine/processor/model.ts`
The constant was already defined but not exported. Exporting it so `map.removeRoom` can `zrem` the room from both `activeRoomsKey` and `sleepingRoomsKey` — otherwise a re-added room would immediately be classified as sleeping and miss its first process cycle.

### 9. `seedTestShard` extracted — `test/import.ts`
Split the seeding logic out of `instantiateTestShard` so the CLI's socket test harness can re-seed a shared shard between test blocks without rebuilding the whole local DB.

### 10. Entry dispatch tweak — `config/entry.ts`
Added `offline` and `admin` subcommand entries. Also: a flag-prefixed first arg (e.g. `xxscreeps --shard shard1`) now falls through to the REPL path instead of erroring with `Invalid command "--shard"`.

### 11. Pre-existing-bug TODO — `engine/service/processor.ts`
Added a `TODO` in `consumeRoomsQueue` documenting a latent bug: `consumeSortedSet(..., 0, 0)` only pops score-0 rooms, but `updateUserRoomRelationships` scores `activeRoomsKey` by intent-capable user count, so spawn-owning rooms are permanently skipped. Masked by the current test seeds (no owned spawns); surfaces on cold-boot of bot-owned worlds. Left the TODO as a pointer — the fix is out of scope here.

## Verification
- `pnpm run build` clean.
- `pnpm run test` green: 354 tests passing. Coverage includes the sandbox evaluator, every registered command (`users`, `bots`, `map`, `rooms`, `shards`, `system`, `auth`), flag translation in the admin CLI, and cross-process smoke tests that spawn `xxscreeps start`, then drive the live launcher through the REPL socket, `xxscreeps admin`, and `xxscreeps offline` end-to-end (including `bots.add` with `--code-dir`/`--package`, admin `--help` listing every group, `importWorld` → shutdown → restart).
- `pnpm run eslint` on the changed files: zero new warnings (repo baseline is ~1400 pre-existing).


packages/xxscreeps/backend/endpoints/game/map-stats.ts
```diff
@@ -15,17 +15,22 @@ export const MapStatsEndpoint: Endpoint = {
...
+				return [ roomName, { status: status.status } ] as const;
+			}
 
 			// TODO: A room status blob that doesn't change every tick would be good
 			const room = await context.backend.shard.loadRoom(roomName, undefined, true);
 
 			// Build rooms payload
 			return [ room.name, {
-				status: 'normal',
+				status: status.status,
```
laverdet, line 33: you know it's normal because you just checked it

packages/xxscreeps/backend/endpoints/game/terrain.ts
```diff
@@ -9,6 +9,19 @@ const cache = new Map<string, {
...
 }>();
+
+/** Evict a single room from the terrain payload cache. Called from the
+ * backend's invalidation subscriber on a per-room signal. */
+export function evictTerrainCacheEntry(roomName: string) {
+	cache.delete(roomName);
+}
+
+/** Drop every cached terrain payload. Called when the world terrain blob
+ * itself has changed (import-world, remove-room). */
```
laverdet, line 20: I know it's nitpicky but I don't like the doc block style here. You can do `/** one liners */` if you want but for multiple lines please break it like all the other comments in the repo.

packages/xxscreeps/backend/endpoints/game/terrain.ts
```diff
@@ -42,6 +55,8 @@ hooks.register('route', {
 		const roomName = `${context.query.room}`;
 		const terrain = getTerrainPayload(context.backend.world, roomName);
 		if (terrain) {
+			// Force revalidation; ETag keeps the unchanged case at 304.
+			context.set('Cache-Control', 'no-cache');
```
laverdet, line 59: I'm not a big fan. At least cache the etag please.

packages/xxscreeps/config/repl.ts
```diff
@@ -1,72 +1,205 @@
...
+import fs from 'node:fs';
+import net from 'node:net';
+import os from 'node:os';
+import path from 'node:path';
+import readline from 'node:readline';
+import { maxBufferSize, socketPath } from 'xxscreeps/mods/cli/socket.js';
+
+// Minimal arg parsing — only `--shard <name>` is recognized. The unused
+// positionals are ignored so the rest of the REPL path is unchanged.
+function parseShardFlag(argv: readonly string[]): string | undefined {
```
laverdet, line 10: There is already a function for this `checkArguments`

packages/xxscreeps/engine/db/storage/local/keyval.ts
```diff
@@ -170,6 +171,17 @@ export class LocalKeyValResponder implements MaybePromises<P.KeyValProvider> {
 		return value;
 	}
 
+	hdel(key: string, fields: string[]) {
+		const map = this.data.get(key) as Map<string, string> | undefined;
+		if (!map) return 0;
+		let removed = 0;
```
laverdet, line 177: `const removed = Fn.accumulate(fields, field => map.delete(field) ? 1 : 0);`

packages/xxscreeps/engine/db/storage/local/keyval.ts
```diff
@@ -279,9 +291,10 @@ export class LocalKeyValResponder implements MaybePromises<P.KeyValProvider> {
 	}
 
 	sinter(key: string, keys: string[]) {
+		const allKeys = [ key, ...keys ];
 		const sets = Fn.pipe(
-			Fn.concat([ [ key ], keys ]),
-			$$ => Fn.map($$, (key): Set<string> | undefined => this.data.get(key)),
+			allKeys,
```
laverdet, line 296: Why is this change here? It looks functionally identical?

packages/xxscreeps/engine/db/storage/local/keyval.ts
```diff
@@ -279,9 +291,10 @@ export class LocalKeyValResponder implements MaybePromises<P.KeyValProvider> {
...
+			$$ => Fn.map($$, (entry): Set<string> | undefined => this.data.get(entry)),
```
laverdet, line 297: Also "entry" means key + value. What you have there is a key.

packages/xxscreeps/scripts/import.ts
```diff
@@ -0,0 +1,352 @@
...
+// what the rest of the engine expects.
+const LAUNCHER_ADMIN_USER_ID = 'f4b532d08c3952a';
+
+function forUser(userId: string | null) {
+	return userId === LAUNCHER_ADMIN_USER_ID ? '1' : userId;
+}
+
+// Union of every `rooms.objects` field — the loader switches on `type` to
+// narrow. Optional across the board to avoid per-type interfaces.
+interface RawObject {
```
laverdet, line 49: This should be a discriminated union if you're trying to type this file.

packages/xxscreeps/mods/cli/admin/argv.ts
```diff
@@ -0,0 +1,157 @@
...
+				flags.set(arg.slice(2, eq), arg.slice(eq + 1));
+			}
+			continue;
+		}
+		positionals.push(arg);
+	}
+	return { globals, positionals, flags };
+}
+
+function coerceValue(raw: string, kind: CommandArgField['kind'] | 'object' | 'callback', where: string): unknown {
```
laverdet, line 78: This whole file is slop. There's npm modules that do all this. argparse is already used in the codebase but since then I've become partial to yargs.

laverdet:
Hey man I can't review all this. This needs to be like 8 different diffs at least.

Other thoughts:
- "offline" cli mode only makes sense in the case of the local storage provider. On a redis backend you would talk directly to the database and it doesn't matter if the main loop is running or not. The cli should detect this.
- This terrain invalidation stuff is a can of worms. Just restart the server if the terrain is modified.

misterwise:
> Hey man I can't review all this. This needs to be like 8 different diffs at least.
> 
> Other thoughts:
> 
> * "offline" cli mode only makes sense in the case of the local storage provider. On a redis backend you would talk directly to the database and it doesn't matter if the main loop is running or not. The cli should detect this.
> * This terrain invalidation stuff is a can of worms. Just restart the server if the terrain is modified.

Yes, I had the feeling this was getting too large to digest. I will split this into much smaller PRs, focusing on the core changes to support CLI functionality first and then getting to the actual features themselves so we can review and discuss separately. I'll integrate the feedback you've already provided.

PR #140, opened 2026-04-24
Fix RawMemory.set guards: eager 2 MB throw + preserve in-tick Memory mutations

misterwise:
Split from #131 per review. Three narrow `RawMemory.set` / `Memory` correctness fixes in `mods/memory/`.

## Changes

1. **Eager 2 MB throw.** Length check moves from `flush()` into `set()` so a user-code `try/catch` actually sees it.

2. **Self-replacing `Memory` + in-tick mutation preservation.** `runtimeConnector.receive` installs `Memory` as a self-replacing lazy getter — first access flips the descriptor to a value property, matching `@screeps/engine/src/game/game.js`. `set()` clears `_parsed` only and sets a `setCalled` flag; cached `json` survives the call so per-object `.memory` reads (which go through `get()`, not `globalThis.Memory`) stay pinned to the originally-parsed object in-tick.

   `flush()` precedence matches vanilla: `_parsed` > `setCalled`-saves-string > memhack-skip > fall-through. Per-object `.memory` accessors continue to call the module-level `get()` directly (no `declare global`).

3. **Cross-tick parse refresh.** Drop cached `json` after every `flush()` save path so the next tick's first `Memory` access re-parses raw — values that `JSON.stringify` coerces (functions stripped, `NaN`/`Infinity` → `null`) round-trip to those coerced forms instead of surviving in the live object across ticks.

## Validation

Verified against screeps-ok: `MEMORY-002`/`004`/`005`/`006`, `UNDOC-MEMHACK-007`/`008`/`009`/`010`/`011`/`012`, and `UNDOC-MEMJSON-001`/`003`/`004`. xxscreeps `pnpm run test` 177/177.

### DIFF
--- a/packages/xxscreeps/mods/memory/game.ts
+++ b/packages/xxscreeps/mods/memory/game.ts
@@ -44,11 +44,19 @@ hooks.register('runtimeConnector', {
 	receive(payload) {
 		loadSegments(payload.memorySegments);
 		loadForeignSegment(payload.foreignSegment);
-		// Redefine memory each tick, expected behavior from vanilla server
+		// Self-replacing getter — first access pins Memory to a value descriptor (vanilla parity).
 		Object.defineProperty(globalThis, 'Memory', {
 			configurable: true,
 			enumerable: true,
-			get,
+			get() {
+				const value: unknown = get();
+				Object.defineProperty(globalThis, 'Memory', {
+					configurable: true,
+					enumerable: true,
+					value,
+				});
+				return value;
+			},
 		});
 	},
 
--- a/.changeset/raw-memory-set-guards.md
+++ b/.changeset/raw-memory-set-guards.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Fix `RawMemory.set` guards: eager 2 MB throw, in-tick `Memory` mutation preservation, cross-tick parse refresh
--- a/packages/xxscreeps/mods/memory/memory.ts
+++ b/packages/xxscreeps/mods/memory/memory.ts
@@ -17,8 +17,8 @@ let requestedPublicSegments: number[] | undefined;
 let memory: Uint16Array;
 let memoryLength = 0;
 let string: string | undefined;
-let accessedJson = false;
 let json: object | undefined;
+let setCalled = false;
 let isBufferOutOfDate = false;
 
 function align(address: number) {
@@ -92,7 +92,12 @@ export const RawMemory = {
 		if (typeof value !== 'string') {
 			throw new TypeError('Memory value must be a string');
 		}
-		RawMemory._parsed = json = undefined;
+		if (value.length > kMaxMemoryLength) {
+			throw new Error(`Reached maximum \`Memory\` limit. Requested: ${value.length} out of ${kMaxMemoryLength}`);
+		}
+		// Keep `json` so in-tick `Memory` reads stay pinned. `setCalled` separates this from a `delete _parsed` memhack-skip.
+		RawMemory._parsed = undefined;
+		setCalled = true;
 		string = value;
 		isBufferOutOfDate = true;
 	},
@@ -171,7 +176,6 @@ export const RawMemory = {
  * `Game.memory` getter
  */
 export function get(): any {
-	accessedJson = true;
 	if (json) {
 		return json;
 	}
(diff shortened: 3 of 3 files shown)


misterwise:
Good catch — you're right, `_parsed` overrides `set()` and the previous `setCalled` branch inverted that. Pushed 65bf5d6.

**Changes**

- `memory.ts`
  - `set()` now clears both `_parsed` and `json` (was only clearing `_parsed`). The cached `json` would otherwise shadow the new string on a post-`set()` read.
  - Removed `setCalled` entirely.
  - `flush()` follows vanilla's rule in order: `if (_parsed)` serialize it (crunch when `=== json`, else replace `json`); else if `json` cached, memhack-skip (restore `_parsed = json`, no save); else fall through so the dirty-buffer path persists `set()`'s raw string. `_parsed` > memhack-skip > set().
- `game.ts`
  - `Memory` binding now self-replaces on first access (matches `@screeps/engine/src/game/game.js`). Subsequent accesses return the same reference without re-firing the getter, which is what gives vanilla its "in-tick Memory reads stay stale after `set()`" behavior by construction — no flag needed.

**Traces I verified against vanilla** (where `V` = parsed pre-set, `parsedFoo` = parsed post-set):

| sequence | vanilla | this PR |
|---|---|---|
| `set` | saves `"foo"` | ✓ |
| `set → access → mutate` | saves mutated `parsedFoo` | ✓ (`_parsed = parsedFoo`) |
| `access → mutate → set` | saves `"foo"`, mutation lost | ✓ (`_parsed` cleared by `set`) |
| `access → mutate → set → access → mutate` | saves `"foo"`, both mutations lost | ✓ (self-replace keeps `Memory` at `V`, no re-fire, `_parsed` stays cleared) |
| `access → memhack skip` | no save | ✓ |
| `access → memhack replace` | saves custom | ✓ |

Verified against ok-screeps.

packages/xxscreeps/mods/memory/game.ts
```diff
@@ -43,11 +37,21 @@ hooks.register('runtimeConnector', {
...
+		// Vanilla-exact Memory binding: first access parses + self-replaces as a value property,
+		// so post-access `RawMemory.set()` does not invalidate the in-tick reference. Matches
+		// `@screeps/engine/src/game/game.js`.
 		Object.defineProperty(globalThis, 'Memory', {
 			configurable: true,
 			enumerable: true,
-			get,
+			get() {
+				const value: unknown = get();
+				Object.defineProperty(globalThis, 'Memory', {
```
laverdet, line 53: I don't think this will work with `creep.memory`, `flag.memory`, `room.memory`, and `spawn.memory`? These getters invoke the exported `get()` directly.

laverdet (changes requested):

packages/xxscreeps/mods/memory/game.ts
misterwise: I'm going to see if I can write some tests to more fully capture this as you're right it's a crucial undocumented feature.

packages/xxscreeps/mods/memory/game.ts
```diff
@@ -28,14 +34,35 @@ declare module 'xxscreeps/engine/runner/index.js' {
...
 	},
 });
 
+// Vanilla-exact Memory binding: first access parses + self-replaces as a value property,
+// so post-access `RawMemory.set()` does not invalidate the in-tick reference. Matches
+// `@screeps/engine/src/game/game.js`. Installed in `gameInitializer` (not `runtimeConnector`)
+// so the binding is present in every per-user game context — driver runtime *and* the
+// processor / test harness paths that enter via `runForUser` without a `TickPayload`.
+hooks.register('gameInitializer', () => {
+	Object.defineProperty(globalThis, 'Memory', {
```
laverdet: This should only live in the runtime.

laverdet:
Definitely don't do the global thing.

laverdet:
I think what you had in c1ececd433d7d151e85600d8be4d1d619301271e was close to right besides the memhack thing.

misterwise:
> I think what you had in [c1ececd](https://github.com/laverdet/xxscreeps/commit/c1ececd433d7d151e85600d8be4d1d619301271e) was close to right besides the memhack thing.

Ok, I think the regression tests are pretty strong now, and it took me some time to understand what you were trying to do with caching that improved upon vanilla, but also made it more complex to troubleshoot. Let me know if I've still missed something.

laverdet:
Closing in favor of 32c9fdb39f31a2cf9b65f31b029110c805aa7f53

PR #145, opened 2026-04-25
fix: Destroy dismantled structures, spill container contents on decay

misterwise:
Three fixes tying structure destruction to vanilla's observable behavior.

1. `mods/construction/processor.ts` — dismantle left a target at `hits=0` in place. Vanilla's `dismantle.js:52` routes through `_damage.js:46-49`, which invokes `structures/_destroy.js` on any non-creep that reaches zero hits — creating a ruin and removing the object. Added `target['#destroy']()` when `target.hits <= 0`; the existing `Structure['#destroy']()` override at `mods/structure/structure.ts:82` already handles ruin creation and removal.

2. `mods/resource/processor/container.ts` — container decay removed the container without spilling its store. Vanilla's `containers/tick.js:14-20` iterates `object.store` and routes each resource through `_create-energy` before `bulk.remove`. Mirror that: iterate `container.store['#entries']()` and call `drop()` per entry, then `#removeObject`.

3. `mods/resource/processor/resource.ts` — `drop()`'s container-catch was not gated on `container.hits > 0`. Vanilla's `_create-energy.js:22` already has the gate. Without it, the container-spill loop above routes resources back into the same container that's being removed, because `#removeObject` only queues removal for a later flush — `lookForStructureAt` still finds the dead container within the same tick.

The one thing that I wonder about is the use of drop and then checking for a dead container inside drop. We could have a dropOnGround that drop uses to make the code a bit cleaner?

Verified against ok-screeps.

### DIFF
--- a/packages/xxscreeps/mods/construction/processor.ts
+++ b/packages/xxscreeps/mods/construction/processor.ts
@@ -76,7 +76,10 @@ const intents = [
 					Resource.drop(creep.pos, 'energy', overflow);
 				}
 				target.hits -= effect;
-				// TODO: dismantle event + destroy hook
+				if (target.hits <= 0) {
+					target['#destroy']();
+				}
+				// TODO: dismantle event
 				// saveAction(creep, 'dismantle', target.pos.x, target.pos.y);
 				context.didUpdate();
 			}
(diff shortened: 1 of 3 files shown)


laverdet:
> The one thing that I wonder about is the use of drop and then checking for a dead container inside drop. We could have a dropOnGround that drop uses to make the code a bit cleaner?

I don't mind the `container.hits` check in drop. I think the bigger issue is that this is the third instance of calling "drop" in a loop, which has to inspect the room each time. It'd be nice if this function accepted a resource store or something. There's just a lot of finicky details like how creeps drop into containers first, and then into tombstones on overflow.

PR #146, opened 2026-04-25
fix(creep): use Fn.take in calculateBoundedEffect to handle sliceCount > deltas.length

misterwise:
`sliceCount = floor(unboosted) = activeParts × power` but `deltas.length = activeParts`. When `power > 1` (repair, build, upgrade), `Fn.slice(deltas, 0, sliceCount)` iterates `deltas[i]` past the array end and yields `undefined`, which poisons `Fn.accumulate` to `NaN` and propagates to `target.hits` in the consuming intents. `Fn.take` short-circuits at iterable end — sibling of `Fn.takeWhile` introduced alongside `iterateActiveParts`.

Verified against ok-screeps.


laverdet:
Good catch. The slice function wasn't working as expected. The intent is to simulate `Array.prototype.slice` without copying the array.

PR #147, opened 2026-04-25
fix: Redirect attack and dismantle damage to ramparts

misterwise:
Restore vanilla rampart-shielding so a same-tile rampart absorbs damage before it reaches the underlying creep or structure.

* Fix `captureDamage` typo (`mods/combat/creep.ts`) — the per-object loop called `target['#captureDamage']` instead of `object['#captureDamage']`, so non-target same-tile objects never had their absorb hook invoked. Forward the caller-provided `type` instead of the hardcoded `EVENT_ATTACK_TYPE_MELEE`. Surface the residual (`iterationPower`) when the target-reach early-return fires; the prior code returned `power`, which still held the previous layer's entry value when the target sat on a lower layer.
* Add `#layer` (= 1) and `#captureDamage` overrides on `StructureRampart`. The override deducts hits in place via `#applyDamage` and returns the residual, matching the contract implied by the base no-op (`return power`).
* Extract `walkLayers` helper so `captureDamage` and `rangedMassAttack` share a single layer-walk + parallel-min implementation. The two loops were duplicates of the same algorithm.
* Update `rangedMassAttack` so `#applyDamage(power)` only fires when nothing was absorbed at this layer; the absorbing object handled its own deduction in `#captureDamage`. Event-log damage reflects the actual hits taken.
* Route dismantle damage through `captureDamage` in the construction processor so a rampart over the target absorbs it; preserve the destroy hook from #145.

Verified against ok-screeps (`RAMPART-PROTECT-001/002`, `COMBAT-MELEE-005`, `COMBAT-RANGED-006`, `COMBAT-RMA-004`, `DISMANTLE-004` all pass).

### DIFF
--- a/packages/xxscreeps/mods/combat/creep.ts
+++ b/packages/xxscreeps/mods/combat/creep.ts
@@ -166,24 +166,24 @@ export function checkDestructible(target: Creep | Structure) {
 }
 
 /**
- * Invokes damage capture callback from top to bottom and returns the remaining power which should
- * be applied to the target.
+ * Walks `objects` descending by `#layer`; `onObject` returns the residual past each one.
+ * `stopAt` ends iteration at that object (throws if absent).
  */
-export function captureDamage(target: RoomObject, initialPower: number, type: number, source: RoomObject | null) {
-	// Sort objects by layer
-	const objects = [ ...Fn.reject(target.room['#lookAt'](target.pos),
-		object => object['#layer'] === undefined || object.hits === undefined) ];
-	objects.sort((left, right) => right['#layer']! - left['#layer']!);
-
-	// Calculate total power, allowing objects on higher layers to deduct damage [ramparts]
+export function walkLayers<T extends RoomObject>(
+	objects: T[],
+	initialPower: number,
+	onObject: (object: T, layerPower: number) => number,
+	stopAt?: T,
+): number {
 	let power = initialPower;
 	let iterationPower = power;
 	let layer: number | undefined;
 	for (const object of objects) {
+		if (object === stopAt) {
+			return iterationPower;
+		}
 		const objectLayer = object['#layer'];
-		if (object === target) {
-			return power;
-		} else if (layer !== objectLayer) {
+		if (layer !== objectLayer) {
 			layer = objectLayer;
 			power = iterationPower;
 			if (power <= 0) {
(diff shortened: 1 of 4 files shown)


packages/xxscreeps/mods/combat/creep.ts
```diff
@@ -167,33 +167,38 @@ export function checkDestructible(target: Creep | Structure) {
...
 			power = iterationPower;
 			if (power <= 0) {
 				return 0;
 			}
 		}
-		// The idea here is that multiple objects on the same layer can capture damage simultaneously,
-		// and whichever one captures more will be used. This doesn't apply to any existing game
-		// objects, but idk maybe it could be interesting.
-		iterationPower = Math.min(iterationPower, target['#captureDamage'](power, C.EVENT_ATTACK_TYPE_MELEE, source));
+		if (object === target) {
```
laverdet: I think this `if` was better where it was?

packages/xxscreeps/mods/defense/rampart.ts
```diff
@@ -25,6 +25,14 @@ export class StructureRampart extends withOverlay(OwnedStructure, shape) {
 
 	override get structureType() { return C.STRUCTURE_RAMPART; }
 
+	// Sort above creeps and other structures so captureDamage sees the rampart first.
+	// eslint-disable-next-line @typescript-eslint/class-literal-property-style
+	override get '#layer'() { return 1; }
+
+	override '#captureDamage'(power: number, _type: number, _source: RoomObject.RoomObject | null) {
+		return Math.max(0, power - this.hits);
```
laverdet: I think the intent of this function was that it would deduct hits in place, returning the remainder of attack power which was not absorbed. That's why the base implementation is a no-op, `return power`. I think the rest of the existing logic makes more sense with that in mind.

laverdet (changes requested):

laverdet:
Note also that ranged mass attack uses similar logic (without the target/object mismatch). The damage calculation there is different because it applies per-position instead of per-target

misterwise:
Ok, reworked it a bit to refactor out the shared logic that rangedMassAttack uses for walking through the layers and preserving the short-circuit for efficiency. #captureDamage was rewritten to applyDamage itself rather than in the helper.

laverdet:
One more thing, and this is inherited from older code, there's a lot of `if (damage > 0) applyDamage(...);` scattered around. '#applyDamage' should just apply that check, otherwise most invocations will need to check one by one. Also the event log calls are duplicated in a lot of cases.

PR #148, opened 2026-04-25
Add hdel to KeyValProvider

misterwise:
## Summary

Adds `hdel(key, fields)` to `KeyValProvider`, with local and redis implementations. Closes an interface gap — local previously had no public `hdel` path, so polymorphic callers had to either reach into the responder's internal `Map` or skip removal.

## Verification

- `pnpm run build` clean.
- `pnpm run test` 177/177 passing.

## Future consumer

This is the first of a series of small PRs splitting up an admin-CLI feature branch I'd previously opened too large (#137). The sibling `users.remove` command in a later PR is the first polymorphic `hdel` caller.

### DIFF
--- a/packages/redis/keyval.ts
+++ b/packages/redis/keyval.ts
@@ -114,6 +114,10 @@ export class RedisProvider implements P.KeyValProvider {
 
 	//
 	// hashes
+	async hdel(key: string, fields: string[]) {
+		return this.redis.invoke<number>(cb => this.redis.batch(key).hdel(key, fields, cb));
+	}
+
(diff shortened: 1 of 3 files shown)


laverdet:
`hgetall` shouldn't be able to return undefined (nil) keys. If you found an instance of the local keyval provider doing this then it is a bug.

misterwise:
> `hgetall` shouldn't be able to return undefined (nil) keys. If you found an instance of the local keyval provider doing this then it is a bug.

Was trying to use hgetall returning undefined to test for key existence which was the wrong approach.

PR #149, opened 2026-04-25
Serialize concurrent save() in local blob and keyval storage

misterwise:
Both `BlobStorage.save()` and `LocalKeyValResponder.save()` write to a fixed `.<basename>.swp` tmp path and then `rename()` it over the live file. Two overlapping `save()` calls race on that single tmp path: the second writeFile clobbers the first's tmp content, the first rename moves the wrong bytes to the live file, and the second rename then fails with `ENOENT` because the tmp is already gone.

The fix is a per-instance `saveChain: Promise<void>` that serializes callers — each `save()` chains its `performSave()` body onto the chain and replaces the chain with an error-swallowed copy so a transient failure doesn't poison subsequent saves. The caller-returned promise still rejects normally; only the chain itself catches.

This race isn't reachable today because only the shutdown path calls `save()`. The first overlapping caller is the upcoming CLI `system.importWorld` command, which calls `save()` after wiping and re-importing the world while shutdown's `backendContext.close()` also triggers a final `save()`. Without this guard, one of the two writes silently disappears.

Includes a regression test at
`engine/db/storage/local/test.ts` covering both providers with two designs:
- a behavior test that wraps `performSave` to record enter/exit timestamps and asserts strict serialization across two concurrent `save()` calls;
- an outcome test that issues two overlapping `set` + `save` pairs and reads the persisted file back from disk to verify last-save-wins. Without `saveChain` the outcome test reproducibly fails with the ENOENT rename error described above.

### DIFF
--- a/.changeset/savechain-serialize-saves.md
+++ b/.changeset/savechain-serialize-saves.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Serialize concurrent `save()` in local blob and keyval storage
--- a/packages/xxscreeps/engine/db/storage/local/blob.ts
+++ b/packages/xxscreeps/engine/db/storage/local/blob.ts
@@ -15,6 +15,7 @@ export class BlobStorage {
 
 	private saveId = 0;
 	private readonly knownPaths = new Set<string>();
+	private saveChain: Promise<void> = Promise.resolve();
 	private readonly path;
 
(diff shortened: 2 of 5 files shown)


laverdet:
I think it's not needed since 6c5d8ca9d36c019d3530f2ed6315d946fe66b59f

PR #150, opened 2026-04-25
Wire up missing Room.getEventLog() events; match vanilla data shape

misterwise:
Closes #107.

Wires up the remaining `Room.getEventLog()` events and reshapes the
player-facing payload to vanilla's `{event, objectId, data?}` shape.

## Events

Newly emitted: `EVENT_BUILD`, `EVENT_REPAIR`, `EVENT_TRANSFER`,
`EVENT_EXIT`, `EVENT_ATTACK_CONTROLLER`, `EVENT_RESERVE_CONTROLLER`,
`EVENT_UPGRADE_CONTROLLER`, `EVENT_OBJECT_DESTROYED` (creep death via
`buryCreep`, structure destruction via `Structure#destroy`, container
decay), and `EVENT_ATTACK` for the previously-unwired `DISMANTLE` and
`HIT_BACK` `attackType` variants. `EVENT_HARVEST`, `EVENT_HEAL`, and
`EVENT_POWER` were already emitted on `main`.

## Single emission at the action site

Each event is emitted exactly once at the call site that owns the
information for it. `Creep.prototype['#applyDamage']` no longer emits
`EVENT_ATTACK` — damage application and event recording are decoupled,
and the previous bug where `RANGED_ATTACK_POWER` (the damage-power
constant, =10) was being passed as the `type` discriminator instead of
`EVENT_ATTACK_TYPE_RANGED` is fixed at the same time. The hit-back
override appends its own entry adjacent to the counter-attack.

## Wire format

Events are stored flat (per-entry density on a per-tick room blob);
`getEventLog` translates at the read boundary via a distributing mapped
type so per-variant `entry.data.foo` narrows on the player side. Entries
with no extras omit `data` rather than emitting `data: {}`, matching
vanilla's JSON exactly. `EVENT_BUILD` carries `structureType`, `x`, `y`,
and `incomplete`.

## Destruction detection

`EVENT_OBJECT_DESTROYED` lives at the lifecycle hook for each death
path:

- **Structures destroyed by damage** (attack, rangedAttack,
  rangedMassAttack, dismantle, rampart absorption death) — emitted in
  `Structure#destroy`. Every damage path that brings hits to zero
  funnels through `super.#applyDamage` → `this.#destroy`, and
  `super['#destroy']()`'s return value is the natural one-shot guard,
  so multi-attacker single-tick destruction emits exactly once.
- **Creep deaths** (combat, suicide, age-out) — emitted in `buryCreep`
  in `mods/creep/tombstone.ts`.
- **Container decay** — emitted in
  `mods/resource/processor/container.ts` because decay deliberately
  bypasses `#destroy` (no Ruin from natural decay; contents spill to
  the ground per vanilla `_create-energy.js`).

Per vanilla `_damage.js`, the order on the wire for a kill-shot is
`[DESTROYED, HIT_BACK, ATTACK]` — destroyed event first, then any
counter-attack, then the originating attack. Intent processors emit
`EVENT_ATTACK` after `target['#applyDamage']` so the destroyed event
(emitted from inside the damage chain) lands first.

Verified against screeps-ok.

### DIFF
--- a/.changeset/wire-event-log-events.md
+++ b/.changeset/wire-event-log-events.md
@@ -1,0 +1,11 @@
+---
+"xxscreeps": patch
+---
+
+Wire up missing `Room.getEventLog()` events and reshape the player-facing payload to vanilla's `{event, objectId, data?}` shape.
+
+Newly emitted: `EVENT_BUILD`, `EVENT_REPAIR`, `EVENT_TRANSFER`, `EVENT_EXIT`, `EVENT_ATTACK_CONTROLLER`, `EVENT_RESERVE_CONTROLLER`, `EVENT_UPGRADE_CONTROLLER`, `EVENT_OBJECT_DESTROYED` (on creep death, structure destruction by attack/dismantle, and container decay), and `EVENT_ATTACK` for the previously-unwired `DISMANTLE` and `HIT_BACK` `attackType` variants. Each event is emitted exactly once at the intent processor that performs the action; `Creep.prototype['#applyDamage']` no longer emits, keeping damage application and event recording decoupled. `EVENT_OBJECT_DESTROYED` is gated on the alive→dead transition so multi-attacker ticks don't duplicate it.
+
+`EVENT_BUILD` carries `structureType`, `x`, `y`, and `incomplete` to match vanilla's payload.
+
+Closes #107. Verified against screeps-ok.
--- a/.changeset/wire-event-log-events.md
+++ b/.changeset/wire-event-log-events.md
@@ -2,10 +2,4 @@
 "xxscreeps": patch
 ---
 
-Wire up missing `Room.getEventLog()` events and reshape the player-facing payload to vanilla's `{event, objectId, data?}` shape.
-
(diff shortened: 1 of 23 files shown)


misterwise:
This one is not ready yet and needs some rework and further testing.

misterwise:
Ok, this is ready for review now.

packages/xxscreeps/game/room/event-log.ts
```diff
@@ -6,6 +6,24 @@ import { Room } from './room.js';
 type RemoveVariant<T> = T extends any ? Omit<T, typeof Variant> : never;
 export type AnyEventLog = RemoveVariant<Room['#eventLog'][number]>;
 
+// Vanilla shape: `{event, objectId, data?}`, with `data` omitted when there are no
+// extra fields. Distributes over the union so per-variant payloads stay narrowable.
+type EventToPlayer<E> = E extends { event: infer Ev; objectId: infer Id }
```
laverdet: Just make it something like this:
```ts
interface GameEvent {
	event: string;
	objectId: string;
	targetId?: string;
	[key: string]: unknown;
}
```

.changeset/wire-event-log-events.md
```diff
@@ -0,0 +1,11 @@
+---
+"xxscreeps": patch
+---
+
+Wire up missing `Room.getEventLog()` events and reshape the player-facing payload to vanilla's `{event, objectId, data?}` shape.
```
laverdet: These changesets should be like 15 words at most.

PR #151, opened 2026-04-25
Plumb active-rooms set into GameMap.getRoomStatus

misterwise:
## Summary

`GameMap.getRoomStatus()` returns `'normal'` for any terrain-backed room,
ignoring whether the room is in the active set (`shard.data.smembers('rooms')`).
Closed/removed rooms appear `'normal'` to every consumer of `Shard.loadWorld()`
— `/api/game/room-status`, `/map-stats`, the map socket filter, and any
downstream caller.

## Change

- `Shard.loadWorld()` reads `'rooms'` in parallel with the terrain blob
  and passes the set to `new World(...)`.
- `World` and `GameMap` accept an optional `accessibleRooms: ReadonlySet<string>`.
  When supplied, `GameMap.getRoomStatus()` returns `'out of borders'` for
  terrain-backed rooms outside the set; otherwise `'normal'` (unchanged).
- The set is optional because the user-code sandbox (`driver/runtime/index.ts`)
  and the room-tick worker (`engine/processor/worker.ts`) build `World` from
  out-of-process terrain blobs with no `shard.data` handle. Plumbing the
  set into those payloads — and invalidating sandboxes when it changes —
  is intentionally out of scope.

## Standalone bug fix

There's no way to *produce* a closed room from xxscreeps itself today, so
the bug isn't observable in normal operation, but the `'rooms'` set is
already written by `scripts/scrape-world.ts` (world import) and the test
fixture (`packages/xxscreeps/test/import.ts`). Both of those writes are
silently ignored by `getRoomStatus` until this fix lands.

## Bonus

Two-character drive-by in `Shard.connect()`: the unknown-shard error
interpolated `${shard}` (the `undefined` result of the just-failed
`find()`) instead of `${name}`. Happy to split into a separate PR if
you'd rather keep this one strictly to the active-rooms work.

Verified against ok-screeps.

### DIFF
--- a/packages/xxscreeps/game/map.ts
+++ b/packages/xxscreeps/game/map.ts
@@ -41,7 +41,7 @@ type FindRoute = {
  */
 export class GameMap {
 	readonly #terrain: TerrainByRoom;
-	readonly #accessibleRooms: ReadonlySet<string> | undefined;
+	readonly #accessibleRooms;
 	readonly #left;
 	readonly #top;
 	readonly #height;
@@ -267,13 +267,11 @@ export class World {
 	name;
 	terrain: TerrainByRoom;
 	terrainBlob;
-	accessibleRooms: ReadonlySet<string> | undefined;
 
 	constructor(name: string, terrainBlob: Readonly<Uint8Array>, accessibleRooms?: ReadonlySet<string>) {
 		this.name = name;
 		this.terrainBlob = terrainBlob;
 		this.terrain = reader(terrainBlob);
-		this.accessibleRooms = accessibleRooms;
 		this.map = new GameMap(this.terrain, accessibleRooms);
 	}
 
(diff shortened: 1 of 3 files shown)


packages/xxscreeps/game/map.ts
```diff
@@ -41,13 +41,15 @@ type FindRoute = {
  */
 export class GameMap {
 	readonly #terrain: TerrainByRoom;
+	readonly #accessibleRooms: ReadonlySet<string> | undefined;
```
laverdet: You don't have to specify the type here.

packages/xxscreeps/game/map.ts
```diff
@@ -260,12 +267,14 @@ export class World {
 	name;
 	terrain: TerrainByRoom;
 	terrainBlob;
+	accessibleRooms: ReadonlySet<string> | undefined;
```
laverdet: I don't see a reason to define it on World if GameMap already gives you access to it. Especially since it's maybe undefined.

misterwise: Yes, you're right, Game.map.getRoomStatus() is what should be used by the CLI. This was the wrong approach.

PR #152, opened 2026-04-26
Fix Creep.transfer redirect to upgradeController; reject pull on spawning

misterwise:
Two small parity fixes against canonical Creep API behavior, exposed by ok-screeps tests TRANSFER-011 and MOVE-PULL-007:spawning.

`transfer(controller, RESOURCE_ENERGY)` previously fell through to the generic Store path and rejected with `ERR_INVALID_TARGET`, since `StructureController` has no Store. Vanilla short-circuits at the top of `Creep.prototype.transfer` and reroutes to `upgradeController` (`@screeps/engine/src/game/creeps.js`). Mirroring that here adds a one-branch `instanceof` check before the existing Store logic.

I considered placing the wrapper in `mods/controller/creep.ts` to avoid the new import edge, but the controller mod already depends on `mods/creep/creep.js` for `Creep`, and there's no transitive cycle through `mods/controller/controller.ts` (which only pulls from `mods/structure` and `mods/schema`). Importing `StructureController` directly into `mods/creep/creep.ts` is the cleanest fit.

`pull()` previously accepted spawning-creep targets. Vanilla rejects them with `ERR_INVALID_TARGET` before any range check fires. Adding the guard at the end of `checkPull` keeps existing target/range error precedence and only rejects when everything else is otherwise OK, matching vanilla's check ordering.

Verified against ok-screeps.

### DIFF
--- a/.changeset/creep-controller-transfer-and-pull-spawning.md
+++ b/.changeset/creep-controller-transfer-and-pull-spawning.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Redirect `Creep.transfer` to `upgradeController` for energy targeting a controller, and reject `Creep.pull` against a spawning creep.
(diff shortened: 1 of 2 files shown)


laverdet:
This is a module concern leak but it's such a dumb case from Screeps that it's probably the best option.

PR #153, opened 2026-04-26
Add Creep.withdraw enemy-rampart guard; fix moveTo noPathFinding return code

misterwise:
Two unrelated `Creep` API fixes batched into one PR (both small), exposed by ok-screeps tests WITHDRAW-005 and MOVE-BASIC-019.

`checkWithdraw` had no rampart-ownership test on the target tile, so a creep could withdraw from a hostile structure sitting under a non-public enemy rampart. Vanilla rejects with `ERR_NOT_OWNER` at `@screeps/engine/src/game/creeps.js:525-527`. Added a `checkEnemyRampart` step in the chain after `checkTarget` and before `checkRange`, gated on `target instanceof OwnedStructure && !target.my` so neutral structures, tombstones, and ruins are unaffected. Public ramparts (`rampart.isPublic === true`) skip the guard, matching vanilla's `!i.isPublic` predicate. Used `lookForStructureAt(..., STRUCTURE_RAMPART)` rather than scanning `lookForAt(LOOK_STRUCTURES, ...)` since the rampart construction check already enforces at most one rampart per tile (`mods/defense/rampart.ts:71-76`).

`Creep.moveTo({ noPathFinding: true })` returned `ERR_NO_PATH` when no cached path existed; vanilla returns `ERR_NOT_FOUND` for that specific case (`@screeps/engine/src/game/creeps.js:279-281`), distinct from "no path exists between these points". `searchOrFetchPath` only returns `null` from the `noPathFinding && !cachedPath` branch (cache hits return arrays, `findPathTo` returns arrays), so the `null` sentinel is uniquely the noPathFinding-no-cache case and updating the caller's error code is sufficient. The empty-path fallback below still returns `ERR_NO_PATH` for genuinely unreachable targets.

Verified against ok-screeps.

### DIFF
--- a/.changeset/brown-buses-argue.md
+++ b/.changeset/brown-buses-argue.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Add Creep.withdraw enemy-rampart guard; fix moveTo noPathFinding return code
--- a/packages/xxscreeps/mods/creep/creep.ts
+++ b/packages/xxscreeps/mods/creep/creep.ts
@@ -22,7 +22,7 @@ import * as Memory from 'xxscreeps/mods/memory/memory.js';
 import { Resource, optionalResourceEnumFormat } from 'xxscreeps/mods/resource/resource.js';
 import { OpenStore, calculateChecked, checkHasCapacity, checkHasResource, openStoreFormat } from 'xxscreeps/mods/resource/store.js';
 import { Ruin } from 'xxscreeps/mods/structure/ruin.js';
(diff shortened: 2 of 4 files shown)


laverdet:
Could you do something like '#doesPreventInteraction'()` on Structure, then override on rampart? Creep could then check all structures on the tile. Theoretically Creep doesn't know anything about Rampart.

misterwise:
> Could you do something like '#doesPreventInteraction'()` on Structure, then override on rampart? Creep could then check all structures on the tile. Theoretically Creep doesn't know anything about Rampart.

Yes, I like the shape of that, have a look now.

PR #155, opened 2026-04-26
Reset module-level Memory state per simulation in test harness

misterwise:
## Summary

`mods/memory/memory.ts` keeps its parse cache (`json`, `string`) and `RawMemory._parsed` in module-level state. In production every sandbox runs `runtimeConnector.initialize(memoryBlob)` once at startup, so that state is per-process-per-user and isolation is implicit. Tests don't go through that path — every `simulate(...)` body shares one Node process, so cached Memory contents from a prior test's `runForUser` calls survive into the next test's first read.

Reset at the `simulate(...)` body boundary using already-exported pieces (`initialize(null)` and `RawMemory._parsed = undefined`). No new exports from `mods/memory/memory.ts`; the test-isolation logic stays in test infrastructure.

## Why now / what risk does this resolve

Today this is **latent, not observable**. Every consumer that reads Memory keys validates the value (e.g. the `_move` path cache rejects mismatched origin/dest), so no existing test observes the leak.

The risk is forward-looking: a future test that asserts on initial Memory shape, or that reuses creep / room / flag / spawn names across tests in the same file, could quietly read the prior test's writes. #140 doesn't widen the leak (per-object `.memory` accessors still go through the module-level `get()` directly, not `globalThis.Memory`), but it does make the leak more deterministic — after #140, `flush()` always drops the `json` cache, so the next tick re-parses from `string`. A sibling test that reads Memory before writing will reliably see the prior test's parsed blob instead of occasionally hitting a stale-but-matching `_parsed`.

Cheap insurance, contained to test infra.

## Validation

- `pnpm run test` — 177/177 passing on this branch.
- Verified against screeps-ok.

### DIFF
--- a/.changeset/test-simulate-memory-isolation.md
+++ b/.changeset/test-simulate-memory-isolation.md
@@ -1,0 +1,9 @@
+---
+"xxscreeps": patch
+---
+
+Reset module-level Memory state per simulation in test harness
+
+`simulate.ts` now clears `mods/memory/memory.ts` module state (`json` /
+`string` caches and `RawMemory._parsed`) at the start of each test body,
+so cached Memory contents from one test don't leak into the next.
--- a/.changeset/test-simulate-memory-isolation.md
+++ b/.changeset/test-simulate-memory-isolation.md
@@ -2,8 +2,4 @@
 "xxscreeps": patch
 ---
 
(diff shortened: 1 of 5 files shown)


laverdet:
Closing in favor of 32c9fdb39f31a2cf9b65f31b029110c805aa7f53

PR #156, opened 2026-04-26
Fix iterateActiveParts; add Fn.dropWhile

misterwise:
`recalculateBody` distributes damage front-to-back, so `body[0]` is the first to drop to zero. `iterateActiveParts` used `Fn.takeWhile`, which halted at the first dead part and yielded nothing whenever index 0 was dead — breaking `getActiveBodyparts`, `calculateCarry`, `calculatePower`, and `calculateBoundedEffect` on damaged creeps.

Replace with `Fn.dropWhile` (new sibling of `takeWhile`): dead parts cluster at the front, so dropping them yields the active suffix.

Verified against screeps-ok.

### DIFF
--- a/.changeset/active-parts-filter.md
+++ b/.changeset/active-parts-filter.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Fix `iterateActiveParts` halting on front-damaged creeps; add `Fn.dropWhile`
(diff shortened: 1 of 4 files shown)


misterwise:
Also looked at reversing it first to preserve the intention, but it doesn't really seem it would be more efficient as it would require Fn.reverse. Actually, I think I see a way to short-circuit for healthy creeps which is a good optimization we can do.

PR #157, opened 2026-04-26
fix: Bootstrap generated mod imports before command modules

misterwise:
`Simplify entry script` removed the eager mod initialization from `config/entry.ts`, but built-in commands still link modules that resolve through `xxscreeps/config/mods.static/*`. On a cold CI build those generated files do not exist yet, so `npx xxscreeps test` fails during ESM linking before `test/run.ts` can call `importMods()`.

This keeps storage provider registration lazy, but restores the minimal runtime bootstrap needed before importing a built-in command module:

- import `config/mods/index.js` to generate `dist/config/mods.static/*`
- import `config/global.js` before runtime decorators are needed
- then dynamically import the resolved command module

Verified under Node 24:

```bash
pnpm run build
pnpm run test
# 177 tests: 177 passed, 0 failed
```


misterwise:
A proposed fix for the CI failures that I think aligns with your intention, you may well have a better approach!

laverdet:
fed673c675139a7918402c0789c5c2384bfc3e67 🗿

PR #158, opened 2026-04-26
Graceful shutdown contract for sibling-process publishers

misterwise:
## Summary

Move SIGINT ownership to the launcher; make `main` a service-channel listener with bailouts at the three windows where a channel-borne `shutdown` could find it blocked. Shutdown always lands on a coherent tick boundary now — never mid-tick.

**Why for the upcoming admin CLI:** commands like `map.removeRoom` and `system.importWorld` mutate state the engine has cached (worker `roomCache`, `World.accessibleRooms`, `PlayerInstance` sandbox bakes). In-place refresh needs sandbox recreation; the alternative is to restart after the mutation — signaled from the launcher by publishing `shutdown` on the service channel. Without these bailouts that publish either does nothing (main only listens for SIGINT today) or lets main run a tick against post-mutation state.

## Bailouts

- **Init loop** — explicit `'shutdown'` check + skip `begetRoomProcessQueue` if halted. The launcher's SIGINT handler now publishes on the channel rather than halting main directly, so an operator hitting Ctrl-C during boot (wrong config, wrong shard) needs the init iterator to react to the channel message.
- **Mutex acquire** — `if (halted) return;` re-check inside `gameMutex.scope`. Covers `system.importWorld` holding the mutex, wiping scratch + reimporting, then publishing `shutdown` before `main` can run a tick body against post-import state.
- **Tick wait** — wrap inner `for await` in `Async.breakable`, skip state update if shutdown (not `tickFinished`) broke the loop. Otherwise main waits the full `intentAbandonTimeout` because workers received `process` for a tick they cannot complete.

Launcher also re-publishes `shutdown` after `mainConnected` (covers SIGINT-before-subscribe race) and skips worker startup if shutdown was already requested.

## Validation

`pnpm run build && pnpm run test` clean (177/177). Verified against screeps-ok.

### DIFF
--- a/.changeset/feat-shutdown-contract.md
+++ b/.changeset/feat-shutdown-contract.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Add a graceful shutdown contract that lets a sibling process publish `shutdown` on the service channel and have main exit cleanly without corrupting state
--- a/packages/xxscreeps/engine/service/launcher.ts
+++ b/packages/xxscreeps/engine/service/launcher.ts
@@ -4,7 +4,7 @@ import { Database, Shard } from 'xxscreeps/engine/db/index.js';
 import * as User from 'xxscreeps/engine/db/user/index.js';
 import { getConsoleChannel } from 'xxscreeps/engine/runner/model.js';
 import { Worker, waitForWorker } from 'xxscreeps/utility/worker.js';
-import { getServiceChannel } from './index.js';
+import { getServiceChannel, handleInterrupt } from './index.js';
 
 const argv = checkArguments({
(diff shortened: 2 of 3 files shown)


laverdet:
I'll take a look at this one

laverdet:
I think 8e74ef128a1c0b82992634c6db14f3bbcf5aa96a and 579213ec7f88a18cdc4aad1a7012ab88fdc05d61 take care of most of what you wanted. The main loop will terminate on a service channel 'shutdown' event without aborting a tick in progress. After the main loop terminates it requests that processor and runner terminate as well. The backend will continue to run but I think that's correct because it doesn't mutate state and is resilient to loop restarts. In the case of terrain modification the administrator would have to manually bounce that service.

PR #159, opened 2026-04-27
Add portal mod with same-shard and cross-shard (stub for now) destinations

misterwise:
Implements `StructurePortal` as a real game object, replacing the empty `registerGlobal` stub in `game/runtime.ts`. Mirrors vanilla `@screeps/engine` semantics: same-shard `destination` returns a `RoomPosition`, cross-shard returns `{shard, room}`. `ticksToDecay` is `undefined` for permanent portals.

## Changes

- New `mods/portal/` module — schema, processor (decay + creep teleport), backend renderers, manifest. Wired into `mods/classic`.
- Extract `teleportCreep` helper from `mods/creep/processor.ts` so the portal processor reuses the existing border-cross `writeRoomObject` / `sendRoomIntent('import')` path.
- `scrape-world` reads portal destination + decay time from imported world data so portals survive a world import.
- Cross-shard teleport short-circuits (single-shard server); destination shape is still preserved on read.
- Same-shard teleport gates on `#user.length > 2` to match the existing border-cross NPC exclusion.

## Validation

- New `mods/portal/test.ts` covers destination shape (same-shard / cross-shard) and `ticksToDecay` (decaying / permanent).
- Verified against screeps-ok.

### DIFF
--- a/packages/xxscreeps/mods/portal/backend.ts
+++ b/packages/xxscreeps/mods/portal/backend.ts
@@ -3,16 +3,8 @@ import { StructurePortal } from './portal.js';
 
 bindMapRenderer(StructurePortal, () => 'p');
 
-bindRenderer(StructurePortal, (portal, next) => {
-	const decayTime = portal['#decayTime'];
-	const out = {
-		...next(),
-		destination: portal['#destShard'] === ''
-			? { x: portal['#destX'], y: portal['#destY'], room: portal['#destRoom'] }
-			: { shard: portal['#destShard'], room: portal['#destRoom'] },
-	};
-	if (decayTime === 0) {
-		return out;
-	}
-	return { ...out, decayTime };
-});
+bindRenderer(StructurePortal, (portal, next) => ({
+	...next(),
+	destination: portal.destination,
+	decayTime: portal['#decayTime'] || undefined,
+}));
--- a/packages/xxscreeps/mods/portal/backend.ts
+++ b/packages/xxscreeps/mods/portal/backend.ts
@@ -1,0 +1,17 @@
+import { bindMapRenderer, bindRenderer } from 'xxscreeps/backend/index.js';
+import { StructurePortal } from './portal.js';
+
+bindMapRenderer(StructurePortal, () => 'p');
+
+bindRenderer(StructurePortal, (portal, next) => {
+	const out = {
+		...next(),
+		destination: portal['#destShard'] === ''
+			? { x: portal['#destX'], y: portal['#destY'], room: portal['#destRoom'] }
+			: { shard: portal['#destShard'], room: portal['#destRoom'] },
+	};
+	if (portal['#decayTime'] !== 0) {
+		(out as any).decayTime = portal['#decayTime'];
+	}
+	return out;
+});
--- a/packages/xxscreeps/mods/portal/backend.ts
+++ b/packages/xxscreeps/mods/portal/backend.ts
@@ -4,14 +4,15 @@ import { StructurePortal } from './portal.js';
 bindMapRenderer(StructurePortal, () => 'p');
 
 bindRenderer(StructurePortal, (portal, next) => {
+	const decayTime = portal['#decayTime'];
 	const out = {
 		...next(),
 		destination: portal['#destShard'] === ''
 			? { x: portal['#destX'], y: portal['#destY'], room: portal['#destRoom'] }
 			: { shard: portal['#destShard'], room: portal['#destRoom'] },
 	};
-	if (portal['#decayTime'] !== 0) {
-		(out as any).decayTime = portal['#decayTime'];
+	if (decayTime === 0) {
+		return out;
 	}
-	return out;
+	return { ...out, decayTime };
 });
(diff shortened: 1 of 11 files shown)


packages/xxscreeps/mods/portal/backend.ts
```diff
@@ -0,0 +1,18 @@
+import { bindMapRenderer, bindRenderer } from 'xxscreeps/backend/index.js';
+import { StructurePortal } from './portal.js';
+
+bindMapRenderer(StructurePortal, () => 'p');
+
+bindRenderer(StructurePortal, (portal, next) => {
+	const decayTime = portal['#decayTime'];
```
laverdet: `portal['#decayTime'] || undefined`

packages/xxscreeps/mods/portal/portal.ts
```diff
@@ -0,0 +1,59 @@
+import * as C from 'xxscreeps/game/constants/index.js';
+import { Game } from 'xxscreeps/game/index.js';
+import * as RoomObject from 'xxscreeps/game/object.js';
+import { RoomPosition } from 'xxscreeps/game/position.js';
+import { Structure, structureFormat } from 'xxscreeps/mods/structure/structure.js';
+import { compose, declare, struct, variant, withOverlay } from 'xxscreeps/schema/index.js';
+
+export type SameShardDestination = { shard?: undefined; room: string; x: number; y: number };
```
laverdet:
```
interface PortalDestination extends RoomPosition { shard?: undefined; }
interface RemotePortalDestination { shard: string; room: string };
```

Otherwise you lose the RoomPosition properties. Also, just as a general rule, please use interfaces where possible.

packages/xxscreeps/mods/portal/backend.ts
```diff
@@ -0,0 +1,18 @@
...
+	const out = {
+		...next(),
+		destination: portal['#destShard'] === ''
```
laverdet: Prefer accessing the "public" non-'#' properties, even for internal code when possible. For decayTime that makes sense but here you could just use portal.destination.

packages/xxscreeps/mods/portal/backend.ts
```diff
@@ -0,0 +1,18 @@
...
+			? { x: portal['#destX'], y: portal['#destY'], room: portal['#destRoom'] }
+			: { shard: portal['#destShard'], room: portal['#destRoom'] },
+	};
+	if (decayTime === 0) {
```
laverdet: You also wouldn't need this `=== 0` with the undefined trick above.

packages/xxscreeps/mods/portal/processor.ts
```diff
@@ -0,0 +1,30 @@
...
+	// Decay
+	const decayTime = portal['#decayTime'];
+	if (decayTime !== 0 && Game.time >= decayTime) {
+		portal.room['#removeObject'](portal);
+		context.didUpdate();
+		return;
+	}
+
+	// Cross-shard portals are not yet supported (single-shard server)
+	if (portal['#destShard'] === '') {
```
laverdet: Like this code would be a lot better if you just used the public property since it already knows how to make a RoomPosition

packages/xxscreeps/scripts/scrape-world.ts
```diff
@@ -77,6 +79,35 @@ function withStore(from: any, into: { store: Store }) {
...
+		return { room, x: xx, y: yy };
+	}
+}
+
+function getPortalDecayTime(object: any): number {
+	const decayTime = object.decayTime as unknown;
+	if (typeof decayTime === 'number') {
+		return decayTime;
+	}
+	const ticksToDecay = object.ticksToDecay as unknown;
```
laverdet: This isn't correct. Screeps always uses the same `decayTime` trick that we use, to avoid updating the objects every single tick.

https://github.com/screeps/engine/blob/1b9b1541923f061311474a2f1bac0fea37911f70/src/game/structures.js#L1416

packages/xxscreeps/scripts/scrape-world.ts
```diff
@@ -77,6 +79,35 @@ function withStore(from: any, into: { store: Store }) {
 	}
 }
 
+function getPortalDestination(object: any): Destination | undefined {
+	const rawDestination = object.destination as unknown;
+	const destination = rawDestination !== null && typeof rawDestination === 'object'
+		? rawDestination as Record<string, unknown>
+		: {};
+	const shard = destination.shard ?? object.destShard as unknown;
```
laverdet: destShard & destRoom are xxscreeps properties that you made. They don't exist in the screeps codebase.

packages/xxscreeps/scripts/scrape-world.ts
```diff
@@ -77,6 +79,35 @@ function withStore(from: any, into: { store: Store }) {
 	}
 }
 
+function getPortalDestination(object: any): Destination | undefined {
+	const rawDestination = object.destination as unknown;
+	const destination = rawDestination !== null && typeof rawDestination === 'object'
```
laverdet: destination is always defined, and always an object on these things. I don't think you need to be *this* defensive here, throwing is probably better than returning nothing in the case of an invariant violation.

laverdet (changes requested):

misterwise:
Addressed feedback, and used it to strengthen behavioural coverage in screeps-ok as well.

PR #161, opened 2026-04-28
Implement Game.notify queueing layer

misterwise:
## Summary

Implements `Game.notify(message, groupInterval?)` end-to-end. Replaces the
`console.error('TODO: notify')` stub.

## Approach

- New `IntentManager.pushUser` channel for user-scope intents (no room or object
  receiver). Vanilla-style 20/tick cap returning `OK` / `ERR_FULL`.
- Module-augmentable `UserIntent` interface in `game/intents.ts` keeps the
  surface type-safe; mods register typed handlers via
  `registerUserIntentHandler` in `engine/runner/intents.ts`.
- Dispatch runs in the runner connector save hook (same shape as
  flag/memory/visual driver hooks). Per-user save() is single-flighted, which
  is what lets the coalescing upsert read-then-write safely.
- Storage: `user/${id}/notifications` index set + per-row hash at
  `user/${id}/notifications/${rowId}`, where `rowId` is a content hash of
  `(type, date, message)` so coalescing is a single read.
- Vanilla parity: `'' + i.message` coercion, 500-char truncation, groupInterval
  clamped to `[0, 1440]` minutes, bucket date is `ceil(now / intervalMs) * intervalMs`.

## Validation

10 cases in `mods/notifications/test.ts` cover the cap (returns `OK`, 21st call
returns `ERR_FULL`, cap resets next tick), row persistence, coalescing within a
`groupInterval` bucket, clamp bounds, message coercion (`null` → `"null"`,
`{a:1}` → `"[object Object]"`), no-args, and non-numeric `groupInterval`
fall-through.

## Follow-up

Persistence only — no reader yet. A backend endpoint (matching vanilla's
notification API) and a pruning policy are intentionally out of scope here.


packages/xxscreeps/mods/notifications/processor.ts
```diff
@@ -0,0 +1,19 @@
...
+import { upsertNotification } from './model.js';
+
+declare module 'xxscreeps/game/intents.js' {
+	interface UserIntent {
+		notify: [ message: string, groupInterval: number | undefined ];
+	}
+}
+
+registerUserIntentHandler('notify', async (shard, userId, message, groupInterval) => {
+	const interval = typeof groupInterval === 'number' && Number.isFinite(groupInterval)
```
laverdet: I find this kind of cumbersome. Maybe instead:
```ts
Fn.pipe(
  Number(groupInterval),
  $$ => $$ > 0 ? $$ : 0,
  $$ => Math.min($$, 1440) * 60_000);
```

packages/xxscreeps/mods/notifications/processor.ts
```diff
@@ -0,0 +1,19 @@
...
+		? Math.min(1440, Math.max(0, groupInterval)) : 0;
+	const intervalMs = interval * 60_000;
+	const now = Date.now();
+	const date = intervalMs > 0 ? Math.ceil(now / intervalMs) * intervalMs : now;
+	// Vanilla parity: engine uses `'' + i.message`.
```
laverdet: Vanilla uses `'' + i.message` because it was written with Crockford-era standards. Prefer: 
```ts
`${message}`.slice(0, 500)
```

packages/xxscreeps/mods/notifications/model.ts
```diff
@@ -0,0 +1,57 @@
...
+	message: string;
+	date: number;
+	count: number;
+	type: NotificationType;
+};
+
+const userIndexKey = (userId: string) => `user/${userId}/notifications`;
+const rowKey = (userId: string, rowId: string) => `user/${userId}/notifications/${rowId}`;
+
+function rowIdFor(type: NotificationType, date: number, message: string) {
```
laverdet: `date` should be renamed throughout to `timeGroup` or something because it's not actually a date.

packages/xxscreeps/mods/notifications/model.ts
```diff
@@ -0,0 +1,57 @@
...
+	return createHash('sha1').update(`${type}${date}${message}`).digest('hex');
+}
+
+export async function getNotifications(shard: Shard, userId: string): Promise<NotificationRow[]> {
+	const ids = await shard.data.smembers(userIndexKey(userId));
+	return Promise.all(Fn.map(ids, async id => {
```
laverdet: `Fn.mapAwait` does Fn.map + Promise.all

packages/xxscreeps/mods/notifications/model.ts
```diff
@@ -0,0 +1,57 @@
...
+export async function upsertNotification(
+	shard: Shard, userId: string, message: string, date: number, type: NotificationType,
+) {
+	if (typeof message !== 'string') {
+		throw new TypeError('notification message must be a string');
+	}
+	if (!Number.isFinite(date)) {
+		throw new TypeError('notification date must be a finite number');
+	}
+	const id = rowIdFor(type, date, message);
```
laverdet: I think if you moved the date grouping logic from the intent handler into this function it would be more clear. As it is with the way you have it written you are storing the truncated (grouped) date in the record which probably isn't what the user wants.

packages/xxscreeps/mods/notifications/model.ts
```diff
@@ -0,0 +1,57 @@
...
+	const key = rowKey(userId, id);
+	// Read-then-write is safe because the runner serializes save() per user.
+	const existing = await shard.data.hget(key, 'count');
+	if (existing === null) {
+		await Promise.all([
+			shard.data.hmset(key, { message, date, count: 1, type }),
```
laverdet: These should probably expire otherwise they will build up forever without a consumer. I would recommending adding a TODO since the keyval implementation is incomplete.

packages/xxscreeps/mods/notifications/test.ts
```diff
@@ -0,0 +1,167 @@
...
+import { getNotifications } from './model.js';
+
+const user = '100';
+
+const empty = simulate({
+	W0N0: () => {},
+});
+
+// Manual clock control. xxscreeps' test runner has no fake-timer integration, so we patch
+// `Date.now` globally for the duration of the closure. Safe only because vitest runs tests within
```
laverdet: This is a hallucination. We don't use vitest.

packages/xxscreeps/mods/notifications/driver.ts
```diff
@@ -0,0 +1,9 @@
+import { hooks } from 'xxscreeps/engine/runner/index.js';
+import { dispatchUserIntents } from 'xxscreeps/engine/runner/intents.js';
+import './processor.js';
+
+hooks.register('runnerConnector', player => [ undefined, {
+	async save(result) {
+		await dispatchUserIntents(player.shard, player.userId, result.userIntents);
```
laverdet: This is a fake abstraction. 'runner' owns `dispatchUserIntents` but 'notification' owns the connector? It doesn't make any sense.

I think it'd be better to structure it more similar to flags or memory, both of which operate in a similar manner.

packages/xxscreeps/game/game.ts
```diff
@@ -119,8 +120,8 @@ export class Game extends GameBase {
 	 * on any occasion within the game. You can schedule up to 20 notifications during one game tick.
 	 * Not available in the Simulation Room.
 	 */
-	notify(_message: string, _groupInterval?: number) {
-		console.error('TODO: notify');
+	notify(message: string, groupInterval?: number) {
+		return intents.pushUser('notify', [ message, groupInterval ], 20);
```
laverdet: In regards to the previous comment about structuring it more like 'flag' or 'memory' this function would remain a stub, maybe console.warn with a message about not having a notification mod installed. Then the notification mod would replace the shim with the implementation in 'gameInitializer'

packages/xxscreeps/mods/notifications/model.ts
```diff
@@ -0,0 +1,57 @@
...
+			type: fields.type as NotificationType,
+		};
+	}));
+}
+
+export async function upsertNotification(
+	shard: Shard, userId: string, message: string, date: number, type: NotificationType,
+) {
+	if (typeof message !== 'string') {
+		throw new TypeError('notification message must be a string');
```
laverdet: I don't think we want to throw in cases like this. The concern of coercing user input is not this function, it would be in the 'runnerConnector'.

packages/xxscreeps/mods/notifications/test.ts
```diff
@@ -0,0 +1,167 @@
...
+// a describe serially; two parallel callers would clobber each other's restore.
+async function withFrozenTime<Type>(now: number, fn: () => Promise<Type>) {
```
laverdet: Not sure if you saw my recent commits with `using xx = ...` but it would make this read a lot better. You wouldn't need a nested function anymore.

laverdet (changes requested):

packages/xxscreeps/mods/notifications/notifications.ts
```diff
@@ -0,0 +1,30 @@
...
+	groupInterval: unknown;
+};
+
+let queue: QueuedNotification[] = [];
+
+export function notify(message: unknown, groupInterval?: unknown) {
+	if (queue.length >= kPerTickCap) {
+		return C.ERR_FULL;
+	}
+	queue.push({ type: 'msg', message, groupInterval });
```
laverdet, line 20: Don't forget `intents.cpu += kCpuCost`

packages/xxscreeps/mods/notifications/notifications.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export function flush() {
+	const tmp = queue;
+	queue = [];
+	return tmp;
+}
+
+export function reset() {
+	queue = [];
```
laverdet: I'm a bit confused, is this actually needed?

packages/xxscreeps/mods/notifications/notifications.ts
misterwise: Thanks, good catch.

packages/xxscreeps/mods/notifications/notifications.ts
misterwise: It was not, this was a fix to solve a test runner issue, the fix was moved to the test runner instead.

PR #163, opened 2026-04-28
Strip boost from unboosted body parts to match vanilla shape

misterwise:
## Issue

Body parts deserialized via the schema struct always have `boost` as an own property — `Object.getOwnPropertyNames(part)` includes `boost: undefined` for unboosted parts. Vanilla only sets `boost` after `applyBoost`, so own-property introspection (`'boost' in part`, `Object.keys`) returns true only for boosted parts. Player code that walks body shape directly sees a different surface on xxscreeps than on the official engine.

## Fix

Wrap the body-part struct in a `compose()` interceptor that strips `boost` on read when undefined and supplies it on write. Switch `unboostCreep` (chemistry) and `renewCreep` (spawn) from `part.boost = undefined` to `delete part.boost` so the in-tick processor view also matches vanilla.

After this change, `creep.body[i]` is `{ hits, type }` for unboosted parts and `{ boost, hits, type }` for boosted parts. Schema binary layout and existing archive entries are unchanged.

## Validation

New `body parts match vanilla own-property shape` test in `mods/chemistry/test.ts` asserts `'boost' in part` is absent pre-boost, present on TOUGH parts post-boost, and absent on the others. `unboostCreep removes all boosts` extended to assert the property is deleted on the strip path. Verified against screeps-ok.

### DIFF
--- a/.changeset/body-part-boost-shape.md
+++ b/.changeset/body-part-boost-shape.md
@@ -2,4 +2,4 @@
 "xxscreeps": patch
 ---
 
-Strip `boost` from unboosted body parts so `creep.body[i]` matches the vanilla `{hits, type}` own-property shape; `boost` is only present on boosted parts.
+Strip `boost` from unboosted body parts to match vanilla own-property shape.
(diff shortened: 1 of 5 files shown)


laverdet:
I don't think we should do this.

PR #165, opened 2026-04-29
Add shard tick processors for Game.notify delivery

misterwise:
## Depends on

Depends on #161, which adds the `Game.notify` runtime queueing layer. This branch is stacked on top of that work; the diff will narrow once #161 lands.

## Summary

- Add shard-level tick processors that run after room finalization and before the next tick queue is snapshotted.
- Use the shard processor context to support async tasks, next-tick room activation, and room wake scheduling.
- Persist queued `Game.notify` entries, coalesce grouped rows, and drain due users through notification transports.
- Add default stdout delivery plus focused coverage for queueing, delivery, throttling, disabled prefs, and failure isolation.

## Not Included

- `notifyWhenAttacked` delivery is still not wired up. The existing TODO in `mods/structure/processor.ts` still needs damage processors to read `#noAttackNotify` and emit attack notifications.
- Notification retention for repeatedly failed transports is still unresolved. Successful deliveries and disabled-user rows are removed, but the existing model TODO should be tightened around transport-failed rows and eventually backed by TTL/expire or pruning.
- No user-facing preference/config API is included beyond the internal prefs model covered by tests.
- The default transport is stdout-only; production email/webhook-style delivery would plug into `sendUserNotifications`.

## Verification

- `npm run build`
- `npm test -- "ShardProcessorContext"`
- `npm test -- "Game.notify"`
- `npm test -- "Notification delivery worker"`
- `npm test -- "Default stdout transport"`


packages/xxscreeps/engine/processor/room.ts
```diff
@@ -106,9 +106,14 @@ export class RoomProcessor implements ProcessorContext {
 			}
 			this.room['#flushObjects'](this.state);
 
-			// Run `registerRoomTickProcessor` hooks
+			// Run `registerRoomTickProcessor` hooks. Each callback is isolated: a throw is logged
```
laverdet: This is "on error continue next". If there is an uncaught error it means the game has reached an invariant. Continuing is not the correct choice.

packages/xxscreeps/engine/processor/shard.ts
```diff
@@ -0,0 +1,93 @@
...
+			const tasks = this.tasks;
+			this.tasks = [];
+			const results = await Promise.allSettled(Fn.map(tasks, task => task.promise));
+			for (const [ ii, task ] of tasks.entries()) {
+				const result = results[ii];
+				if (result.status === 'fulfilled') {
+					try {
+						task.finalize?.(result.value);
+					} catch (err) {
+						console.error(`shard tick processor task finalize threw on tick ${this.time}:`, err);
```
laverdet: nope

packages/xxscreeps/engine/processor/shard.ts
```diff
@@ -0,0 +1,93 @@
...
+					console.error('shard tick processor task threw:', result.reason);
+				}
+			}
+		}
+	}
+
+	async commit() {
+		await Promise.all([
+			this.shard.scratch.zadd(activeRoomsKey,
+				[ ...Fn.map(this.activations, name => [ 0, name ] as [ number, string ]) ]),
```
laverdet: Prefer return type over `as` cast. Or you can use `as const`

packages/xxscreeps/engine/processor/shard.ts
```diff
@@ -0,0 +1,93 @@
...
+		}
+	};
+
+export async function runShardTickProcessors(shard: Shard, time: number) {
+	const ctx = new ShardProcessor(shard, time);
+	for (const fn of shardTickProcessors) {
+		try {
+			await fn(shard, ctx);
+		} catch (err) {
+			console.error(`shard tick processor threw on tick ${time}:`, err);
```
laverdet: !

packages/xxscreeps/mods/notifications/prefs.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type NotifyPrefs = {
+	disabled?: boolean;
+	disabledOnMessages?: boolean;
+	sendOnline?: boolean;
+	interval?: number;
+	errorsInterval?: number;
+};
+
+const prefsKey = (userId: string) => `user/${userId}/notifications-prefs`;
```
laverdet: Hyphens to slashes.

packages/xxscreeps/mods/notifications/prefs.ts
```diff
@@ -0,0 +1,53 @@
...
+export async function getNotifyPrefs(shard: Shard, userId: string): Promise<NotifyPrefs> {
+	interface PrefsFields {
+		disabled?: string;
+		disabledOnMessages?: string;
+		sendOnline?: string;
+		interval?: string;
+		errorsInterval?: string;
+	}
+	const fields = await shard.data.hgetall(prefsKey(userId)) as PrefsFields;
+	const out: NotifyPrefs = {};
```
laverdet:
```
return {
disabled: fields.disabled === 1,
// and so on
};
```

packages/xxscreeps/engine/processor/shard.ts
```diff
@@ -0,0 +1,93 @@
...
+	private readonly wakeUps: [ number, string ][] = [];
+	private tasks: { promise: Promise<any>; finalize: ((result: any) => void) | undefined }[] = [];
+
+	constructor(shard: Shard, time: number) {
+		this.shard = shard;
+		this.time = time;
+	}
+
+	activateRoom(roomName: string) {
+		this.activations.add(roomName);
```
laverdet: I'm kind of confused at this functionality. It doesn't appear to be used anywhere, except in the tests which only serve to ensure the unused function works.

packages/xxscreeps/engine/processor/model.ts
```diff
@@ -267,6 +268,9 @@ export async function roomsDidFinalize(shard: Shard, roomsCount: number, time: n
 		// Decrement number of finalization rooms remaining
 		const remaining = await shard.scratch.decrBy(finalizedRoomsPendingKey(time), roomsCount);
 		if (remaining === 0) {
+			// Run shard tick processors before snapshotting tick N+1's queue, so any activateRoom
+			// or wakeAt issued from a callback lands on tick N+1.
+			await runShardTickProcessors(shard, time);
```
laverdet: I think it'd make sense to move this, and the `begetRoomProcessQueue` stuff below over into the main loop in response to the tickFinished event.

packages/xxscreeps/engine/processor/test.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+const empty = simulate({
+	W0N0: () => {},
+	W1N1: () => {},
+	W2N2: () => {},
+	W3N3: () => {},
+});
+
+describe('registerShardTickProcessor', () => {
```
laverdet: None of these tests seem particularly valuable.

misterwise:
While shrinking `ShardProcessorContext` I noticed `task()` itself had no load-bearing rationale at shard scope — `RoomProcessor.task()` exists because its finalize re-enters `runWithState` / `runAsUser` so `Game`/`me`/the room are restored when the result handler runs; shard scope has no equivalent continuation, so the finalize is just `.then()` and the queue/flush machinery collapses to `Promise.all(hooks.map(fn))` at the registrar.

Collapsed it too: `runShardTickProcessors` is one line, the `ShardProcessor` class and `ShardProcessorContext` interface are gone, and the notifications call site drops its `ctx` wrapper. Split across two commits so the strict-feedback pass and the follow-on collapse are reviewable independently.

packages/xxscreeps/engine/processor/room.ts
misterwise: Agreed

packages/xxscreeps/engine/processor/shard.ts
misterwise: Was getting ahead of myself and didn't slice the refactor cleanly, will include this in the PR that actually consumes these.

laverdet:
Looking closer I think it needs another pass:
- `user/xy/notifications` should be a sorted set. Grouped notifications should be delivered by group, not by user. For example if a user schedules a 60 minute group, and a different 60 second group, the 60 second group should not trigger the longer group.
- Remove `time` parameter from shard processor
- notifications/hooks.ts should either use `makeHookRegistration` or just use a dead simple array. It shouldn't implement a new hook framework.

misterwise:
> Looking closer I think it needs another pass:
> 
> * `user/xy/notifications` should be a sorted set. Grouped notifications should be delivered by group, not by user. For example if a user schedules a 60 minute group, and a different 60 second group, the 60 second group should not trigger the longer group.
> * Remove `time` parameter from shard processor
> * notifications/hooks.ts should either use `makeHookRegistration` or just use a dead simple array. It shouldn't implement a new hook framework.

I've been trying to figure out a small slice that would implement a meaningful "cron-like" ability in xxscreeps because a lot of the remaining feature set relies on the ability to efficiently do shard level tasks with regularity (Invader Strongholds, deposits, PowerBanks. After reviewing the current hook infrastructure, a sorted set by tick that drains should work just fine so I'll pursue that approach.

laverdet:
> > Looking closer I think it needs another pass:
> 
> > 
> 
> > * `user/xy/notifications` should be a sorted set. Grouped notifications should be delivered by group, not by user. For example if a user schedules a 60 minute group, and a different 60 second group, the 60 second group should not trigger the longer group.
> 
> > * Remove `time` parameter from shard processor
> 
> > * notifications/hooks.ts should either use `makeHookRegistration` or just use a dead simple array. It shouldn't implement a new hook framework.
> 
> 
> 
> I've been trying to figure out a small slice that would implement a meaningful "cron-like" ability in xxscreeps because a lot of the remaining feature set relies on the ability to efficiently do shard level tasks with regularity (Invader Strongholds, deposits, PowerBanks. After reviewing the current hook infrastructure, a sorted set by tick that drains should work just fine so I'll pursue that approach.

Please dont overthink it!

PR #170, opened 2026-05-01
Add `xxscreeps cli` / `eval` foundation

misterwise:
Redesign in response to review. The original revision routed the CLI through engine/runtime layers to expose admin-scope `Game`; the next revision moved to a curated `node:vm` context. This version keeps the CLI as a core command surface but runs offline eval in the CLI process's host realm instead.

## What this PR is now

Adds two top-level offline commands under `packages/xxscreeps/cli/`. No engine state is exposed and no live server is required.

- **`xxscreeps cli`** — interactive REPL via `node:repl`. Vars persist across turns, top-level `await` works, multi-line input is recoverable. Bare `xxscreeps` is the same.
- **`xxscreeps eval`** — one-shot evaluator. Source comes from exactly one of `-e/--expression`, `--file`, or `--stdin`. Trailing `-- arg ...` values populate global `argv`. `--json` emits a structured envelope for tooling.

```ts
interface EvalEnvelope {
    ok: boolean;
    result?: unknown;
    logs: string[];
    warnings: string[];
    errors: string[];
    thrown?: { name: string; message: string; stack: string };
}
```

Plain mode streams console output as it happens and prints `util.inspect(result)`. Thrown errors go to stderr and exit 1. JSON mode buffers console output and the result into one envelope on stdout. Host-side failures such as argparse errors or unreadable files still print plain text to stderr.

## Realm

Offline CLI code runs in the CLI process's host realm. There is no `vm.Context` and no curated globals list. `process`, `require`, dynamic `import()`, `Buffer`, `__dirname`, `__filename`, Node built-ins, and loaded dependencies are reachable.

`eval` temporarily routes `console` through the streaming or JSON output contract and exposes `argv`. `cli` uses the host console directly.

`Game`, `Memory`, `shard`, and `db` are absent because the engine is not running. A live-mode bridge can add engine-side eval later over a socket, with sandboxing on the server side rather than in the offline client.

## Removed

Removes the old backend CLI socket REPL and bare-`xxscreeps` socket client plumbing. The offline CLI host now lives in `packages/xxscreeps/cli/`, not under `mods/`.

## Deferred to follow-ups

- **Live-mode bridge** — `--socket` support that ships source text to a running engine and evaluates there with engine-side helpers.
- **Domain commands** — `users`, `bots`, `rooms`, `objects`, `system`, `config`, `storage`, `map`, `profiler`, `import`, `export`, `backup`, `intents`, plus safety classifications and destructive-op confirmation.
- **`Manifest.commands`** — a command extension point so mods can contribute commands without hardcoding every command in `config/entry.ts`.

## Verification

- `npm run build`
- `npm test -- cli`
- CLI smoke from the xxscreeps workspace at `f69b79cf`:
  - `xxscreeps eval --json` exposes host globals: `[typeof process, typeof require, typeof Buffer, typeof Game, typeof __dirname, typeof __filename]` returned `["object","function","function","undefined","string","string"]`.
  - `xxscreeps eval --json` handled top-level `await`, console envelope capture, `argv`, non-JSON result markers, and runtime `SyntaxError` without re-running side effects.
  - Plain `xxscreeps eval` handled success and thrown-error exit paths.
  - `xxscreeps eval --file` and `xxscreeps eval --stdin` both evaluated source successfully.
  - PTY `xxscreeps cli` session verified var persistence (`41` then `42`), top-level `await` (`99`), recoverable multiline input, and `.exit`.


packages/xxscreeps/config/entry.ts
```diff
@@ -104,7 +124,7 @@ if (missingFlags.length) {
 		}
 
 	} else {
-		// Start CLI client — connects to the server's Unix socket
-		await import('./repl.js');
+		// Start CLI client
+		await import('../mods/cli/repl.js');
```
laverdet: I think it's ok if cli is not a mod. What you have here is that the cli mod is a dependency of xxscreeps, which makes it not a mod.

packages/xxscreeps/driver/runtime/index.ts
```diff
@@ -1,22 +1,47 @@
...
+import { Game, GameState, hooks, initializeGameEnvironment, runForPlayer, runWithStateAsync, userInfo } from 'xxscreeps/game/index.js';
 import { World } from 'xxscreeps/game/map.js';
 import { detach } from 'xxscreeps/schema/buffer-object.js';
 import { setupConsole } from './console.js';
 import { makeEnvironment } from './module.js';
 import { flush, print, resultPrefix } from './print.js';
+import { asyncEvalCandidates } from './source.js';
+
+declare global {
+	var Game: GameBase | undefined;
```
laverdet: This global applies to the entire project. It will definitely cause confusion if you do this.

packages/xxscreeps/driver/runtime/index.ts
```diff
@@ -75,6 +100,29 @@ let world: World;
 let loop: (() => any) | undefined;
 let requireMain: () => any;
 
+// eslint-disable-next-line no-eval
+const globalEval = eval;
```
laverdet: Is the point here to do "indirect eval"? If so just use `(0, eval)(string)` and document it as such.

packages/xxscreeps/driver/runtime/index.ts
```diff
@@ -75,6 +100,29 @@ let world: World;
...
+
+function evaluateSource(source: string) {
+	const candidates = asyncEvalCandidates(source);
+	const last = candidates.length - 1;
+	for (let ii = 0; ii < last; ii++) {
+		try {
+			return globalEval(candidates[ii]) as unknown;
+		} catch (err) {
+			if (!(err instanceof SyntaxError)) {
+				throw err;
```
laverdet: I've pointed this out before, but this is not how you do this. If you do `evaluateSource("console.log(1); JSON.parse('}');")` it will log 3 or 4 times. That is not what you want.

packages/xxscreeps/mods/cli/eval.ts
```diff
@@ -0,0 +1,27 @@
...
+interface Argv {
+	source: string[];
+}
+
+function parseArgs<Type>(parser: ArgumentParser) {
+	return parser.parse_args() as Type;
+}
+
+const parser = new ArgumentParser({
+	prog: 'xxscreeps eval',
```
laverdet: Maybe just make it an argument on `xxscreeps cli`, matching `node -e "whatever"`

laverdet (commented):
I think you're making your life harder than it needs to be with all the sandboxing code. Is there a compelling reason to run this in the sandbox as opposed to the client's local realm?

packages/xxscreeps/config/entry.ts
misterwise: Yeah, I was trying to hard to make this into a mod. I'm going to refocus this with core CLI built into xxscreeps and then mods having the ability to register commands based on their surface, I think that should be interesting and a useful way to break things up.

misterwise:
> I think you're making your life harder than it needs to be with all the sandboxing code. Is there a compelling reason to run this in the sandbox as opposed to the client's local realm?

Yeah, sandbox was originally how it blended both live and offline, I was thinking this was the right way to approach it, but it doesn't offer us anything anymore, and this is the offline version of the CLI first anyways.

packages/xxscreeps/mods/cli/eval.ts
misterwise: I think these are different enough surfaces it may be worth keeping them separate, that way we don't have flags exposed on cli that only make sense on the one-shot eval surface.

misterwise:
This is still not it, the offline version doesn't need sandbox, I've just swapped one for another here. The live version can use one to sandbox engine, trying to keep the split clean so I have good reuse of code.

misterwise:
Ok that's better now, and updated body. Ready for another review, hopefully closer to a solid implementation?

laverdet:
I removed the `--json` option because it really doesn't seem compelling and it's a ton of code for nothing.

I think the focus of the cli should be quick tinkering. I imagine for serious administration you would just collect a bunch of scripts for the tasks that you want to do. A big difference between screeps and xxscreeps is that you can just `import { Shard } from 'xxscreeps/engine/db/index.js'` in a script and run it directly in nodejs. So where the screeps CLI is *required* for server administration, in xxworld it is just fun to have around.

I'll be following up with a refactor to the eval stuff.

Thank you for your patience.

misterwise:
> I removed the `--json` option because it really doesn't seem compelling and it's a ton of code for nothing.
> 
> I think the focus of the cli should be quick tinkering. I imagine for serious administration you would just collect a bunch of scripts for the tasks that you want to do. A big difference between screeps and xxscreeps is that you can just `import { Shard } from 'xxscreeps/engine/db/index.js'` in a script and run it directly in nodejs. So where the screeps CLI is _required_ for server administration, in xxworld it is just fun to have around.
> 
> I'll be following up with a refactor to the eval stuff.
> 
> Thank you for your patience.

Ok thanks, so I'll hold off on a PR for adding live CLI until I see the eval refactor land.

laverdet:
it is here 3b7f22001d1aa19ee9d89aca1f25452d4d22541f

PR #173, opened 2026-05-02
Fix client Steam package discovery

misterwise:
## Summary
  This PR changes @xxscreeps/client to discover package.nw automatically from Steam install locations, including secondary library folders, while preserving existing behavior for explicit browserClient.package configuration.

  It addresses the common failure mode where a Steam install is not in one of the previously hardcoded paths, so the client can no longer find the package on systems with custom/secondary libraries.

  ## Compatibility / behavior notes

  - Existing browserClient.package override continues to work.
  - No behavior is changed for users that already configure the path explicitly.
  - Startup diagnostics now reflect all attempted search locations when discovery fails.

### DIFF
--- a/.changeset/client-package-nw-discovery.md
+++ b/.changeset/client-package-nw-discovery.md
@@ -1,0 +1,5 @@
+---
+"@xxscreeps/client": patch
+---
+
+Auto-locate `package.nw` across Steam install locations and secondary library folders.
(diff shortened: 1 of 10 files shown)


laverdet:
I just pulled most of the implementation from the steamless client fork. 5dad80f84ec5e5bad63045833095b33e1b0e0e86

PR #174, opened 2026-05-02
Fix `xxscreeps import` mod auto-detection from project package.json

misterwise:
`xxscreeps import` is supposed to seed the generated `.screepsrc.yaml` with mods declared in the project's `package.json`, but the seed never opens the project's package.json — it opens xxscreeps's own.

The seed is `fetch('.', 2)` in `packages/xxscreeps/scripts/scrape-world.ts`, which resolves `.` via `import.meta.resolve('.')`. That anchors on the calling module's URL — xxscreeps's own package — so the walk-up looking for
`package.json` lands on `packages/xxscreeps/package.json`. The user's project package.json is never seen; mods declared there (e.g. `@xxscreeps/client`) are silently dropped from the generated `.screepsrc.yaml`.

Anchor `.` on `pathToFileURL(`${process.cwd()}/`)` so the walk starts at the project. Non-`.` specifiers keep using `import.meta.resolve` unchanged — Node's parent-dir walk still reaches the user's `node_modules` from inside
xxscreeps's installed location, so the recursive step works under npm, yarn, and pnpm.

## Validation

Built the dist locally, ran `xxscreeps import` from a fresh temp dir whose only file is a `package.json` listing `xxscreeps` and `@xxscreeps/client` as deps.

- Before: generated `.screepsrc.yaml` lists xxscreeps's stock mods only.
- After: `@xxscreeps/client` appears in the mods list (emitted as `- '@xxscreeps/client'` since js-yaml quotes leading `@`).

Two consecutive runs in fresh temp dirs produce identical mods lists (only `backend.secret` differs, expected).


packages/xxscreeps/scripts/scrape-world.ts
```diff
@@ -92,8 +93,10 @@ if ((rcInfo?.size ?? 0) === 0) {
 		}
 		fetched.add(specifier);
 		try {
-			// Find `package.json` for this specifier
-			const indexPath = new URL(import.meta.resolve(specifier));
+			// Find `package.json` for this specifier. Anchor `.` on cwd;
+			const indexPath = specifier === '.'
+				? pathToFileURL(`${process.cwd()}/`)
```
laverdet: `pathToFileURL(process.cwd() + path.sep)` otherwise it might break on windows

PR #175, opened 2026-05-02
Pin the Node version to v24.14.1 since that's what the GH runner default to

tiennou:


laverdet:
Is there a reason to pin this strictly on a specific nodejs patch version?

tiennou:
No, I just picked the version CI is at. With those changes you could set it at anything and CI should change as well.

laverdet:
Thanks, I merged without the more restrictive nvmrc

PR #176, opened 2026-05-02
Fix LOOK_ENERGY room look alias

misterwise:
## Summary

- Add generic room look alias registration.
- Register `LOOK_ENERGY` from the resource mod as an alias of `LOOK_RESOURCES` for dropped energy resources.
- Make `lookForAt(LOOK_ENERGY, ...)`, `lookForAtArea(LOOK_ENERGY, ...)`, `lookAt(...)`, and `lookAtArea(...)` expose the legacy energy alias consistently.

The resource mod declares the concrete alias because it owns `Resource`, `LOOK_RESOURCES`, and `RESOURCE_ENERGY`. The generic room look layer only resolves registered aliases, which keeps resource-specific constant knowledge out of room lookup internals.

## Validation

Validated against screeps-ok:

- `ROOM-LOOK-007 lookForAt(LOOK_ENERGY) returns the same Resource as LOOK_RESOURCES`
- `ROOM-LOOK-008 lookForAtArea(LOOK_ENERGY) returns the same Resource shaped under the energy key`
- `ROOM-LOOK-009 lookAt yields both energy and resource entries on a dropped-resource tile`

### DIFF
--- a/packages/xxscreeps/game/object.ts
+++ b/packages/xxscreeps/game/object.ts
@@ -56,6 +56,8 @@ export abstract class RoomObject extends withOverlay(BufferObject.BufferObject, 
 	get '#pathCost'(): undefined | number { return undefined; }
 	// eslint-disable-next-line @typescript-eslint/class-literal-property-style
 	get '#providesVision'() { return false; }
+	// eslint-disable-next-line @typescript-eslint/class-literal-property-style
+	get '#secondaryLookType'(): string | null { return null; }
 	get '#user'(): string | null { return null; }
 	abstract get ['#lookType'](): string | null;
 
--- a/packages/xxscreeps/game/room/index.ts
+++ b/packages/xxscreeps/game/room/index.ts
@@ -8,5 +8,5 @@ export type { Look } from './look.js';
 export type { AnyRoomObject } from './room.js';
 
 export { Room } from './room.js';
-export { registerFindHandlers, registerLook } from './symbols.js';
+export { registerFindHandlers, registerLook, registerLookAlias } from './symbols.js';
 export interface Schema {}
--- a/packages/xxscreeps/game/room/index.ts
+++ b/packages/xxscreeps/game/room/index.ts
@@ -8,5 +8,5 @@ export type { Look } from './look.js';
 export type { AnyRoomObject } from './room.js';
 
 export { Room } from './room.js';
-export { registerFindHandlers, registerLook, registerLookAlias } from './symbols.js';
+export { registerFindHandlers, registerLook } from './symbols.js';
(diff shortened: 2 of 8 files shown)


laverdet:
Unless I'm missing something it doesn't look this test tests what it says it tests?
https://github.com/misterwise/screeps-ok/blob/4f976ca557a56defc5ed66f872e1c776d4cc349b/tests/16-room-mechanics/16.4-look.test.ts#L153-L171

The only resource placed is energy, so it doesn't test what LOOK_ENERGY does when it encounters ghodium or whatever.

There's complaints in the forum about LOOK_ENERGY returning resources. From what I see in the engine code that seems to be the case: https://screeps.com/forum/topic/788/room-lookforatarea-look_energy-returns-resources

Anyway I think it might be better to implement it one of these ways:
```ts
	get ['#secondaryLookType'](): string | null { return null; }
	abstract get ['#lookType'](): string | null;
```

-or-

```ts
	get ['#lookTypeSlot'](): string | null { return this['#lookType']; }
	abstract get ['#lookType'](): string | null;
```

Which option is better depends on the exact behavior of LOOK_ENERGY vs LOOK_RESOURCE

misterwise:
> Unless I'm missing something it doesn't look this test tests what it says it tests? https://github.com/misterwise/screeps-ok/blob/4f976ca557a56defc5ed66f872e1c776d4cc349b/tests/16-room-mechanics/16.4-look.test.ts#L153-L171
> 
> The only resource placed is energy, so it doesn't test what LOOK_ENERGY does when it encounters ghodium or whatever.
> 
> There's complaints in the forum about LOOK_ENERGY returning resources. From what I see in the engine code that seems to be the case: https://screeps.com/forum/topic/788/room-lookforatarea-look_energy-returns-resources
> 
> Anyway I think it might be better to implement it one of these ways:
> 
> ```ts
> 	get ['#secondaryLookType'](): string | null { return null; }
> 	abstract get ['#lookType'](): string | null;
> ```
> 
> -or-
> 
> ```ts
> 	get ['#lookTypeSlot'](): string | null { return this['#lookType']; }
> 	abstract get ['#lookType'](): string | null;
> ```
> 
> Which option is better depends on the exact behavior of LOOK_ENERGY vs LOOK_RESOURCE

Yes, it should have been made clearer. In vanilla LOOK_ENERGY is a direct ALIAS of LOOK_RESOURCES, they have exactly the same behaviour. This is counterintuitive, hence the 10 years of complaints, but LOOK_ENERGY existed before LOOK_RESOURCES was an idea. Really it should be deprecated at this point, but I see no evidence of that happening. The correct shape is #secondaryLookType

PR #177, opened 2026-05-03
Tighten register endpoint input checks

misterwise:
## Summary

This tightens validation in the built-in registration endpoints so request/query values are checked before being passed into user lookup and creation helpers.

Changes:
- Require `check-username` query input to be a string instead of stringifying arbitrary values.
- Use explicit `null` checks for user lookup results.
- Validate `set-username` body shape before calling `User.checkUsername()` or `validateEmail()`.
- Only attach the email provider when the submitted email is a non-empty string.


laverdet:
I pushed c51b0f9c8be12813625b80ce95e84f305b29f6da which has a bunch of validation of backend requests. I think I fixed the blank email too thing, I'm not sure.

PR #178, opened 2026-05-03
Fix id-string object constructors

misterwise:
## Summary

- Centralize RoomObject id-string construction on the canonical `Game.getObjectById(id)` object and intentionally drop the constructor-path `instanceof` validation to match vanilla, including `new Creep(structureId)`.
- Add a `ReadOnlyView` sentinel for id-constructed views so writes to overlay-defined fields are explicitly discarded while ad-hoc own-property writes still work.
- Remove the duplicate placeholder `Ruin` runtime global. That placeholder shadowed the real structure mod `Ruin` class in isolated player runtime; Tombstone did not have that duplicate, which is why Tombstone id construction already worked.

## Notes

The exported `getById(Type, id)` helper remains strict; only the constructor id-string branch uses vanilla's direct lookup behavior.

The id-constructed view still shares the canonical buffer, but public fields are exposed through read-only getters so cross-type views read the canonical object's public values and requested-type-only fields read `undefined` instead of stale offset data.

## Validation

- `pnpm run build`
- `pnpm test` -> 200/200 passing
- `pnpm exec eslint --max-warnings 0` on changed files: no new warnings; the command still reports the repo's existing warnings in those files
- `screeps-ok-pr` parity now passes the id-constructor coverage that previously failed:
  - `tests/27-undocumented/27.11-id-constructors.test.ts`: `UNDOC-IDCTOR-001`, `UNDOC-IDCTOR-003`, `UNDOC-IDCTOR-004`
  - `tests/27-undocumented/27.6-identity.test.ts`: `UNDOC-CTOR-002`, `UNDOC-CTOR-003`, `UNDOC-CTOR-004`, `UNDOC-CTOR-006`, `UNDOC-CTOR-007`, `UNDOC-CTOR-008`
  - The parity reporter still exits non-zero until `screeps-ok` removes the stale expected-failure registrations for the two now-fixed gaps.

### DIFF
--- a/.changeset/id-ctor-overlay-share.md
+++ b/.changeset/id-ctor-overlay-share.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Fix `new Creep(id)` and other id-string constructors to expose primitive overlay fields like `hits` and `fatigue`.
--- a/packages/xxscreeps/game/object.ts
+++ b/packages/xxscreeps/game/object.ts
@@ -3,15 +3,17 @@ import type { RoomPosition } from './position.js';
 import type { Room } from './room/index.js';
 import type { InspectOptionsStylized } from 'node:util';
 import type { BufferView, TypeOf } from 'xxscreeps/schema/index.js';
-import { inspect } from 'node:util';
 import * as Id from 'xxscreeps/engine/schema/id.js';
 import { enumeratedForPath } from 'xxscreeps/engine/schema/index.js';
 import * as BufferObject from 'xxscreeps/schema/buffer-object.js';
 import { compose, declare, enumerated, struct, union, vector, withOverlay } from 'xxscreeps/schema/index.js';
+import { ReadOnlyView } from 'xxscreeps/schema/overlay.js';
 import { expandGetters } from 'xxscreeps/utility/inspect.js';
 import { assign } from 'xxscreeps/utility/utility.js';
 import { format as roomPositionFormat } from './position.js';
 import { Game, registerGlobal } from './index.js';
+
+const getPrototypeOf = Object.getPrototypeOf as (value: object) => object | null;
 
 // eslint-disable-next-line @typescript-eslint/no-empty-object-type
 export interface Schema {}
@@ -30,6 +32,11 @@ export abstract class RoomObject extends withOverlay(BufferObject.BufferObject, 
 	 */
 	declare room: Room;
 
+	/** Set on id-string-constructed views to the canonical handle from the registry. */
+	declare '#liveHandle'?: RoomObject;
+
+	/** @internal */
+	constructor(id: string);
 	/** @internal */
 	constructor(buffer?: BufferView, offset?: number);
 	/** @deprecated */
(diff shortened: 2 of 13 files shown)


laverdet:
I pushed 73d1a265456b1cd0b06b878990f08fecd393583d which is a different approach using `Object.setPrototypeOf`.

I think your tests around "writes on a view do not propagate to the canonical handle" were flawed. All registry-backed properties in vanilla are read-only. So even a simple test of `Game.creeps.Emily.hits = 123; assert(Game.creeps.Emily.hits !== 123)` would fail, I think. It would be simple enough to install the overlay handlers without write-back functionality but I'm not sure it's worth it.

misterwise:
@laverdet — fair point on the framing: that internal test doesn't isolate view-vs-canonical separation from read-only-getter semantics, since vanilla rejects writes on either handle.

Ran the four `UNDOC-IDCTOR-*` parity tests from [screeps-ok](https://github.com/misterwise/screeps-ok) against `73d1a26` (vanilla via `screeps-server-mockup`, xxscreeps via the commit). Vanilla: 11/11 sub-cases pass. xxscreeps under `73d1a26`: 1/11. Three divergences:

**IDCTOR-001** (8 sub-cases, every RoomObject type): `view.room` is `undefined`. Player-code pattern that crashes:
```js
const target = new Creep(Memory.targetId);
target.room.find(FIND_HOSTILE_CREEPS);  // TypeError: target.room is undefined
```
Root cause: the id-string branch builds the buffer view and re-points the prototype but never copies `this.room = object.room`. One-line fix in the same branch.

**IDCTOR-003** (`new Creep(structureId)`): the `instanceof new.target` guard rejects wrong-type ids and falls through to `super()` empty, so the view has no buffer and `view.id` is `null`. Vanilla doesn't type-check (`@screeps/engine/src/game/creeps.js:60-66`). Bots holding a generic "target id" and duck-typing the result —
```js
const view = new Creep(Memory.targetId);
if (view.fatigue !== undefined) { /* creep */ }
else if (view.hits) { /* structure-shaped */ }
```
— misclassify under `73d1a26` because neither branch reflects the underlying object's data. Fix: drop the `instanceof` guard.

**IDCTOR-004** (writes on view): `overlay.ts`'s lazy-init setter still promotes `view.hits = 50` to a writable own property, so the view returns the patched value rather than the canonical — the divergence your test comment already flags. Needs the `ReadOnlyView` flag (or equivalent gate on the overlay setter) to close.

001 and 003 read like incidental bugs — the surrounding ctor in `73d1a26` is clearly aiming for vanilla parity, those two just got missed. 004 is the genuine "is this worth it" call.

laverdet:
> IDCTOR-003

Have you encountered user code in the wild which does this? When it comes to undocumented features the goal is to start as restrictive as possible and implement more and more deranged workarounds as necessary. I see these behaviors as something that will reduce my freedom to tweak the engine going forward. Indeed, it's very difficult for vanilla Screeps to change anything about the engine because of the user code relying on things which were not locked down.

> IDCTOR-004

What I'm saying is that it is a general divergence with the vanilla engine, it has nothing to do with id-construction. In xxscreeps you can set registry-backed properties on any object, regardless of whether or not the id-constructor is invoked. `installReadOnlyView` in your PR only modifies the behavior for id-constructed objects, but does not affect it for naturally constructed objects. So that fix doesn't make sense because the issue would still be there. The actual fix would be threading a "readOnly" parameter through to injectGetters, but I don't think we should do it either.

misterwise:
> > IDCTOR-003
> 
> Have you encountered user code in the wild which does this? When it comes to undocumented features the goal is to start as restrictive as possible and implement more and more deranged workarounds as necessary. I see these behaviors as something that will reduce my freedom to tweak the engine going forward. Indeed, it's very difficult for vanilla Screeps to change anything about the engine because of the user code relying on things which were not locked down.
> 
> > IDCTOR-004
> 
> What I'm saying is that it is a general divergence with the vanilla engine, it has nothing to do with id-construction. In xxscreeps you can set registry-backed properties on any object, regardless of whether or not the id-constructor is invoked. `installReadOnlyView` in your PR only modifies the behavior for id-constructed objects, but does not affect it for naturally constructed objects. So that fix doesn't make sense because the issue would still be there. The actual fix would be threading a "readOnly" parameter through to injectGetters, but I don't think we should do it either.

No user code, and I agree that these would be burdensome to implement. I had inferred that these were in support of parity around ID based constructors (which I don't even use myself) but wasn't able to assess their priority. I think I get the point around reducing freedom to actually improve, I hadn't considered that angle myself yet. I will remove these from the parity check, and only focus on confirmed breaking undocumented surface going forwards.

Issue #117, opened 2026-04-09
Audit intent check ordering across all mods to match vanilla


misterwise:
During review of #114, @laverdet noted that the error check ordering in `StructureFactory.prototype.produce` didn't match vanilla, and that "most of them are wrong" across the codebase. The factory was fixed in #114, but every other intent using `chainIntentChecks` should be audited against the vanilla engine source.

Ordering only affects which error code a player sees when multiple conditions fail simultaneously — low severity but a correctness gap.

### Checklist

- [x] **factory** — `checkProduce` (fixed in #114)
- [ ] **chemistry/lab** — `checkRunReaction`, `checkBoostCreep`, `checkReverseReaction`, `checkUnboostCreep`
- [ ] **market/terminal** — `checkSend`
- [ ] **logistics/link** — `checkTransferEnergy`
- [ ] **defense/tower** — `checkTower` (attack/heal/repair)
- [ ] **observer** — `checkObserveRoom`
- [ ] **spawn** — `checkSpawnCreep`, `checkRenewCreep`, `checkRecycleCreep`
- [ ] **controller** — `checkActivateSafeMode`, `checkUnclaim`
- [ ] **structure** — `checkDestroy`
- [ ] **combat/creep** — `checkAttack`, `checkRangedAttack`, `checkRangedMassAttack`, `checkHeal`, `checkRangedHeal`
- [ ] **construction/creep** — `checkBuild`, `checkDismantle`, `checkRepair`
- [ ] **controller/creep** — `checkAttackController`, `checkClaimController`, `checkReserveController`, `checkUpgradeController`, `checkGenerateSafeMode`, `checkSignController`
- [ ] **harvestable/creep** — `checkHarvest`
- [ ] **creep** — `checkDrop`, `checkMove`, `checkPull`, `checkPickup`, `checkTransfer`, `checkWithdraw`
- [ ] **construction/room** — `checkCreateConstructionSite`, `checkRemove`
- [ ] **flag** — `checkCreateFlag`

misterwise:
All checklist rows are complete. The last open row (construction/creep) resolved with #215, which landed the effects substrate and the dismantle EFFECT_INVULNERABILITY -> ERR_INVALID_TARGET clause, ordered after safe mode to match vanilla.

PR #181, opened 2026-05-04
Validate observer roomName before RCL gate

misterwise:
## Summary

`StructureObserver.observeRoom` returned `ERR_RCL_NOT_ENOUGH` before `ERR_INVALID_ARGS` when both an inactive observer and a malformed `roomName` applied. Vanilla validates the room name before the RCL gate. This PR hoists the room-name validation while leaving range validation after RCL.

Part of #117 (`observer` row, `StructureObserver.observeRoom`).

## Vanilla precedence

`@screeps/engine/src/game/structures.js` `StructureObserver.prototype.observeRoom`:

```js
if(!_.isString(roomName) || !/^(W|E)\d+(S|N)\d+$/.test(roomName)) {
    return C.ERR_INVALID_ARGS;
}
if(!utils.checkStructureAgainstController(...)) {
    return C.ERR_RCL_NOT_ENOUGH;
}
```

Order: `NOT_OWNER -> INVALID_ARGS -> RCL_NOT_ENOUGH -> NOT_IN_RANGE`.

## Before

`packages/xxscreeps/mods/observer/observer.ts` `observeRoom`:

```ts
() => checkMyStructure(this, StructureObserver),
() => checkIsActive(this),              // RCL_NOT_ENOUGH before roomName validation
() => checkObserveRoom(this, roomName),
```

## After

```ts
() => checkMyStructure(this, StructureObserver),
() => checkObserveRoomName(roomName),   // INVALID_ARGS now precedes RCL
() => checkIsActive(this),
() => checkObserveRoomRange(this, roomName),
```

`checkObserveRoom` composes the room-name and range checks for processor-side revalidation.

## Validation

- New regression test `observer_illegal_arg_before_min_level` in `packages/xxscreeps/mods/observer/test.ts`. Confirmed failing before the fix (`-14 !== -10`), passing after.
- `pnpm run build` clean. `npx xxscreeps test` 209/209 pass.
- Changed-file ESLint clean with `--max-warnings 0`.
- Verified against screeps-ok from `screeps-ok-pr`: `OBSERVER-007:invalidArgsBeforeRcl` now passes and is reported as a fixed expected-failure gap.

### DIFF
--- a/packages/xxscreeps/mods/observer/observer.ts
+++ b/packages/xxscreeps/mods/observer/observer.ts
@@ -20,10 +20,7 @@ export class StructureObserver extends withOverlay(OwnedStructure, shape) {
 
 	observeRoom(roomName: string) {
 		return chainIntentChecks(
-			() => checkMyStructure(this, StructureObserver),
-			() => checkObserveRoomName(roomName),
-			() => checkIsActive(this),
-			() => checkObserveRoomRange(this, roomName),
+			() => checkObserveRoom(this, roomName),
 			() => intents.save(this, 'observeRoom', roomName));
 	}
 }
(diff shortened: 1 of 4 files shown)


packages/xxscreeps/mods/observer/observer.ts
```diff
@@ -49,11 +50,13 @@ registerBuildableStructure(C.STRUCTURE_OBSERVER, {
 
 //
 // Intent checks
-export function checkObserveRoom(observer: StructureObserver, target: string) {
+export function checkObserveRoomName(target: string) {
+	return /^(W|E)\d+(S|N)\d+$/.test(target) ? C.OK : C.ERR_INVALID_ARGS;
```
laverdet: `Game.map.getRoomStatus(target) ? C.OK : C.ERR_INVALID_ARGS`

It might also have a use for StructureTerminal which has similar roomName arguments.

packages/xxscreeps/mods/observer/processor.ts
```diff
@@ -12,7 +12,7 @@ declare module 'xxscreeps/engine/processor/index.js' {
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 const intents = [
 	registerIntentProcessor(StructureObserver, 'observeRoom', {}, (observer, context, target: string) => {
-		if (checkObserveRoom(observer, target) === C.OK) {
+		if (checkObserveRoomName(target) === C.OK && checkObserveRoomRange(observer, target) === C.OK) {
```
laverdet: Actually, `checkObserveRoom` should be broken out into its own function with all the intent checks. The existing code may be vulnerable to intent injection.

misterwise: Understood, this will inform the shape of many of these intent ordering touchups.

packages/xxscreeps/mods/observer/observer.ts
misterwise: Will make sure this is the pattern there as well.

PR #194, opened 2026-05-05
Fix FIND_HOSTILE_STRUCTURES and add simple tests for structure finds.

### LINKED ISSUE #193: room.find(FIND_HOSTILE_STRUCTURES) returns unowned structures
The saga of unowned structures continues! This time I was noticing my bot was destroying all roads, containers and walls on RCL up. The reason: `Game.rooms.W7N3.find(FIND_HOSTILE_STRUCTURES)` returns all unowned structures as well, like for example
```
  StructureRoad {
    hits: 4300,
    hitsMax: 5000,
    id: 'beb962d243a477808accbaef',
    pos: [RoomPosition W7N3 {32, 18}],
    room: [Room W7N3],
    structureType: 'road',
    ticksToDecay: 360
  },
```

I might just dig into this myself to get more familiar with the code and submit a PR myself.

Mirroar:
Fixes #193

I've added some tests as well so we'll hopefully catch future regressions on this. I hope everything is where it needs to be, if not I'll gladly adjust things.

### DIFF
--- a/packages/xxscreeps/mods/structure/game.ts
+++ b/packages/xxscreeps/mods/structure/game.ts
@@ -27,7 +27,9 @@ const find = registerFindHandlers({
 	[C.FIND_MY_STRUCTURES]: room =>
 		room['#lookFor'](C.LOOK_STRUCTURES).filter(structure => structure.my) as OwnedStructure[],
 	[C.FIND_HOSTILE_STRUCTURES]: room =>
-		room['#lookFor'](C.LOOK_STRUCTURES).filter(structure => structure instanceof OwnedStructure && !structure.my) as OwnedStructure[],
+		room['#lookFor'](C.LOOK_STRUCTURES).filter(structure => {
+			return structure instanceof OwnedStructure && structure['#user'] !== null && !structure.my;
+		}) as OwnedStructure[],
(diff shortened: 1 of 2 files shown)


misterwise:
I may be wrong but I think you also want to check owner otherwise an unowned controller would be returned, something like:

  [C.FIND_HOSTILE_STRUCTURES]: room =>
  	room['#lookFor'](C.LOOK_STRUCTURES)
  		.filter((structure): structure is OwnedStructure =>
  			structure instanceof OwnedStructure &&
  			structure.my === false &&
  			structure.owner !== undefined),

Mirroar:
You're right, I missed that case. I'll push a fix when I get it to work.

Mirroar:
I think I have it fixed for now. I was a bit confused because within the find filter, I don't see the structure's `owner` property and have to go via `#user` instead. But both the unit tests, as well as some testing ingame, comfirm that this seems to fix the problem.

laverdet:
This is not the correct fix https://github.com/laverdet/xxscreeps/issues/193#issuecomment-4392572648

laverdet:
6cabc5cb1a338618b9f657302cea0fb7d24317a6

I pulled the tests from this PR, thank you.

PR #212, opened 2026-05-10
Send attack notifications for owned creeps and structures

misterwise:
## Summary

- Vanilla parity for attack notifications (`engine/utils.js#sendAttackingNotification`): when an owned creep or structure takes damage, write a `(user, type, message)` row that increments `count` on repeat events. No time bucketing — one row per message, forever.
- Wires the dispatcher into combat (`attack`/`rangedAttack`/`rangedMassAttack`), defense (tower `attack`), and construction (`dismantle`). Recipient resolves to owner with controller-owner fallback for structures; self-attacks and NPC sources/recipients (`'2'`/`'3'`) suppressed.
- Implements `Creep.notifyWhenAttacked` (was a no-op stub), with vanilla check ordering: not-owner → spawning → non-bool. Adds `'#noAttackNotify'` to the creep schema (default off → notify on).
- New `sendNotification` model helper backs the dispatcher with a single Lua script for atomic read-modify-write; the existing `upsertNotification` (used by `Game.notify`) gets the same atomicity treatment so same-tick events on the same row can't race.

PowerCreep is intentionally out of scope — the mod isn't in xxscreeps yet.

## Notes

- `Game.notify` keeps its time-bucketing `upsertNotification` path (its API exposes `groupInterval`); attack notifications use the new `sendNotification` helper to match vanilla's single-row-per-message semantics. Two helpers, intentionally different.
- NPC-source filter (`sourceUser` `'2'`/`'3'`) and NPC-recipient filter (`userId` `'2'`/`'3'`) aren't unit-tested directly: the `simulate` harness treats short user IDs as system users without per-user room visibility. Both filters are explicit and adjacent in `mods/notifications/processor.ts`; covered by inspection.

## Validation

- `pnpm run build` — clean
- `npx xxscreeps test` — 382/382 pass; combat suite covers melee, ranged, mass-attack, tower, dismantle, NPC source suppression, self-attack suppression, controller fallback, and per-event count increments
- `npx eslint` on changed files — 3 warnings, all pre-existing on `main`; 0 new
- screeps-ok targeted (`ATTACK-NOTIFY-*` + `STRUCTURE-API-00{4,5,6}`): 4 pass (`STRUCTURE-API-004`/`005`, `ATTACK-NOTIFY-005`/`006`), 8 fail — all hit the missing `notifiesWhenAttacked()` getter, a separate `attack-notify-getter-api-missing` gap shared with vanilla and not in this PR's scope. PowerCreep tests skipped. Creep `ATTACK-NOTIFY-002` now advances past its `notifyWhenAttacked(false) === OK` assertion (was `null` pre-PR).


packages/xxscreeps/mods/combat/creep.ts
```diff
@@ -171,3 +175,46 @@ export function checkDestructible(target: Creep | Structure) {
...
+) {
+	target['#applyDamage'](power, type, source ?? undefined);
+	notifyAttackDamage(target, context, source);
+}
+
+/**
+ * Like `captureDamage`, but also notifies any intermediate layer object (e.g. a rampart) that
+ * absorbed some damage on the way to `target`.
+ */
+export function captureDamageWithNotify(
```
laverdet, line 198: Copy/pasting this function definitely isn't the way. You can add a callback to the existing function.

packages/xxscreeps/mods/combat/test.ts
```diff
@@ -213,6 +219,253 @@ describe('getEventLog missing events', () => {
...
 
+describe('Structure.notifyWhenAttacked attack notifications', () => {
+	const user = '100';
+	const attackerUser = '101';
+	const sim = simulate({
+		W1N1: room => {
+			room['#level'] = 7;
+			room['#user'] = room.controller!['#user'] = user;
+			room['#insertObject'](createLab(new RoomPosition(25, 25, 'W1N1'), user));
+			room['#insertObject'](createCreep(
```
laverdet: Please just throw these all on one line. Creeps don't need to be 5 lines each.

packages/xxscreeps/mods/creep/creep.ts
```diff
@@ -76,6 +76,7 @@ const shape = struct(objectFormat, {
 	store: openStoreFormat,
 	'#actionLog': actionLogFormat,
 	'#ageTime': 'int32',
+	'#noAttackNotify': 'bool',
```
laverdet, line 79: Good job defaulting to false 👍🏻

packages/xxscreeps/mods/notifications/processor.ts
```diff
@@ -1,11 +1,48 @@
...
 import { DEFAULT_INTERVAL_MIN, getLastNotifyDate, getNotifyPrefs, setLastNotifyDate } from './prefs.js';
 import { transports } from './transports.js';
 import './transport-stdout.js';
 
+interface NamedStructure extends Structure {
+	name: string;
+}
+
+function isNamedStructure(structure: Structure): structure is NamedStructure {
+	return 'name' in structure && typeof structure.name === 'string';
```
laverdet: I'm not a fan of these runtime `in` checks. I know that's what TypeScript pushes you to do but the approach seems wrong. Maybe try accepting something like this instead.

```ts
interface MaybeNamedStructure extends Structure {
  name?: unknown;
}
```

packages/xxscreeps/mods/structure/structure.ts
```diff
@@ -305,4 +326,8 @@ declare module 'xxscreeps/game/game.js' {
 		structures: Record<string, AnyStructure>;
 	}
 }
-hooks.register('gameInitializer', Game => Game.structures = Object.create(null));
+hooks.register('gameInitializer', Game => {
+	const structures: Record<string, AnyStructure> = {};
+	Object.setPrototypeOf(structures, null);
```
laverdet: Definitely do not use `Object.setPrototypeOf` for this.

packages/xxscreeps/mods/notifications/processor.ts
```diff
@@ -1,11 +1,48 @@
...
+}
+
+registerAttackNotification((context, target, source) => {
+	if (!(target instanceof Creep || target instanceof Structure)) {
+		return;
+	}
+	const { label, userId } = describeTarget(target);
+	const sourceUser = source?.['#user'];
+	if (
+		userId !== undefined && userId !== null
```
laverdet: You can combine them with `!= null` (the one exception to triple-equal requirement)

packages/xxscreeps/mods/notifications/model.ts
```diff
@@ -108,3 +109,48 @@ export async function upsertNotification(
...
+
+// Atomic upsert for engine-fired rows. Same-tick events on the same row (e.g. two attackers in
+// one room) race the read-then-write pattern in `upsertNotification`; the runner serializes
+// `Game.notify`, but the processor fans out parallel `context.task` promises.
+const RecordEngineNotification = new KeyvalScript((
+	keyval,
+	[ indexKey, key, dueKey ]: [ string, string, string ],
+	[ id, userId, message, type, score ]: [ string, string, string, NotificationType, number ],
+) => {
+	const existing = keyval.hGet(key, 'count');
```
laverdet: You can just fix the race in `upsertNotification` without LUA. Use `keyval.hSet(key, 'count', 1, { if: 'NX' })`. The zadd commands are idempotent so they can be fired unconditionally. Then, if the result of the `set` was undefined you can dispatch an additional hincrBy. That speeds up the existing function in the optimistic case by removing an additional `await`.

laverdet (changes requested):

packages/xxscreeps/mods/creep/creep.ts
misterwise: Thank you, I do listen! :)

packages/xxscreeps/mods/notifications/model.ts
misterwise: This cleaned things up a lot.

misterwise:
Pushed `3b3f6b53`. I merged current `main` in first — comments 2 & 3 land on the upstream notifications restructure (dispatcher now in `notifications/processor.ts`), so the fixes sit on top of it rather than the old layout.

- **`captureDamage` callback** — it now takes an optional `onCapture(object)` invoked for each intermediate layer that absorbs damage; `captureDamageWithNotify` is a one-line wrapper that supplies the notify. No more copied `#lookAt`/`walkLayers` body.
- **`MaybeNamedStructure`** — dropped the `in` guard for `interface MaybeNamedStructure extends Structure { name?: unknown }` + a `typeof name === 'string'` value check.
- **`!= null`** — done.
- **`Object.setPrototypeOf`** — reverted to `Object.create(null)`.
- **Lua upsert** — removed. `count` and `date` claim their slots with `hSet … NX`, the zadds fire unconditionally, and only an already-present row pays a follow-up `hincrBy`, so the optimistic write is a single round trip. `upsertNotification` and `sendNotification` now share one `recordNotification` helper — the engine path no longer needs its own atomic primitive. (`date` keeps the first occurrence via a second NX write; easy to let it track the latest instead if you'd rather.)
- **Test fixtures** — one line each.

One knock-on: with the read-then-write race gone, `dispatchQueuedNotifications` fans out with `Fn.mapAwait` instead of the sequential loop it used to need.

PR #213, opened 2026-05-10
Notify controller lifecycle events

misterwise:
## Summary

Adds the three controller lifecycle notifications via the `notifications` mod:
level-up, pre-downgrade warning when `ticksToDowngrade` reaches 3000, and the
downgrade itself.

## Core touchpoints

- `mods/controller/index.ts`: declares `notifications` as a dependency.
- `processor.ts`: `upsertNotification` on level-up (in the `upgradeController`
  intent), on downgrade, and on the warning tick — a single
  `ticksToDowngrade === 3000` branch alongside the existing downgrade-timer
  transitions. Owned rooms process every tick, so the equality lands on the
  boundary with no persisted state or extra wake.

## Validation

- xxscreeps test runner: 383/383 pass.
- `Controller > lifecycle notifications`:
  - `upgradeController level-up delivers a notification`
  - `pre-downgrade warning is delivered exactly once when ticksToDowngrade enters the 3000-tick window`
  - `downgrade delivers a notification with the new level`
  - `pre-downgrade warning re-arms after a downgrade tick`
- `pnpm run build` clean; eslint on changed files adds no new warnings.


laverdet:
I don't think you need a flag for this. Not even Screeps does this:
https://github.com/screeps/engine/blob/80977824199a596d174d392fd0cf8c458c21fcbd/src/processor/intents/controllers/tick.js#L45-L47

laverdet (changes requested):
^

misterwise:
> I don't think you need a flag for this. Not even Screeps does this: https://github.com/screeps/engine/blob/80977824199a596d174d392fd0cf8c458c21fcbd/src/processor/intents/controllers/tick.js#L45-L47

Good point, didn't need the flag, and didn't need the wakeAt, got a bit too carried away with the concern of sleeping rooms in xxscreeps vs vanilla, which doesn't apply to owned rooms.

PR #214, opened 2026-05-10
Address PR #165 follow-up review

misterwise:
Addresses the three follow-up review comments on #165.

## Summary

- Per-user notifications index becomes a sorted set scored by each row's group due time, so a short group fires alone without dragging an unrelated long group along with it. The user's drain pointer in `notifications/dueUsers` is moved via `ZADD LT` on insert and to the next pending row's deadline after each drain.
- Drop the `time` parameter from `ShardTickProcessor`. Both call sites advance `shard.time = time` before invoking `runShardTickProcessors(shard)`, and `everyNTicks` reads `shard.time` directly.
- Remove `mods/notifications/hooks.ts`. Replaced with a plain transports array plus a registration helper that returns a `Disposable` so tests can scope a transport via `using`.

## Local keyval bugfixes

Both surfaced while this PR became the first caller of these code paths.

`zadd up: 'lt' | 'gt'` in the local keyval impl was passing `Math.min` / `Math.max` as the `set.insert` accumulator, which then ran `accumulator(0, score)` for new members and stored `Math.min(0, score) = 0` for any positive score. Per Redis semantics, GT/LT only constrain *updates* to existing members; new members are added at the given score regardless. `scheduleUserDrain` is the only caller of `zadd up:` in the codebase — the option itself was added in the commit immediately preceding this branch, so this PR is its first consumer. Without the fix, the redesign still delivers correctly (the row-level filter in `getDueNotifications` saves it) but every user with anything pending is drained every cadence tick rather than only at their deadline, defeating the point of `dueUsersKey` being a scheduled pointer.

`zrangeWithScores` in the local keyval impl ignored `min`/`max` in the index (default) branch — it mapped *every* member with its score regardless of the requested range. `nextPendingDueAt` happened to do the right thing because `head[0]` is the lowest-scored, but the function was broken on its own terms, and the redis variant correctly delegates to `ZRANGE … WITHSCORES`. `nextPendingDueAt` is the only caller of `zrangeWithScores` in the codebase. Fixed by delegating to `zrange` (which already handles index conversion and `limit`) and mapping scores onto the result.

## Validation

- `pnpm run build`
- `npx xxscreeps test` — 291 tests pass
- `pnpm exec eslint` on changed files — zero new warnings


laverdet:
```
diff --git a/packages/xxscreeps/engine/db/storage/local/sorted-set.ts b/packages/xxscreeps/engine/db/storage/local/sorted-set.ts
index cc20912..ad5e273 100644
--- a/packages/xxscreeps/engine/db/storage/local/sorted-set.ts
+++ b/packages/xxscreeps/engine/db/storage/local/sorted-set.ts
@@ -69,7 +69,7 @@ export class SortedSet {
             const currentScore = this.#scores.get(member);
             if (currentScore === undefined) {
                 this.#members.push(member);
-                this.#scores.set(member, accumulator(0, score));
+                this.#scores.set(member, score);
                 ++count;
             } else {
                 this.#scores.set(member, accumulator(currentScore, score));
```

isn't this a better fix?

misterwise:
> ```
> diff --git a/packages/xxscreeps/engine/db/storage/local/sorted-set.ts b/packages/xxscreeps/engine/db/storage/local/sorted-set.ts
> index cc20912..ad5e273 100644
> --- a/packages/xxscreeps/engine/db/storage/local/sorted-set.ts
> +++ b/packages/xxscreeps/engine/db/storage/local/sorted-set.ts
> @@ -69,7 +69,7 @@ export class SortedSet {
>              const currentScore = this.#scores.get(member);
>              if (currentScore === undefined) {
>                  this.#members.push(member);
> -                this.#scores.set(member, accumulator(0, score));
> +                this.#scores.set(member, score);
>                  ++count;
>              } else {
>                  this.#scores.set(member, accumulator(currentScore, score));
> ```
> 
> isn't this a better fix?

Yeah, accumulator in the empty case doesn't make sense, that's right.

PR #215, opened 2026-05-11
Add RoomObject.effects (first slice) via StructureInvaderCore

misterwise:
## Summary

Adds the public `RoomObject.effects` shape and a `hasEffect(object, effectType)` helper, with `StructureInvaderCore` as the first producer.

The core's `effects` array is synthesized from its existing `#deployTime` — no persisted effect state. The getter returns `undefined` when no effects are active. The core's `#applyDamage` override reads `hasEffect(this, EFFECT_INVULNERABILITY)` rather than `#deployTime` directly, so future stored-effect producers (controller invulnerability, stronghold collapse) compose without re-editing the override.

Intent gates for `attack` / `rangedAttack` / `dismantle` reject `EFFECT_INVULNERABILITY` inline; `rangedMassAttack` has no per-target gate, so the `#applyDamage` override is the backstop.

## Validation

`npx xxscreeps test` (289 passed); pre-push `pnpm build` clean.

screeps-ok regression check: `tests/15-structure-common/15.5-effects-substrate.test.ts`, `tests/15-structure-common/15.6-effects-hosts.test.ts`, `tests/00-adapter-contract/{setup,inspection}.test.ts` — no regressions. (`InvaderCore.effects` itself isn't exercised — the adapter's `invaderCore` capability is still false.)


packages/xxscreeps/game/test.ts
```diff
@@ -20,3 +21,16 @@ describe('RoomPosition', () => {
 		}
 	});
 });
+
+describe('RoomObject', () => {
+	// Substrate must not live on the base prototype — would leak 'effects' onto every RoomObject.
+	test('effects is not a property on the base RoomObject prototype chain', () => {
+		for (let proto: object | null = RoomObject.prototype; proto !== null; proto = Reflect.getPrototypeOf(proto)) {
+			assert.strictEqual(
```
laverdet: First, this test could just be `assert.equal('effects' in RoomObject.prototype, false)`

But more importantly I don't want to get into a habit of codifying this strict of a conformance to exact object shape. I don't think many users are reaching for `if ('effects' in creep)` instead of `if (creep.effects)`.

If the resulting implementation is more ergonomic for us then that's fine. But divergent shapes create [hidden classes](https://v8.dev/docs/hidden-classes) in v8 which in general you want to avoid. The performance impact is usually pretty minimal but I want to stress that we should not be bending over backwards to adhere to Screeps' exact undefined-in shapes.

packages/xxscreeps/mods/combat/creep.ts
```diff
@@ -126,6 +127,7 @@ export function checkAttack(creep: Creep, target: AttackTarget) {
 		() => checkSafeMode(creep.room, C.ERR_NO_BODYPART),
 		() => checkTarget(target, Creep, Structure),
 		() => checkDestructible(target),
+		() => hasEffect(target, C.EFFECT_INVULNERABILITY) ? C.ERR_INVALID_TARGET : C.OK,
```
laverdet: I think this should just be a getter defined on `InvaderCore`, like `get '#invulnerable'() { return whatever }`. Then you would do the same `interface RoomObject` merging thing to add it to TypeScript. This will lead to a cleaner implementation of `PWR_SHIELD` and `PWR_FORTIFY` in the future without a tangled web of dependencies on mod/power

packages/xxscreeps/game/object.ts
```diff
@@ -24,6 +24,13 @@ const shape = struct({
 	'#posId': union({ pos: 'int32' }),
 });
 
+export interface RoomObjectEffect {
+	effect: number;
```
laverdet: `effect` is also optional, in the power case.

packages/xxscreeps/mods/invader/invader-core.ts
```diff
@@ -0,0 +1,64 @@
...
+		return ticksRemaining === undefined ? undefined : [ { effect: C.EFFECT_INVULNERABILITY, ticksRemaining } ];
+	}
+
+	// TODO: stronghold spawning
+	// eslint-disable-next-line @typescript-eslint/class-literal-property-style
+	@enumerable get spawning(): null { return null; }
+
+	@enumerable get ticksToDeploy(): number | undefined {
+		const deployTime = this['#deployTime'];
+		return deployTime > Game.time ? deployTime - Game.time : undefined;
```
laverdet: Update with the new helpers

packages/xxscreeps/mods/invader/invader-core.ts
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	override get hitsMax(): number | undefined {
```
laverdet: Return value should be number

laverdet (changes requested):
Main issue is the proposed `#invulnerable` getter instead of checking an effects array.

misterwise:
Addressed in 7e810675:

- Replaced the `hasEffect` array scan in `checkAttack`/`checkRangedAttack`/`checkDismantle` and the core's `#applyDamage` with a `#invulnerable` dispatch getter — base `false` on `RoomObject` (alongside `#providesVision`), `StructureInvaderCore` overrides. Dropped `hasEffect`; `effects` stays a derived player-facing view.

  I went with a base default rather than the type-only `interface RoomObject` merge so it lines up with the existing `#providesVision` / `#hasIntent` capability getters and consumers get a plain `boolean` on every object rather than `boolean | undefined`. Happy to switch to the merge (InvaderCore-only) if you'd prefer that.

- Simplified the prototype test to a single leak guard: `assert.equal('effects' in RoomObject.prototype, false)`.
- `RoomObjectEffect.effect` optional.
- `ticksToDeploy` via `optionalExpiryTime`.
- `hitsMax` returns `number`.

packages/xxscreeps/game/test.ts
misterwise: Ok, I've made a stricter note for myself of this, I agree with your assessment of this and the check didn't match the intent of hte comment.

packages/xxscreeps/mods/combat/creep.ts
misterwise: I put it on the base, as that looks cleaner to me but if you think the interface merge is a better approach or more consistent, I will change that.

laverdet:
I think for effects they will all be implemented with semi-private properties. All Operator does is flip a flag on specific structures. My normal guideline is that the processor should check the same game properties that the user checks, but I think `object.effects` will be mostly a user-facing object that we never actually inspect.

Issue #217, opened 2026-05-13
duplicate console messages when connecting with two clients


bastianh:
looks like the websocket subscriptions are not correctly scoped to the connection but to the user? when connecting with two clients I get messages duplicated on both clients.

will investigate further and update the ticket or maybe I find a fix .)

PR #221, opened 2026-05-15
fix(backend): deduplicate websocket subscriptions and prevent pubsub …

bastianh:
…N^2 duplication

- Implement canonical IDs for websocket subscription endpoints to prevent redundant data streams on a single socket.
- Add a processing lock to the room update timer to prevent overlapping async loads.
- Refactor LocalPubSubProviderClient to deduplicate upstream subscriptions, fixing the N^2 message duplication bug in development environments.
- Normalise room and map subscription IDs to ensure consistent deduplication regardless of shard prefix usage.

fix for laverdet/xxscreeps#217


laverdet:
Could you let me know if d6c402baf18410e84655caca134e10292770c509 resolves the issue? If so, is the remainder of the PR still relevant? Thank you!

bastianh:
yes .. looks like the issue is gone in current main

PR #224, opened 2026-05-18
Fix local zUnionStore to apply WEIGHTS to single-set members

misterwise:
## Summary

`LocalKeyValResponder.zUnionStore` drops the WEIGHTS factor on members
present in only one input set: `SortedSet.insert` stores new members at
their literal score and only invokes the accumulator on existing ones,
but the multiplication lives inside the accumulator.

`engine/processor/model.ts` uses
`zUnionStore(processSet, [activeRoomsKey, tmpKey], { weights: [1, 0] })`
to merge waking rooms into the processing queue at score `0`. Under the
local provider, waking rooms instead landed at their raw wake-time score
(~1.5M), never got consumed, and tripped `intentAbandonTimeout` — 5s
tick stalls and `"Abandoning intents in rooms"` warnings.

Pre-multiply scores by weight before `insert`.

## Validation

- `pnpm run build` clean
- `npx xxscreeps test` — 343/343 passing, including the new
  `LocalKeyValResponder` regression test
- Confirmed the new test fails on `main` and passes with the fix

### DIFF
--- a/.changeset/local-zunionstore-weights.md
+++ b/.changeset/local-zunionstore-weights.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Fix local `zUnionStore` to apply WEIGHTS to members present in only one input set.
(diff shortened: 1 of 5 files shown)


laverdet:
Looks like we can trace it back to my comment here :)
https://github.com/laverdet/xxscreeps/pull/214#issuecomment-4418524480

PR #225, opened 2026-05-18
engine: load shard tick processors in main service

misterwise:
## Summary

`registerShardTickProcessor` registrations have silently never fired in production. The helper (introduced in #165) pushes into a module-level array; the dispatcher (`runShardTickProcessors`) runs in main's `tickFinished` arm. But main calls `importMods` for nothing — registrations land in worker/backend (which import the `processor` slot) while main's array stays empty. `notifications/processor.ts` was the only registrant in the tree, so its drain has been dead since #165 merged. `test/simulate.ts` masks the bug by loading every slot in a single process.

Adds a narrow `shardTick` slot loaded by main, migrates the notifications registration into `notifications/shardTick.ts`, and loads the same slot in simulate so the harness stays aligned with production wiring.

Reusing the existing `processor` slot would also fix the bug, but that slot loads worker-side surface (intent processors, room/object-tick processors, harvest hooks) that main has no business holding registrations for. A separate slot keeps main's loaded set narrow and makes scope visible at the loader level: a mod's `processor.ts` runs in the worker; its `shardTick.ts` runs in main.

## Validation

- `pnpm run build` clean.
- `pnpm exec eslint` on changed files — zero new warnings.
- Live engine run drained a long-accumulated `Game.notify` backlog on first tick after start; pre-fix the queue had been silently growing since #165.


laverdet:
Renamed it to 'main'

PR #226, opened 2026-05-18
redis: require Redis >= 8.2 and probe version at connect

misterwise:
## Summary

- `@xxscreeps/redis` started using `SET ... IFEQ` in the mutex (`packages/redis/keyval.ts:115,160`), which is a Redis 8.2 feature. On older servers the first mutex lease renewal returns `ERR syntax error` and crashes the launcher at a non-deterministic tick (reproduced on 7.4.9: tick 264).
- Add a connect-time `INFO server` probe in `acquireRedisClient` so older servers fail fast at startup with `@xxscreeps/redis requires Redis >= 8.2 (found <x.y>)` instead of a cryptic decoder error hours later.
- Document the requirement in `packages/redis/README.md` and add an informational `engines.redis: ">=8.2"` to the package manifest.

## Validation

- `pnpm run build` — clean.
- `npx eslint --max-warnings 0 packages/redis/client.ts` — clean.
- Spun up `redis:7-alpine` (7.4.9) on a side port; `acquireRedisClient` threw `@xxscreeps/redis requires Redis >= 8.2 (found 7.4)` at connect, as expected.
- Spun up `redis:8-alpine` (8.6.3); engine launched and ran 2374 ticks past the prior 264 crash point, then shut down cleanly on SIGINT.

### DIFF
--- a/.changeset/redis-min-version.md
+++ b/.changeset/redis-min-version.md
@@ -1,0 +1,5 @@
+---
+"@xxscreeps/redis": patch
+---
+
+Require Redis >= 8.2 (mutex uses `SET ... IFEQ`); probe version at connect.
--- a/packages/redis/README.md
+++ b/packages/redis/README.md
@@ -1,6 +1,12 @@
 # @xxscreeps/redis
 
-Sample configuration:
+## Requirements
+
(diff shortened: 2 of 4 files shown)


laverdet:
I don't think "engines": { "redis": ... } is a thing that anyone does. Thanks for the check though!

Issue #168, opened 2026-04-30
Define the xxscreeps privileged CLI product surface


misterwise:
## Goal

Define the intended privileged CLI surface for xxscreeps: interactive operation, one-shot scripting, structured domain commands, offline DB maintenance, import/export, backups, and future extensibility.

The CLI should be a native xxscreeps operator surface, not command-for-command parity with prior Screeps private-server CLIs. It should provide local privileged access, shard-aware context, thin clients, structured errors, and reusable command/runtime infrastructure.

## Non-Goals

- Changing `xxscreeps` default startup into a REPL.
- Command-for-command parity with another private-server CLI.
- HTTP admin endpoints, remote auth, or remote administration.
- Node built-ins, imports, or environment access inside REPL/eval scripts.
- Audit logging design.
- Initial support for live fixture import, live terrain invalidation, or object mutation convenience wrappers.

## User-Facing Surface

- `xxscreeps`: existing server/default behavior.
- `xxscreeps cli`: explicit interactive REPL.
- `xxscreeps eval`: one-shot scripted evaluation.
- Top-level privileged domains, without an `admin` prefix:
  `users`, `bots`, `rooms`, `objects`, `system`, `config`, `storage`, `map`, `profiler`, `import`, `export`, `backup`, and future `intents`.
- Shared options such as `--socket`, `--shard`, and explicit offline mode.
- Help, examples, domain help, REPL help, sandbox `help()`, and shell completion.

## Domain Scope

- `users`: create/show/list/remove, badge defaults/overrides, ownership and reservation inspection.
- `bots`: add/reload/remove, load code, show/list status, badge defaults/overrides.
- `rooms`: inspect and mutate through shared invariant-preserving helpers; expose `rooms.edit` for lower-level caller-responsible mutation.
- `objects`: start with `objects show <id>`; mutation commands can later wrap room editor helpers.
- `system`: pause/resume, tick duration, server message, status/health, reset data, and live-only `system step [count]`.
- `config`: distinguish effective runtime config from persisted `.screepsrc.yaml`.
- `storage`: explicit raw db/shard access as a caller-responsible maintenance escape hatch.
- `map`: generate/open/close/remove rooms plus terrain import/export.
- `profiler`: timing and performance data separate from `system status`.
- `import`/`export`/`backup`: human-editable scenario fixtures and exact internal/versioned snapshots.
- `intents`: reserved for future privileged intent inspection or manipulation.

[...]

PR #227, opened 2026-05-19
Add live REPL to xxscreeps cli

misterwise:
## Summary

Adds a live REPL to `xxscreeps cli`. With a local provider + launcher up, the REPL forwards each input over the launcher's RPC socket, sharing the launcher's in-process `db`/`shard` and persisting `let`/`const`/`var` across turns within a connection. With a networked provider (`redis:`) the REPL runs in the CLI's own process — storage is the rendezvous, so it still reaches the live engine state the launcher is writing. Local-without-launcher falls back to a host-realm REPL with no engine state pre-bound. `xxscreeps eval` stays host-realm only.

Slice 2 of #168.

## Design

**Auto-detect, no flags.** `xxscreeps cli` inspects `config.database.data`. Local schemes (`file:`, `local:`) probe the launcher RPC socket; networked schemes (`redis:`) skip the probe outright — there's no in-process state to share, and the live data is already reachable through storage. The banner names the chosen mode on stderr; the prompt is `xxscreeps live>` vs `xxscreeps>`.

**Launcher RPC = Unix socket / Windows named pipe at `screeps/cli.sock`** next to `.screepsrc.yaml`. The name says what it is — a server location with a call/return protocol shape. Unix permissions `0o600` socket / `0o700` parent; same-UID is the trust boundary, not the sandbox.

**Per-connection `vm.Context`.** Each connection allocates one context that lives for the connection's lifetime, so `var x = 41` from one turn is visible to `x + 1` in the next. Requests within a connection serialize through a promise chain so the context never observes a half-applied declaration. Parallel sessions get isolated contexts.

**Console proxy via AsyncLocalStorage.** `console` in the per-connection context is a stable proxy; each request runs inside `sinkStore.run(sink, …)`. A late `setTimeout` callback inherits the async context of the request that scheduled it, so its log lands in that request's drained sink rather than bleeding into a later request's envelope.

**Top-level declarations hoisted in the async wrap.** `let x = await …` previously vanished after the IIFE returned. Acorn parses the source, hoists top-level `let`/`const`/`var` (including destructuring) as `var x;` at script scope, and rewrites the declaration to an assignment in the arrow body. Same shape Node's REPL uses for `--experimental-repl-await`. Class/function declarations don't persist; bind via `let C = class { … }`.

**Drain on shutdown.** Launcher RPC cleanup is `disposable.defer`'d after the save defer; LIFO disposal means RPC drain runs first, so an in-flight eval can't mutate state between flush and shutdown.

**`xxscreeps eval` stays host-realm.** Never RPC-routed.

**Opt-out.** `xxscreeps start --no-launcher-rpc` runs the launcher without the socket; CLI falls back to direct mode.

## Out of scope

- Per-command transport routing (`Manifest.commands` + `requiresLauncher`) — slice 3; `detectMode` carries a TODO marking this detection as REPL-eval-transport only.
- Provider capability flag instead of the hardcoded scheme list — deferred per #168.
- Multi-shard handshake — launcher RPC serves the launcher's default shard.

## Validation

- `pnpm run build` clean.
- `npx eslint` on `packages/xxscreeps/cli/` — zero warnings introduced.
- `npx xxscreeps test` — 387/387 pass; new describes: `cli` (15), `launcher RPC` (30), `cli live` (8) all green.

### DIFF
--- a/.changeset/cli-live-repl.md
+++ b/.changeset/cli-live-repl.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": minor
+---
+
+Add `xxscreeps cli` launcher-RPC live REPL with provider-aware auto-detection; drain async eval tails on EOF.
--- a/README.md
+++ b/README.md
@@ -123,11 +123,11 @@ If you want to use your bot script, use `npx xxscreeps import --overwrite-code .
 
 ## Operator CLI
 
-`xxscreeps cli` opens a JavaScript REPL in the CLI process's host realm, and
-`xxscreeps eval` runs one-shot scripts with `-e`, `--file`, or `--stdin`, plus
-an optional `--json` envelope mode for tooling. See
-[packages/xxscreeps/cli/README.md](packages/xxscreeps/cli/README.md) for flags
-and the JSON envelope schema.
+`xxscreeps cli` opens a JavaScript REPL that auto-detects a running launcher
+and forwards over its RPC socket when available, falling back to a host-realm
+REPL otherwise. `xxscreeps eval` runs one-shot scripts with `-e`, `--file`, or
+`--stdin`. See [packages/xxscreeps/cli/README.md](packages/xxscreeps/cli/README.md)
+for flags and launcher-mode details.
 
 ## Docker
 
--- a/packages/xxscreeps/cli/README.md
+++ b/packages/xxscreeps/cli/README.md
@@ -1,24 +1,34 @@
 # xxscreeps CLI
 
-Operator-facing JavaScript REPL and one-shot evaluator. Both run inside the CLI process's host realm
-— operator typing has the same access the launcher does. There is no `vm.Context`, no curated
-globals list.
+Operator-facing JavaScript REPL and one-shot evaluator.
+
+- `xxscreeps cli` evaluates each line against the launcher's `db`/`shard` when the launcher's RPC
+  socket is reachable (**launcher mode**), or in the CLI's own process otherwise (**direct mode**).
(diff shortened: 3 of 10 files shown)


laverdet:
This is all questionably over-engineered.

I really don't want to do the acorn thing. Users can just do `a = await expr` to dump it out to globals. Building a generalized and rich repl library is not a goal of this project. If you want to try and get the node guys to make a public API for the ridiculous autocompletion and hoisting logic [here](https://github.com/nodejs/node/blob/main/lib/internal/repl/completion.js) that would be a fine place for that to live.

Also the intent of the work in 6c5d8ca9d36c019d3530f2ed6315d946fe66b59f is that you would not need a socket for this thing. Why does it matter whether the commands run in the main process or in the cli? They are accessing the same underlying data.

misterwise:
1. Ok. I had seen it as a simplification of access, and reduction of complexity but not seen the opportunity it offered in terms fo moving out of launcher entirely and persisted with this idea that there should be a "live" version of the CLI with access to game objects, but I guess you're right, it's the underlying data access that matters. I think I'm going to shift instead to just implementing a few aspects of functionality that would be useful to me in the simplest way possible and see if that fits better.  Being able to modify users (probably a script) and bot management (also script most likely), and then some quality of life for debugging like pause/unpause, and step(numticks). I might also look at being able to poke rooms and modify state, I have found that helpful, but it might also be possible just with scripts as well. Your clarity that a rich repl is outside the scope is helpful. I might add overengineering as a specialty after this.

PR #231, opened 2026-05-21
engine: pass current time to inter-room finalize phase

misterwise:
## Summary
- `worker.ts:125` constructed the inter-room finalize `RoomProcessor` with `time + 1`, so `finalize` saved the target room at `shard.time + 2` while `shard.time` was still T. Live engine crashed with `Invalid time: T+2 [current: T]` whenever an inter-room intent (observer, terminal, nuke, room-crossing creep) landed on a room with no other processing this tick.
- Extracted the inter-room finalize loop into `engine/processor/finalize-extra.ts`; both `worker.ts` and `test/simulate.ts` now call it so production and the test harness exercise the same code path.

## Validation
- `pnpm run build` clean.
- `npx xxscreeps test`: 365 pass (was 364; +1 regression test).
- New `Inter-room finalize > intent into non-processed room saves at next tick` test reproduces the live `Invalid time: T+2 [current: T]` stack against the pre-fix helper, and passes against the fix.
- `npx eslint` on the changed files: zero new warnings.

### DIFF
--- a/.changeset/inter-room-finalize-time.md
+++ b/.changeset/inter-room-finalize-time.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Fix processor crash when finalizing inter-room intent target rooms.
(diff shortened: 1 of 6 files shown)


laverdet:
afc9b1cea340f329b38c612c6234623f1b27ce71

PR #232, opened 2026-05-21
Fix nuker launch timing, impact visibility, and safe-mode withdraw

misterwise:
## Summary
- Nuker launch: `#cooldownTime` and the dispatched `landTime` now write `Game.time + COOLDOWN/LAND_TIME - 1`, so first observable values match vanilla (`NUKER_COOLDOWN - 1`, `NUKE_LAND_TIME - 1`).
- Nuke lifecycle: impact applies when `Game.time === #landTime`, the nuke remains visible with `timeToLand === 0` on the impact tick, and `#removeObject` runs the following tick.
- `Nuke.timeToLand` now wraps `cooldownTime` instead of `optionalExpiryTime`. `optionalExpiryTime` throws on overdue, which would fire on the post-impact tick where `#landTime < Game.time` by design.
- `checkWithdraw`: `checkSafeMode` runs before the nuker store's `#doesAllowWithdraw()` rejection so a hostile creep in a safe-mode room targeting a nuker gets `ERR_NOT_OWNER` instead of `ERR_INVALID_TARGET`.

## Note on the time helpers
`cooldownTime` has the right *semantics* for `Nuke.timeToLand` (clamps at 0, doesn't throw, no 0-sentinel), but the name is a misnomer — a landing nuke isn't recovering from a cooldown. The four helpers in `game/object.ts` tangle two axes: whether `0` is a meaningful "unset" sentinel (`required` vs `optional`), and what happens on overdue (clamp / undefined / throw). Nuke wants "no 0-sentinel, clamp on overdue", which only `cooldownTime` provides — but under a cooldown-themed name. Probably worth a follow-up to either rename to a neutral `expiryTime` or split the throw axis from the cooldown-vs-expiry distinction.

## Validation
- `pnpm run build`.
- `pnpm exec eslint --max-warnings 0` on every changed file — zero new warnings (pre-existing `transfer` warning unchanged).
- `npx xxscreeps test` — 365 passed, 0 failed. New assertions: `timeToLand === 0` at impact tick + nuke removed next tick; hostile withdraw of a safe-mode nuker returns `ERR_NOT_OWNER`.
- `screeps-ok` parity suite — three previously-known xxscreeps parity gaps now pass:
  - `NUKE-LAUNCH-012` (cooldown first visible = `NUKER_COOLDOWN - 1`)
  - `NUKE-IMPACT-013` (nuke visible with `timeToLand === 0` on impact tick, removed next tick)
  - `WITHDRAW-017:safemodeNotOwnerBeforeInvalidNuker`


laverdet:
`cooldownTime` plasters over the safety we're trying to accomplish with these functions. Instead I fixed it with a +1 and corresponding -1 to extend the domain of this function. The way nukes sit around for a tick is kind of weird anyway.

PR #234, opened 2026-05-23
construction: reject site stacking unless road or rampart

misterwise:
## Summary

`checkCreateConstructionSite` gated the on-tile pair check on the *placed* type's `obstacle` flag, missing two cases: a container (non-obstacle) sitting on a tile never blocked a non-road/non-rampart placement, and a non-obstacle placed type slipped past the check entirely against any existing structure. Both directions returned `OK`.

The new check encodes the stacking rule symmetrically: a buildable pair on the same tile is rejected unless either side is road or rampart. `object.structureType in C.CONSTRUCTION_COST` is the same buildable predicate already used in `construction/creep.ts:81` and `structure/structure.ts:291`.

## Validation

- `pnpm run build`
- `npx eslint packages/xxscreeps/mods/construction/{room,test}.ts` → 12 warnings (was 13 on `main`; all pre-existing in unrelated overload code)
- `npx xxscreeps test Construction` → 34/34
- `npx xxscreeps test` → 375/375
- screeps-ok `CONSTRUCTION-SITE-017` → 12/12 (5 previously-failing rejection rows: `container-blocks-{tower,spawn,extension}`, `{tower,spawn}-blocks-container`)


laverdet:
Please don't special case structure types in the construction mod.

laverdet (changes requested):
^

misterwise:
> Please don't special case structure types in the construction mod.

Yes, this is the second time I've done this. I moved it into a property on the factory, I think that keeps it cleaner?

laverdet:
I made observable modifications to this line: `if (existing && !factory.stackable && !existing.stackable) {`. As I understand it this would default non-constructible structures (portal, invader core) to stackable.

PR #235, opened 2026-05-23
engine: fix processor hang on SIGINT in idle gap between ticks

misterwise:
## Summary

- Ctrl-C against the engine occasionally failed to shut down cleanly: most of the time it exited fine, but every few attempts it would hang at `Shutting down...` (or, depending on where the responder traffic landed, error out before exiting) instead of reaching `💾 Engine shut down successfully.`. The race lived in the idle gap between ticks — frequent enough to notice, rare enough to be annoying to reproduce.
- Root cause: the processor's SIGINT broadcast handler in `packages/xxscreeps/engine/service/processor.ts` set `halted = true` and called `halt?.()` (the breaker for the main message loop) whenever `processing` was false. When SIGINT landed after the previous `finalize` cleared `processing` and before the next `process` arrived, the loop broke and the pubsub subscription tore down. Main's next-tick `process` publish (and the recovery publish from `abandonIntentsForTick`) then landed on nothing, and main hung waiting for a `tickFinished` no one could produce.
- Main forwards `shutdown` on the processor channel after its own `while (true)` exits — that's the authoritative exit signal. Drop the `processing` flag and the `halt?.()` call from the SIGINT handler entirely; let main own loop teardown. Keep `halted = true` so the existing post-`finalize` `if (halted) break loop` still terminates promptly when SIGINT lands mid-tick.

## Validation

- `pnpm run build` — clean (pre-push hook).
- On `origin/main`: Ctrl-C against the launcher reliably fails to reach `💾 Engine shut down successfully.` within a handful of attempts.
- On this branch: Ctrl-C at varying points in the tick cycle consistently reaches `💾 Engine shut down successfully.` and exits cleanly.

### DIFF
--- a/.changeset/processor-shutdown-race.md
+++ b/.changeset/processor-shutdown-race.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Fix engine hang on SIGINT during the idle gap between ticks
--- a/packages/xxscreeps/engine/service/processor.ts
+++ b/packages/xxscreeps/engine/service/processor.ts
@@ -1,5 +1,4 @@
 import type { ProcessorRequest } from 'xxscreeps/engine/processor/worker.js';
-import type { Effect } from 'xxscreeps/utility/types.js';
 import config from 'xxscreeps/config/index.js';
 import { consumeSet, consumeSortedSet, consumeSortedSetMembers } from 'xxscreeps/engine/db/async.js';
 import { Database, Shard } from 'xxscreeps/engine/db/index.js';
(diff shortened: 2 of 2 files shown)


laverdet:
In a larger distribution you'd run xxscreeps like this: `npx xxscreeps start --no-processor --no-backend --no-runner`. Then you would run `npx xxscreeps processor`, `npx xxscreeps backend` and `npx xxscreeps backend` on separate machines or clusters of machines. Workers for each role can come and go. So each service still needs to respond to its own signals.

You might be able to make this behavior conditional on isEntry

misterwise:
Good catch — the original change broke standalone deployments. Pushed [18310645](https://github.com/misterwise/xxscreeps/commit/18310645) to gate the new behavior on `isEntry`:

- `isEntry === false` (combined launcher, single- or multi-threaded): main owns shutdown via the processor channel, same as before — this is the path that fixes the original idle-gap race.
- `isEntry === true` (`npx xxscreeps processor` on its own host): self-halt restored — SIGINT calls `halt?.()` between ticks so the message loop breaks and the process exits.

Verified locally:

- Standalone: `npx xxscreeps start --no-processor` + `npx xxscreeps processor`, SIGINT'd the processor → exit 0 in 49ms via loop-break cleanup.
- Combined: 10 race-loop iterations with and without the new commit are at parity (both finish in ~5s/run, no SIGKILL hangs). The pre-existing `ClientClosedError` in `Shard.loadRoomBlob` during shutdown surfaces at the same rate in both — separate race, unrelated to this PR.

PR #236, opened 2026-05-23
map: return 'closed' from getRoomStatus for off-world and inaccessible rooms

misterwise:
## Summary

- `Game.map.getRoomStatus()` diverged from vanilla in two places
  (`packages/xxscreeps/game/map.ts`): off-world room names returned
  `undefined` (vanilla returns `{status:'closed', timestamp:null}`),
  and inaccessible rooms (terrain present, not in the active-rooms
  set) returned a made-up `'out of borders'` status (vanilla returns
  `'closed'`).
- Add a vanilla-style `^[WE]\d+[NS]\d+$` format check (invalid names
  still return `undefined`), and collapse the off-world and
  inaccessible cases to `{status:'closed', timestamp:null}`. Drop the
  `'out of borders'` literal from the `RoomStatus` type union.
- Six internal callers were using `if (map.getRoomStatus(x))` as a
  "has terrain" gate (backend room subscription, map-stats, terrain
  thumbnails, last-viewed-room start, inter-room intent routing,
  `observer.observeRoom`). After the parity fix off-world names would
  have flipped truthy and routed those callers to load nonexistent
  rooms. Introduce `GameMap.hasRoom(name)` (wraps the private terrain
  map) and point all six call sites at it so their existing gate
  semantics are preserved.
- `observer.observeRoom` retains the existing xxscreeps-specific
  defensive rejection of off-world targets (locked in by the
  `observer_nonexistent_room` unit test); only
  `Game.map.getRoomStatus()`'s return value changes shape.

## Validation

- `pnpm run build` — clean.
- `npx eslint` on the seven changed files — zero new warnings.
- `npx xxscreeps test` — 373/373 passed, including the
  `Observer observer_nonexistent_room` regression-defense test that
  asserts `observeRoom` to an off-world target returns
  `ERR_INVALID_ARGS`.
- screeps-ok suite against this branch:
  - `MAP-ROOM-004:normal` — `getRoomStatus` returns
    `{status:'normal', timestamp:null}` for an in-world room with no
    admin status set. Passes on xxscreeps and vanilla.
  - `MAP-ROOM-004:offWorld` — `getRoomStatus` returns
    `{status:'closed', timestamp:null}` for a valid-format room name
    that does not exist on the world. Was returning `undefined` on
    xxscreeps before this PR; now passes on xxscreeps and vanilla.
  - `MAP-ROOM-004:adminClosed` — `getRoomStatus` returns
    `{status:'closed', timestamp:<number>}` for an admin-closed
    in-world room. Passes on vanilla; skipped on xxscreeps via
    `shard.requires('roomStatus')` because xxscreeps has no admin
    room-status storage yet.

### DIFF
--- a/.changeset/get-room-status-vanilla-parity.md
+++ b/.changeset/get-room-status-vanilla-parity.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Return `{status:'closed', timestamp:null}` from `Game.map.getRoomStatus()` for off-world or inaccessible rooms.
--- a/packages/xxscreeps/backend/endpoints/assets/terrain.ts
+++ b/packages/xxscreeps/backend/endpoints/assets/terrain.ts
@@ -204,7 +204,7 @@ function register(paths: string[], fn: (shard: Shard, map: GameMap, room: string
 
 // Full thumbnail
 register([ '/assets/map/:room.png', '/assets/map/:shard/:room.png' ], async (shard, map, roomName) => {
-	if (map.getRoomStatus(roomName)) {
+	if (map.hasRoom(roomName)) {
 		return generate(map, [ [ await shard.loadRoom(roomName) ] ], 3);
 	} else {
 		return null;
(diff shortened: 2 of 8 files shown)


laverdet:
'out of borders' looks to be an internal database value, which as far as I can tell, Screeps never actually sets. It was added by 7517ba7d152f9c76d5f0734b7b13164ad984fe45 / #151.

Anyway the problems here:
- Please don't copy/paste regular expressions from the Screeps engine. We have a utility for this already.
- `Game.map` is a public interface, please don't add extraneous functions to it.
- `Game.map.getRoomLinearDistance` returns NaN for room names which don't follow the correct format (this is Screeps canon). The fix for observers is to invert the conditional; comparisons to NaN always return false.

I patched it here 2cf66aafa1c6649994ff12725c647e1f5c4ba45d

PR #237, opened 2026-05-23
engine: snapshot processor Tick/PreTick loops over `#objects`

misterwise:
## The crash

A running engine died on tick 298196 with:

```
Tick 298196 ran in 30ms; avg: 30.24ms
Error: TypeError: Cannot read properties of undefined (reading 'Symbol(tick)')
    at <anonymous> (/.../packages/xxscreeps/engine/processor/room.ts:173:12)
    at runWithState (/.../packages/xxscreeps/game/index.ts:36:10)
    at RoomProcessor.process (/.../packages/xxscreeps/engine/processor/room.ts:98:3)
    at <anonymous> (/.../packages/xxscreeps/engine/processor/worker.ts:112:18)
```

`room.ts:173` is `object[Tick]?.(object, this)` inside the post-intent
Tick loop. `object` is `undefined` — `objects[ii]` is missing even
though `ii < length`, where `length` was captured once before the loop.

## Root cause

The Tick and PreTick loops captured `objects.length` once and iterated
by index, on the assumption that `#objects` could only grow during the
loop (via `now=true` inserts). It can also shrink:
`StructureController`'s Tick processor calls
`updateRoomStatus → room['#flushObjects']` on downgrade, and a queued
removal from any earlier Tick in the same room — decayed `Resource`,
post-impact `Nuke`, decayed `Rampart`, etc. — then splices `#objects`
mid-iteration. Subsequent iterations dereference past the new end and
throw the `Symbol(tick)` error above.

The trigger is rare but not new: any Tick processor that calls
`#removeObject` without a compensating `#insertObject(_, true)`
combined with a same-tick controller downgrade in the same room is
enough.

## Fix

Snapshot `#objects` with `.slice()` at the top of both loops. Same
"skip newly-pushed `now` objects" semantic, plus tolerance for mid-loop
array mutations.

## Considered alternative

The narrower root-cause variant is to drop the `#flushObjects` call
inside `mods/controller/processor.ts` `updateRoomStatus`. Audit:

- `#roomStatusDidChange` is only overridden by `Source` (never
  inserted at runtime) and `StructureExtension` (only created via
  construction-site completion, which uses `now=true` and is already
  in `#objects` without a flush). No queued-insert object in the tree
  overrides `#roomStatusDidChange`.
- `checkActiveStructures` does iterate `#objects` and could transiently
  miscount a queued-remove `OwnedStructure` without the flush; the
  miscount is corrected at the next flush boundary on the same tick.
- `git log -L` shows the flush was added in `a38fe565` ("Ensure
  extension capacity is updated…") when construction completion
  *queued* the new extension. Construction now uses `now=true`, so
  the original load-bearing reason no longer applies.

Snapshot was chosen because B rests on an unwritten invariant ("any
future OwnedStructure override of `#roomStatusDidChange` must be
inserted with `now=true`") while the snapshot makes the loop robust to
any mid-Tick mutation. Happy to swap to B (or layer it on top) if you
prefer the targeted fix.

## Validation

- `pnpm run build`, then `pnpm test` — 374/374 pass on this branch.
  The new `Engine › RoomProcessor` test fails on `main` with the
  exact reported stack trace and passes here.
- `npx eslint --max-warnings 0` on the three changed `.ts` files —
  zero new warnings (room.ts went from 12 → 11 pre-existing warnings;
  new files clean).


laverdet:
What about: remove '#flushObjects' invocation and use '#immediateObjects' instead?

misterwise:
Done — swapped `updateRoomStatus` and `checkActiveStructures` to `#immediateObjects()`, dropped the mid-Tick `#flushObjects`, and reverted the engine-loop snapshot. Pushed as 4da85be2.

Also added three behavior tests pinning the `#roomStatusDidChange` and `checkActiveStructures` side effects across the new path (downgrade flips `#active` on over-cap extensions; reserveController updates source `energyCapacity`; downgrade updates extension `energyCapacity`). Validated each by breaking the relevant call site and confirming the test catches it.

laverdet:
I removed the processor test because it tested an internal condition which doesn't seem like ever happened.

A benchmark for a really good test case is one that fails without the changeset. All of the tests in this PR pass even if you revert the rest of the changeset.

That's not to say it's required that all new tests fix a previous failure, but in those cases you should be more picky about what you add. The other tests here look fine because there wasn't coverage for those runtime conditions.

PR #238, opened 2026-05-23
service: add serviceInitialized and afterTick mod hooks

bastianh:
Adds a 'service' mod context loaded via importMods at startup, alongside the existing 'driver', 'backend', and 'processor' contexts.

Two hooks are defined in engine/service/symbols.ts:

- serviceInitialized(shard): called once when the main loop is ready; returning an Effect registers a cleanup to run on shutdown
- afterTick({ shard, timeTaken, averageTime }): called after every tick

This allows mods to observe tick timing and register background tasks without modifying core engine code.

### DIFF
--- a/packages/xxscreeps/config/mods/index.ts
+++ b/packages/xxscreeps/config/mods/index.ts
@@ -8,7 +8,7 @@ import { defaultAsyncFileSystem } from '@loaderkit/resolve/fs';
 import { configDefaults } from 'xxscreeps/config/config.js';
 import config from 'xxscreeps/config/raw.js';
 
-type Provide = 'backend' | 'config' | 'constants' | 'driver' | 'game' | 'processor' | 'storage' | 'test';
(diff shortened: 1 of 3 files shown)


laverdet:
a1fcfb7b07e3cb2b42724d7b3edcbef8cc46d882 adds a `main` slot which can be used to load on the main loop. You can use the shard channel to listen for tick notifications. There's not a channel notification between loop initialization and first tick, but if you're working on a mod which wants that I'd like to hear about it.

bastianh:
Thank you @laverdet. I'm using the hook in a prometheus mod that I have created .. I have now updated it to the current main but it still needs to hook into tick lifecycle for recording stats about the users code.

on my server it looks like this:  https://xxscreeps.w4rl0ck.dev/metrics example dashboard: https://grafana.w4rl0ck.dev/public-dashboards/525d01383de54c4f8e5229537374875f

thats the branch with the mod: https://github.com/bastianh/xxscreeps/tree/feat/prometheus-mod

PR #239, opened 2026-05-25
auth: allow login with email address in addition to username

bastianh:
Both the form signin endpoint and HTTP Basic Auth middleware now fall back to an email provider lookup when no user is found by username.

This is to be in sync with the usual private server and the official client that allows login by email or username.

### DIFF
--- a/.changeset/auth-email-login-lookup.md
+++ b/.changeset/auth-email-login-lookup.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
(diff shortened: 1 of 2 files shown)


laverdet:
The sets of valid username and valid email have no overlap. So it could be broken into a short helper with a an if statement to lookup by email if it contains '@' or by username otherwise.

bastianh:
thanks for your suggestion .. I have updated the PR and also added a changeset file.

Issue #241, opened 2026-05-30
Isolated VM Memory Leak on Code Reset


bastianh:
I noticed a crash on my xxscreeps private server that looks related to a memory problem on global resets:

```
2026-05-29T19:35:58.502701782Z Tick 1389328 ran in 21ms; avg: 14.34ms
2026-05-29T19:35:58.745955657Z Tick 1389329 ran in 14ms; avg: 14.26ms
2026-05-29T19:35:59.042812412Z Error: Isolate was disposed during execution due to memory limit
2026-05-29T19:35:59.042855928Z     at (<isolated-vm boundary>)
2026-05-29T19:35:59.042861939Z     at /data/node_modules/.pnpm/ivm-inspect@5.0.0_isolated-vm@6.0.2/node_modules/ivm-inspect/dist/index.js:26:46
2026-05-29T19:35:59.055244976Z node: ../src/isolate/environment.h:202: static ivm::IsolateEnvironment& ivm::IsolateEnvironment::GetCurrent(): Assertion `environment != nul
2026-05-29T19:39:26.399007Z    stream closed: EOF for xxscreeps/xxscreeps-67c8896db6-7b2f4 (xxscreeps)
```

I saw on grafana that the memory the container used increased similar to the graph of global resets so I asked kimi to look into that problem.
It found some fixes and I have asked codex to review the patch and the findings.

documentation: 

# Isolated VM Memory Leak on Code Reset

## Summary

When a player submitted new code, the runner correctly detected the stale state, disposed the old sandbox, and created a new one. However, the **old isolate's memory was not fully released** because `IsolatedSandbox.dispose()` failed to drop all native references to the underlying V8 isolate. Over time — especially on active servers with frequent code updates — this caused the runner process RSS to grow without bound.

## Background: How Sandboxes Are Managed

The runner maintains one `PlayerInstance` per active user. Each `PlayerInstance` holds a `Sandbox` (by default an `IsolatedSandbox`) that encapsulates an `isolated-vm` isolate where the player's code runs.

### Reset Path (`PlayerInstance.run`)

When a code update arrives via the `codeChannel`, `this.stale` is set to `true`. On the next tick:

```ts
const wasStale = this.stale;
if (wasStale) {
    this.sandbox?.dispose();   // <- old sandbox disposed here
    this.sandbox = undefined;
    this.seenUsers.clear();
    this.stale = false;
}

if (!this.sandbox) {
    // ... load new code blob ...
[...]

PR #242, opened 2026-05-30
Isolated VM Memory Leak or crash on Code Reset

bastianh:
see #241  for documentation

will deploy it to my testserver today and see what happens

### DIFF
--- a/.changeset/isolated-vm-memory-leak.md
+++ b/.changeset/isolated-vm-memory-leak.md
@@ -1,0 +1,9 @@
+---
+"xxscreeps": patch
+---
+
+Fix isolated VM memory leak on code reset
+
+- `IsolatedSandbox.dispose()` now explicitly releases the `tick` reference and clears the `isolate` handle, allowing the underlying V8 isolate to be fully destroyed.
+- `createSandbox()` now disposes the sandbox if `initialize()` throws, preventing leaked isolates from partially constructed sandboxes.
+- `PlayerInstance.disconnect()` now clears the `sandbox` reference after disposal.
--- a/.changeset/isolated-vm-memory-leak.md
+++ b/.changeset/isolated-vm-memory-leak.md
@@ -3,7 +3,3 @@
 ---
 
 Fix isolated VM memory leak on code reset
-
-- `IsolatedSandbox.dispose()` now explicitly releases the `tick` reference and clears the `isolate` handle, allowing the underlying V8 isolate to be fully destroyed.
-- `createSandbox()` now disposes the sandbox if `initialize()` throws, preventing leaked isolates from partially constructed sandboxes.
-- `PlayerInstance.disconnect()` now clears the `sandbox` reference after disposal.
--- a/packages/xxscreeps/driver/sandbox/index.ts
+++ b/packages/xxscreeps/driver/sandbox/index.ts
@@ -75,7 +75,12 @@ export async function createSandbox(userId: string, payload: InitializationPaylo
 			default: throw new Error(`Invalid sandbox mode: ${config.runner.sandbox}`);
 		}
 	}();
-	await sandbox.initialize(payload);
+	try {
+		await sandbox.initialize(payload);
+	} catch (err) {
+		sandbox.dispose();
+		throw err;
+	}
 	didMakeSandbox(sandbox, userId);
 	return sandbox;
 }
--- a/packages/xxscreeps/driver/sandbox/index.ts
+++ b/packages/xxscreeps/driver/sandbox/index.ts
@@ -75,12 +75,7 @@ export async function createSandbox(userId: string, payload: InitializationPaylo
 			default: throw new Error(`Invalid sandbox mode: ${config.runner.sandbox}`);
 		}
 	}();
-	try {
-		await sandbox.initialize(payload);
-	} catch (err) {
-		sandbox.dispose();
-		throw err;
-	}
+	await sandbox.initialize(payload);
 	didMakeSandbox(sandbox, userId);
 	return sandbox;
 }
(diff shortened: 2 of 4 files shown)


bastianh:
hmm.. stressing the server currently with `Game.cpu.halt()` every 25 ticks and saw that in log:

```
Tick 1743775: +W4rl0ck, +Leebe, -Leebe, Error: Isolate was disposed during execution
    at (<isolated-vm boundary>)
    at run (file:///data/node_modules/.pnpm/xxscreeps@file+..+opt+xxscreeps+xxscreeps.tgz/node_modules/xxscreeps/dist/driver/sandbox/isolated/index.js:101:48)
    at sandbox.run (file:///data/node_modules/.pnpm/xxscreeps@file+..+opt+xxscreeps+xxscreeps.tgz/node_modules/xxscreeps/dist/mods/prometheus/driver.js:26:46)
    at file:///data/node_modules/.pnpm/xxscreeps@file+..+opt+xxscreeps+xxscreeps.tgz/node_modules/xxscreeps/dist/engine/runner/instance.js:179:43
-W4rl0ck, ran
Tick 1743776: +W4rl0ck, +Leebe, -W4rl0ck, -Leebe, ran
Tick 1743777: +W4rl0ck, +Leebe, -Leebe, -W4rl0ck, ran
Tick 1743778: +W4rl0ck, +Leebe, -Leebe, -W4rl0ck, ran
Tick 1743779: +W4rl0ck, +Leebe, -Leebe, -W4rl0ck, ran
Tick 1743780: +W4rl0ck, +Leebe, -Leebe, -W4rl0ck, ran
Tick 1743781: +W4rl0ck, +Leebe, -Leebe, -W4rl0ck, ran
malloc(): unaligned tcache chunk detected
stream closed: EOF for xxscreeps/xxscreeps-65b8f7f98f-cqj2n (runner)
```

bastianh:
ah .. commits 3 and 4 had some more information in the commit message .. not very visible in this pr:

```
fix: correct memoryLimit unit bug and prevent concurrent-OOM crash in IsolatedSandbox
Root cause of the 850 MB growth + GetCurrent() assertion crash:

1. memoryLimit unit mismatch (primary cause)

   IVM's memoryLimit option is in MB, but the old formula added
   data.terrainBlob.length (bytes) directly:

     memoryLimit: config.runner.cpu.memoryLimit + data.terrainBlob.length
     // e.g. 256 + 62500 = 62756 … interpreted by IVM as 62756 MB (~61 GB)

   This made the per-isolate heap limit effectively unlimited, so user
   code could consume unbounded memory and grow the process RSS until the
   container ran out of memory. Fixed by converting the terrain blob size
   to MB before adding:

     memoryLimit: config.runner.cpu.memoryLimit + Math.ceil(data.terrainBlob.byteLength / (1024 * 1024))

   This enforces the intended ~256 MB per-isolate cap, preventing any single
   user from inflating the process RSS.

2. Concurrent-OOM crash: GetCurrent() assertion (secondary cause)

   When the process ran out of memory, a new isolate could OOM during
   initialize(). The old code ran ivmInspect.create(), pf.create(), and
   isolate.compileScript() concurrently via Promise.all. ivm-inspect itself
   also uses Promise.all internally. When OOM auto-disposed the isolate mid-
   flight, the other in-flight branches on the same (now-disposed) isolate
   called IsolateEnvironment::GetCurrent() — which asserts on a null
   environment — crashing the process.

   Fixed by running the setup steps sequentially: ivm-inspect first (and
   alone), then pathfinder, then compileScript. If any step OOMs, nothing
   else is racing on the same isolate.

3. Minor: release IVM references that were leaking as GC-pending wrappers

   - instance.derefInto() → instance.derefInto({ release: true }) for the
     PathFinder module instance — releases the JS-side IVM reference
     immediately instead of waiting for GC.
   - script.run(...) → script.run(..., { release: true, ... }) — releases
     the compiled Script reference after it has been run.
```

and 

```
Two follow-up fixes based on crash log analysis (malloc: unaligned tcache
chunk, triggered ~6 ticks after an OOM-during-execution event):

1. IsolatedSandbox.run() – catch all disposal-during-execution errors

   The existing catch only matched 'Isolate is disposed'. The variant thrown
   when the isolate is disposed *while running* user code (including OOM
   disposal) is 'Isolate was disposed during execution [due to memory limit]'.
   Because this wasn't caught, the error propagated through callers that wrap
   sandbox.run (notably the prometheus sandboxCreated hook), where it was
   unhandled and re-thrown to instance.ts's outer catch instead of returning
   { result: 'disposed' } cleanly.

   Fixed by matching any message starting with 'Isolate was disposed'.

2. PlayerInstance.run() – set stale=true in the unexpected-error catch path

   When sandbox.run() threw unexpectedly (after the bug above), the outer
   catch in instance.ts logged the error and returned undefined. Because
   result was falsy, stale was never set, so the broken sandbox stayed alive
   for one full extra tick before the following tick's 'disposed' result
   finally triggered the reset. The corrupted sandbox was therefore used one
   more time.

   Fixed by setting this.stale = true in the catch block, so the sandbox is
   replaced on the very next tick.

Note: the underlying C-level heap corruption on concurrent OOM-disposal is
a bug in isolated-vm 6.0.2 itself. These fixes prevent the broken sandbox
from being reused after the event, but upgrading isolated-vm when a fixed
version is available is still advisable.
```

laverdet:
I removed the initialization changes (besides release:true). If this section is failing then we have bigger problems.

Also removed other over-defensive spackling against invariants like invoking `run` after `dispose`.

Also tweaked the internal error case. The logic should be handled down in the other branch where all the other completion checks go. Really the internal error case should probably just crash the process or something because who knows what's going on at that point.

PR #243, opened 2026-05-31
backend: render room-socket updates at the tick their blob was loaded for

misterwise:
## Bug

Opening a room in the client intermittently crashes the room socket:

```
Error: Invalid expiry time 340535 vs 340537
  requiredExpiryTime            game/object.ts
  StructureRampart.ticksToDecay mods/defense/rampart.ts
  rampart renderer              mods/defense/backend.ts
  room render                   backend/sockets/room.ts
```

Backend-only. The processor never throws — ticks keep advancing — and the room is healthy on a fresh `loadRoom` a few ticks later. So the persisted `#nextDecayTime` is **not** actually overdue; the backend reads it under the wrong clock.

## Cause

In `subscribeToRoom`, the `time` closure is advanced by the `channel/game` tick listener on every tick. The throttled publisher then does:

```js
state.room = await shard.loadRoom(roomName, time);  // loads blob[time]
state.time = time;
publish(state.room, time, didUpdate);               // renders at `time`
```

A tick can land **during** `await loadRoom`, so the blob saved for tick `T` is published at a later `Game.time` (`T+k`). The renderer computes `Game.time + rampart.ticksToDecay`, and `requiredExpiryTime` throws when `#nextDecayTime < Game.time`. A rampart at its decay boundary (`#nextDecayTime == T`) is valid at `T` but overdue at `T+k`. Rooms are double-buffered (`room${time % 2}`), so a stale slot doesn't help; faster ticks just widen `k`.

Instrumenting the publisher confirmed the seam directly (one open room, ~34ms ticks):

```
loaded tick 340653, time advanced to 340656 during load
loaded tick 340660, time advanced to 340662 during load
```

## Fix

Snapshot the render tick once and use it for both the load and the publish:

```diff
 mustNotReject(async () => {
+	const renderTime = time;
 	if (didUpdate) {
-		state.room = await shard.loadRoom(roomName, time);
+		state.room = await shard.loadRoom(roomName, renderTime);
 	}
-	state.time = time;
-	publish(state.room, time, didUpdate);
+	state.time = renderTime;
+	publish(state.room, renderTime, didUpdate);
 	didUpdate = false;
 	timer.set(config.backend.socketThrottle);
 });
```

## Why this and not the decay tripwire

The `requiredExpiryTime` throw is left untouched (`mods/defense/backend.ts` and `game/object.ts` unchanged). This fix renders `blob[T]` at `Game.time = T`, so a *genuinely* overdue blob — a real processor decay skip — still throws `T-1 vs T`. It only removes the case where the reader supplied a future `Game.time`. A clamp/floor on the decay value would have suppressed a real skip; this doesn't.

## Test

`backend/sockets/test.ts` drives a tick into the middle of the publisher's `await loadRoom` (deterministic via in-process pubsub) and asserts the `didUpdate` render lands at the loaded tick. Fails on `main` (`2 !== 1`), passes with the fix.


laverdet:
I removed the test case here because it's huge and touches a lot of internals. You can skip tests in the future if a fix is self-evident.

PR #246, opened 2026-06-07
pf: treat CostMatrix value 255 as impassable

misterwise:
The rewrite of `cost_matrix_look` dropped the `0xff` → impassable case, so a
CostMatrix value of 255 reads as a finite cost of 255 instead of blocking. The
search walks through walled-off tiles and `findClosestByPath` treats them as
reachable. Terrain walls still block (their `look_table_` slot is `obstacle`),
so it went unnoticed.

`obstacle` moves to `:room` because `room.cc` now references it and `:room`
can't import `:pf`.

## Validation

Verified against screeps-ok:
- COSTMATRIX-007 — CostMatrix 255 is unwalkable
- PATHFINDER-012 — `search` returns `incomplete` with a partial path when blocked
- ROOMPOS-FIND-007 — `findClosestByPath` returns null for an unreachable target

### DIFF
--- a/.changeset/pathfinder-cost-matrix-255-impassable.md
+++ b/.changeset/pathfinder-cost-matrix-255-impassable.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Treat a CostMatrix value of 255 as an impassable obstacle.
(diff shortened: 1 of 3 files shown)


misterwise:
This looks like it was handled in https://github.com/laverdet/xxscreeps/commit/e61801707e21b13a482ea2a35fc22bf5c417c39a

PR #247, opened 2026-06-10
cli: remove CommonJS host shims from the eval console

misterwise:
`installHostShims` put `require`, `__filename`, and `__dirname` on `globalThis` for the eval/REPL console — CommonJS globals salvaged from the closed #227. Per review they aren't needed: the eval surface is ESM-only (`SourceTextModule` + dynamic `import()`, static `import` rejected), so `require` was never load-bearing and `import.meta.url` covers the path globals.

This removes them from both the one-shot eval path and the REPL. Nothing is installed, so nothing leaks — the Disposable restoration from this PR's prior revision is gone too. README realm note updated to match.

## Validation

- Confirmed nothing in `cli/` or the eval engine references the three globals (grep); the ESM `import()` loading path is untouched.
- `tsc -b` clean; eslint over the changed files: no new warnings.

### DIFF
--- a/.changeset/cli-remove-cjs-shims.md
+++ b/.changeset/cli-remove-cjs-shims.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Remove CommonJS `require`/`__filename`/`__dirname` shims from the eval console.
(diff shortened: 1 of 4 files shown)


laverdet:
What are these here for anyway? These are old CommonJS properties and my position on CJS is that continued tolerance of it is an act of violence on the JavaScript community. Do we need them in the administrative console?

misterwise:
> What are these here for anyway? These are old CommonJS properties and my position on CJS is that continued tolerance of it is an act of violence on the JavaScript community. Do we need them in the administrative console?

Not needed, and I'll pivot this to removing them rather than cleaning up the leak.

PR #166, opened 2026-04-29
Add Deposit mod

misterwise:
## Summary
- Add `xxscreeps/mods/deposit` with Deposit object registration, find/look support, backend rendering, harvest processing, cooldown tracking, and decay removal.
- Add Deposit to the classic mod set and depend on factory for deposit resource registration.
- Add internal Deposit tests and a patch changeset.

## Scope
- This PR implements the seeded Deposit engine surface exercised by the existing screeps-ok catalog.
- Highway deposit generation, stronghold deposit loot, power creep deposit interactions, and screeps-ok post-merge adapter/docs updates are left for follow-up work.
- Constants/resource enum cleanup is intentionally deferred so this diff stays focused on Deposit behavior. The current import/dependency path is covered by build, unit tests, and screeps-ok xxscreeps parity.

## Testing
- `npm run build`
- `npm test -- Deposit`
- `npm test`
- `screeps-ok`:
  - `1161/1342 passed, 18 expected failures, 163 skipped`
  - Deposit-related cases passed: `DEPOSIT-*`, `DEPOSIT-HARVEST-*`, `SHAPE-DEPOSIT-001`

PR #249, opened 2026-06-11
Place deposits in highway-room sectors

misterwise:
The Deposit mod (#166) added the object, harvest, and decay, but nothing creates deposits. This generates them in highway-room sectors: a per-sector wall-clock schedule in `shard.scratch`, drained by a shard tick processor, precipitates a deposit into a free ring room when the sector's throughput is below target (tile selection deferred to the room processor via an internal `precipitateDeposit` intent). Decay marks its owning sector due immediately, refilling freed capacity the same tick.

Reworked from review:
- Storage accessors split into `mods/deposit/model.ts`; the schedule lives in `shard.scratch` (rebuildable on restart).
- One-time seeding moved to a new `registerShardInitializer` hook (run once from `main`), replacing a per-tick bootstrap-flag read.
- `game/room/sector.ts` rebuilt on `name.ts`'s signed-coordinate parser (no `kMaxWorldSize2`), with generators and `makeSectorRadiusFilter`.

Deviations from official Screeps, flagged in code:
- Deposit type is hash-derived (stable per room) because imported worlds don't carry the generator-pinned `depositType`. TODO at the constant.
- The 250-tile radius is enforced at placement; the official loop computes `inSector` but never checks it. A test pins that out-of-radius deposits decay without confusing the scheduler.

## Validation
- `npx xxscreeps test` — 404 passed, including five precipitation specs: bootstrap seeds one deposit per pass on wall terrain inside the radius, saturated sectors don't gain a second, occupied rooms are excluded, decay drives a same-tick re-eval/refill, and out-of-radius decay yields no refill.
- Revert-verified each spawn spec — inverting its production line (init seeding, saturation threshold, busy-room exclusion, decay scheduling, radius gate) fails the matching test.
- Verified `sector.ts` behavior-identical to the prior offset-coordinate math via a full-coordinate-space sweep (0 mismatches across all 65,536 rooms and every edge-room position).


packages/xxscreeps/game/room/sector.ts
```diff
@@ -0,0 +1,71 @@
...
+	const { rx, ry } = parseRoomName(roomName);
+	const xCandidates: number[] = [];
+	const yCandidates: number[] = [];
+	for (let dd = -5; dd <= 5; ++dd) {
+		if (isCentralAxis(rx + dd)) xCandidates.push(rx + dd);
+		if (isCentralAxis(ry + dd)) yCandidates.push(ry + dd);
+	}
+	const out: string[] = [];
+	for (const cx of xCandidates) {
+		const xOnRing = Math.abs(rx - cx) === 5;
```
laverdet: I'm struggling to follow this. Why can't you do this check in the loop above? Then you wouldn't need these two candidate temporaries. `cx` is `rx + dd` but then you subtract it from `rx` in this loop so basically you just have `dd` which was your loop iterator in the first place.

packages/xxscreeps/game/room/sector.ts
```diff
@@ -0,0 +1,71 @@
...
+function isCentralAxis(coord: number): boolean {
+	return coord < kMaxWorldSize2 ? coord % 10 === 2 : coord % 10 === 3;
+}
+
+function isCentralCoord(rx: number, ry: number): boolean {
+	return isCentralAxis(rx) && isCentralAxis(ry);
+}
+
+export function isCentralRoom(roomName: string): boolean {
+	const { rx, ry } = parseRoomName(roomName);
```
laverdet: Could you drop this function into `name.ts` and base the calculations on the result of it? This is from some work that I have on pause. `kMaxWorldSize2` is a kludge and limits us to 256x256, which I would hate to see baked into this logic.

```ts
export function parseSignedRoomName(name: string) {
 // Parse X and calculate str position of Y
 const rx = parseInt(name.slice(1), 10);
 let verticalPos = 2;
 if (rx >= 100) {
     verticalPos = 4;
 } else if (rx >= 10) {
     verticalPos = 3;
 }
 // Parse Y and return adjusted coordinates
 const ry = parseInt(name.slice(verticalPos + 1), 10);
 const horizontalDir = name.charAt(0);
 const verticalDir = name.charAt(verticalPos);
 // The range of int8 is -128 to 127, since someone has to "pay" for 0 and it's the positive side.
 // We pay it back here by accounting for it on the W and N (negative) sides. So the range is a
 // nice round W127N127 to E127S127.
 return {
     rx: horizontalDir === 'W' || horizontalDir === 'w' ? -1 - rx : rx,
     ry: verticalDir === 'N' || verticalDir === 'n' ? -1 - ry : ry,
 };
}
```

packages/xxscreeps/game/room/sector.ts
laverdet: Also now that I'm understanding the deposit logic more closely I think you can just do this whole function with modulus and 4-5 if's. I bet Claude would nail that in one shot.

packages/xxscreeps/game/room/sector.ts
```diff
@@ -0,0 +1,71 @@
...
+		}
+	}
+	return out;
+}
+
+const SECTOR_HALF_EXTENT = 250;
+
+// Returns a tile predicate for `roomName`'s membership in `centralRoom`'s 250-tile sector
+// radius. Parses both room names once; the closure does two adds and two abs per call.
+export function sectorContainsTile(centralRoom: string, roomName: string): (xx: number, yy: number) => boolean {
```
laverdet: If I'm reading this logic correctly this serves to disambiguate positions in non-central axis highway rooms. Positions in those rooms could belong to either 2 or 4 sectors, depending on which side or quadrant the position is.

Can you clarify this a bit in the code. `tile` isn't really a term we use too much in screeps. We have rooms and positions. I used "tile" occasionally in the pathfinder but I find it confusing here. It should also make the invariant of "`roomName` must be a highway" more clear.

Also, if this is on the central axis highway it could return an always-true predicate.

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
+import type { Shard } from 'xxscreeps/engine/db/index.js';
```
laverdet: `spawn.ts` is going to drive me crazy since that means something entirely different in screeps. How about "precipitate".

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
...
+import { pushIntentsForRoomNextTick } from 'xxscreeps/engine/processor/model.js';
+import { Fn } from 'xxscreeps/functional/fn.js';
+import * as C from 'xxscreeps/game/constants/index.js';
+import { Game } from 'xxscreeps/game/index.js';
+import * as RoomObject from 'xxscreeps/game/object.js';
+import { RoomPosition } from 'xxscreeps/game/position.js';
+import { Room as RoomClass } from 'xxscreeps/game/room/index.js';
+import { isCentralRoom, sectorContainsTile, sectorEdgeRooms } from 'xxscreeps/game/room/sector.js';
+import {
+	DEPOSIT_DECAY_TIME, DEPOSIT_EXHAUST_MULTIPLY, DEPOSIT_EXHAUST_POW,
```
laverdet: I prefer not to break imports on multiple lines. Literally, I never read these with my eyes so it can just be one long line.

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
...
+const MAX_TILE_ATTEMPTS = 1000;
+
+// FNV-1a used both to pin deposit types and to scatter bootstrap scores per sector.
+function fnv1a(str: string): number {
+	let hash = 0x811c9dc5;
+	for (let ii = 0; ii < str.length; ++ii) {
+		hash ^= str.charCodeAt(ii);
+		hash = Math.imul(hash, 0x01000193);
+	}
+	return hash >>> 0;
```
laverdet: Just return `hash` since it's the natural type of this operation in JavaScript. Then replace `fnv1a(roomName) % DEPOSIT_TYPES.length` later with `Math.abs(fnv1a(roomName) % DEPOSIT_TYPES.length)`

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
...
+) {
+	await shard.data.zAdd(dueSectorsKey, [ [ dueAt, sector ] ],
+		options?.earliest === true ? { up: 'LT' } : undefined);
+}
+
+function depositThroughput(harvested: number): number {
+	return 20 / Math.max(1, DEPOSIT_EXHAUST_MULTIPLY * harvested ** DEPOSIT_EXHAUST_POW);
+}
+
+type SectorEvalResult = {
```
laverdet: interface

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
...
+async function loadSectorDeposits(
+	shard: Shard, world: World, centralRoom: string, time: number,
+): Promise<SectorEvalResult & { normalEdges: string[] }> {
+	// Out-of-borders / closed rooms are excluded from both throughput tallying and spawning.
+	const normalEdges = sectorEdgeRooms(centralRoom).filter(name =>
+		world.map.getRoomStatus(name).status === 'normal');
+	const depositsByRoom = new Map<string, Deposit[]>();
+	let throughput = 0;
+	await Fn.mapAwait(normalEdges, async edgeRoom => {
+		const room = await shard.loadRoom(edgeRoom).catch(() => undefined);
```
laverdet: Is there a reason to swallow this exception?

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
...
+			// Rooms process tick `time` with Game.time = time + 1, so a deposit whose decay
+			// time has arrived (<= time + 1) was removed by this tick's room pass — exclude
+			// the corpse so the decay path's same-tick re-eval doesn't tally it.
+			if (object instanceof Deposit && object['#nextDecayTime'] > time + 1 &&
+				inSector(object.pos.x, object.pos.y)) {
+				all.push(object);
+			}
+		}
+		if (all.length > 0) {
+			depositsByRoom.set(edgeRoom, all);
```
laverdet: This closure is pretty smelly. The only time `depositsByRoom` is used outside the top function is `new Set(depositsByRoom.keys())`, so it seems like it shouldn't be part of the overall return value. You should just be returning a list of rooms that have deposits.

Using a closure and Map here for `depositsByRoom` is weird. You're in a function and you can return a value. This whole thing should just be `return room['#objects'].filter(/* the predicate */)`.

To get throughput it's:
```
const throughput = Fn.accumulate(Fn.transform(
	allDeposits,
	deposits =>
		Fn.map(deposits, deposit => depositThroughput(deposit['#harvested']))));
```

To get rooms it's:
```ts
const roomNames =
	Fn.transform(allDeposits, objects => Fn.map(objects, object => object.room.name));
```

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
...
+	return { throughput, depositsByRoom, normalEdges };
+}
+
+type RngFn = () => number;
+
+// Picks a wall tile in 5..44 with at least one non-wall neighbour (incl. diagonals), inside
+// the 250-tile sector radius, and 2 tiles clear of any other room object.
+function findSpawnTile(
+	world: World, centralRoom: string, targetRoom: Room, rng: RngFn,
+): { x: number; y: number } | undefined {
```
laverdet: `return { xx, yy } as const` and remove annotation. I don't want to be annotating with dictionary alias types anymore. If we want explicit return annotation it should be a named type or a basic union of named types.

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
...
+	const terrain = world.map.getRoomTerrain(targetRoom.name);
+	const objects = targetRoom['#objects'];
+	const inSector = sectorContainsTile(centralRoom, targetRoom.name);
+	for (let attempt = 0; attempt < MAX_TILE_ATTEMPTS; ++attempt) {
+		const xx = Math.floor(rng() * 40) + 5;
+		const yy = Math.floor(rng() * 40) + 5;
+		if ((terrain.get(xx, yy) & C.TERRAIN_MASK_WALL) === 0) continue;
```
laverdet: I know the constant is called `TERRAIN_MASK_WALL` but it's not a mask. Well, it kind of is. Anyway just compare to `TERRAIN_MASK_WALL`, don't bitwise it.

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
...
+		// Divergence from the official cron, which computes this check but never enforces it.
+		if (!inSector(xx, yy)) continue;
+		let hasExit = false;
```
laverdet: These loops have weird exit conditions. Almost like Claude is afraid of `break`. Just make this an IIFE `const hasExit = function() { ... }()` and `hasExit = true;` would become `return true`.

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
...
+		for (let dx = -1; dx <= 1 && !hasExit; ++dx) {
+			for (let dy = -1; dy <= 1 && !hasExit; ++dy) {
+				if ((terrain.get(xx + dx, yy + dy) & C.TERRAIN_MASK_WALL) === 0) {
+					hasExit = true;
+				}
+			}
+		}
+		if (!hasExit) continue;
+		if (Fn.some(objects, obj => Math.abs(obj.pos.x - xx) <= 2 && Math.abs(obj.pos.y - yy) <= 2)) continue;
```
laverdet: `obj.pos.getRangeTo`

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
...
+		return { x: xx, y: yy };
+	}
+	return undefined;
+}
+
+// Tests can swap in a seeded RNG via `setDepositSpawnRng`.
+let rng: RngFn = Math.random;
+export function setDepositSpawnRng(fn: RngFn | undefined) {
+	rng = fn ?? Math.random;
```
laverdet: `setDepositSpawnRandomForTesting`. Also make `Disposable`

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
...
+type ScatterFn = (roomName: string) => number;
+const defaultScatter: ScatterFn = roomName => fnv1a(roomName) % DEPOSIT_CHECK_INTERVAL;
+let bootstrapScatter: ScatterFn = defaultScatter;
+export function setDepositBootstrapScatterForTest(fn: ScatterFn | undefined) {
+	bootstrapScatter = fn ?? defaultScatter;
+}
+
+function pickRandomFreeRoom(edges: string[], busyRooms: Set<string>): string | undefined {
+	const free = edges.filter(name => !busyRooms.has(name));
+	if (free.length === 0) return undefined;
```
laverdet: This `if` doesn't really do anything.

packages/xxscreeps/mods/deposit/spawn.ts
```diff
@@ -0,0 +1,254 @@
...
+		await shard.data.zAdd(dueSectorsKey, seeds, { up: 'LT' });
+	}
+	await shard.data.set(bootstrapFlagKey, '1');
+}
+
+// Peek-and-reschedule (no zrem): a crash between peek and reschedule leaves the original entry
+// for retry next tick instead of dropping it silently.
+registerShardTickProcessor(async (shard, time) => {
+	const now = Date.now();
+	const bootstrapDone = await shard.data.get(bootstrapFlagKey);
```
laverdet: So every tick we're accessing a database key that will pretty much always be set? Make a new initialization hook and do it once on start.

packages/xxscreeps/mods/deposit/test.ts
```diff
@@ -106,3 +110,168 @@ describe('Deposit', () => {
...
+	return () => {
+		state = (Math.imul(state, 1664525) + 1013904223) >>> 0;
+		return state / 0x100000000;
+	};
+}
+
+// Scope a deterministic RNG (and a zero-scatter bootstrap) to the test; restores defaults on
+// disposal. Disabling scatter keeps the single-sector test world firing on tick 1.
+function withFixedSpawnRng(seed = 1): Disposable {
+	setDepositSpawnRng(makeRng(seed));
```
laverdet: This isn't a good pattern. If you want to compose multiple disposables you use DisposableStack

packages/xxscreeps/mods/deposit/test.ts
```diff
@@ -106,3 +110,168 @@ describe('Deposit', () => {
...
+			setDepositSpawnRng(undefined);
+			setDepositBootstrapScatterForTest(undefined);
+		},
+	};
+}
+
+async function findDepositsInSector(shard: Shard, centralRoom: string) {
+	const out: { roomName: string; deposit: Deposit }[] = [];
+	for (const edgeRoom of sectorEdgeRooms(centralRoom)) {
+		const room = await shard.loadRoom(edgeRoom).catch(() => undefined);
```
laverdet: Don't swallow exceptions.

packages/xxscreeps/mods/deposit/test.ts
```diff
@@ -106,3 +110,168 @@ describe('Deposit', () => {
...
+		if (!room) continue;
+		for (const object of room['#objects']) {
+			if (object instanceof Deposit) {
+				out.push({ roomName: edgeRoom, deposit: object });
```
laverdet: Is this whole thing just a different version of `loadSectorDeposits`?

packages/xxscreeps/mods/deposit/test.ts
```diff
@@ -106,3 +110,168 @@ describe('Deposit', () => {
...
+	return {
+		[Symbol.dispose]() {
+			setDepositSpawnRng(undefined);
+			setDepositBootstrapScatterForTest(undefined);
+		},
+	};
+}
+
+async function findDepositsInSector(shard: Shard, centralRoom: string) {
+	const out: { roomName: string; deposit: Deposit }[] = [];
```
laverdet: If you used an IIFE generator function you wouldn't need this type annotation.

packages/xxscreeps/test/import.ts
```diff
@@ -131,8 +131,10 @@ export async function instantiateTestShard() {
 		shard.data.set('terrain', terrain),
 		shard.data.set('time', shard.time),
 		shard.data.sAdd('rooms', [ ...Fn.map(rooms, room => room.room.name) ]),
-		Promise.all(Fn.map(rooms, ({ room }) =>
-			shard.saveRoom(room.name, shard.time, room))),
+		Promise.all(Fn.map(rooms, async ({ room }) => {
+			await shard.saveRoom(room.name, shard.time, room);
+			await shard.copyRoomFromPreviousTick(room.name, shard.time + 1);
```
laverdet: Is this needed?

laverdet (commented):
Additionally:
- Split out database functions into model.ts
- "Due" keys should live in scratch. It's fine if we drop wall time on processor restart
- Bootstrap flag has to go

Also as a general rule if you find yourself writing something like:
```ts
const out: whatever[] = [];
...
return out
```

Then it means you actually want a generator function.

packages/xxscreeps/mods/deposit/spawn.ts
misterwise: changed the language to place since it's easier to type repeatedly, but can use precipitate if you prefer.

packages/xxscreeps/game/room/sector.ts
misterwise: reframed use of tile to pos or roomPos where possible

PR #250, opened 2026-06-11
Throw on intent save when the receiver no longer exists

### LINKED ISSUE #205: Stale cached RoomObject references can act after live object removal
## Summary

Cached player-facing `RoomObject` references can keep acting after the live object has been removed — but the exposure is much narrower than this issue originally claimed.

xxscreeps detaches every room buffer at end of tick (`driver/runtime/index.ts`), so cross-tick access to a stale wrapper generally throws `Accessed a released object from a previous tick`. The real loophole is overlay memoization: `schema/overlay.ts` caches every member except raw int/bool reads on the instance after first access, and `RoomObject.room` is a plain JS property assigned at insert. A public method whose check + enqueue path touches only memoized or JS-assigned fields can validate against a dead receiver and enqueue an intent where vanilla throws:

```text
Could not find an object with ID <id>
```

## Confirmed example

`ConstructionSite.remove()`:

```js
// tick N
global.site = Game.getObjectById(siteId);
global.site.remove(); // OK — and memoizes #user, room.name, id on the instance

// tick N+1, after the site is gone
global.site.remove(); // vanilla throws; xxscreeps returns OK and enqueues a no-op
```

The repro depends on the first call: it memoizes everything the second call needs. `checkRemove` reads only `my` (backed by `#user`, an Id — memoized) and `room` (JS property); `IntentManager.save` reads `room.name` and `id`, both memoized by then. A wrapper that was cached but never touched throws the released-object error instead.

On the vanilla side the throw is deterministic for a different reason than it first appears: vanilla property getters memoize per instance too (`this['_my']` caching in `utils.defineGameObjectProperties`), but `remove()` reads `data(this.id)` inline at the intent push, and the `data` closure reads the module-level `runtimeData` that `make()` reassigns every tick. So the method body, not the getters, enforces the current-object lookup.

## Why the original suspect list mostly doesn't apply

Spawn/tower/link/lab/terminal/factory/observer/creep check chains all read int-typed members — store amounts, cooldowns, hits, fatigue, energy — which are never memoized, so a stale receiver throws the released-object error before reaching the enqueue. Even `Structure.notifyWhenAttacked()` is out: it reads `#noAttackNotify`, a bool, also never memoized.

The residual divergence for those methods is the error message (`Accessed a released object from a previous tick` vs `Could not find an object with ID`). Both throw; chasing message parity there does not seem worthwhile.

Stale *target* (as opposed to receiver) validation is a separate question and remains unverified.

## Impact

Bounded to the player-facing return value. The room processor already drops object intents whose id doesn't resolve (`Game.getObjectById(id)` guard in `engine/processor/room.ts`), so the enqueued no-op never touches state — worst case is `OK` instead of a throw, plus 0.2 intent CPU.

## Fix

A single existence guard at the shared enqueue point, `IntentManager.save`:

```ts
if (Game.getObjectById(receiver.id) === null) {
	throw new Error(`Could not find an object with ID ${receiver.id}`);
}
```

Existence, not identity — a cached wrapper whose object is still alive keeps working, same as vanilla where intents enqueue by id against current data. This covers every receiver in one place (the backend-intent injection in `driver/runtime/index.ts` already resolves receivers through `Game.getObjectById` before calling `save`); no per-method `requireCurrentReceiver` helper is needed.

Related vanilla wart spotted during the audit (upstream PR candidate): `StructureSpawn.recycleCreep` reads `runtimeData.roomObjects[this.id].off` directly instead of routing through `data(this.id)`, so a stale spawn crashes with a raw TypeError instead of the clean missing-object error.

misterwise:
Fixes #205.

`IntentManager.save` now throws `Could not find an object with ID <id>` when `BufferObject.check(receiver)` is false — i.e. the wrapper's buffer was released at end of tick. That catches stale wrappers which pass a method's validation through memoized fields and act on a dead receiver, the confirmed case being a `ConstructionSite.remove()` cached across ticks that re-enqueued `remove` and returned `OK` after the site was gone.

## Validation

- Confirmed `remove()` on a wrapper whose room buffer was detached (as end-of-tick does) throws `Could not find an object with ID <id>`.
- Full xxscreeps suite: 399 passed.

### DIFF
--- a/.changeset/intent-save-missing-receiver.md
+++ b/.changeset/intent-save-missing-receiver.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Throw `Could not find an object` when saving an intent for a stale object reference
--- a/packages/xxscreeps/game/intents.ts
+++ b/packages/xxscreeps/game/intents.ts
@@ -3,6 +3,7 @@ import type { Room } from './room/index.js';
 import type { IntentParameters, IntentReceivers, IntentsForReceiver } from 'xxscreeps/engine/processor/index.js';
 import type { ObjectReceivers, RoomIntentPayload } from 'xxscreeps/engine/processor/room.js';
 import type { Dictionary } from 'xxscreeps/utility/types.js';
+import * as BufferObject from 'xxscreeps/schema/buffer-object.js';
 import * as C from './constants/index.js';
 
 const kCpuCost = 0.2;
@@ -47,6 +48,9 @@ export class IntentManager {
 		Action extends IntentsForReceiver<any>,
 		//Action extends IntentsForReceiver<Receiver>,
 	>(receiver: Receiver, intent: Action, ...args: IntentParameters<Receiver, Action>) {
+		if (!BufferObject.check(receiver)) {
+			throw new Error(`Could not find an object with ID ${receiver.id}`);
+		}
 		const intents = this.makeIntentsForRoom(receiver.room.name).object[receiver.id] ??= {};
 		if (intents[intent as keyof typeof intents] === undefined) {
 			this.cpu += kCpuCost;


misterwise:
I'm running into console errors that may be related to this, I would hold on this merge for now.

edit: It was actually doing its job, my bot was trying to move creeps on the very tick they spawned because of a latent bug patched in #261

laverdet:
Could this be done with `BufferObject.check(receiver)`? From `import * as BufferObject from 'xxscreeps/schema/buffer-object.js';`

misterwise:
Yes, it can, switched to `BufferObject.check(receiver)` and it's cheaper than a per-intent `getObjectById` lookup.

PR #251, opened 2026-06-12
Add StructureInvaderCore NPC actions

misterwise:
## Summary

`StructureInvaderCore` gets `reserveController` / `attackController` /
`upgradeController` / `transferEnergy` intents, driven by a per-tick NPC
behavior loop, plus an action log and a collapse-expiry path. Only the
controller behaviors are wired for now — the stronghold-template behaviors
(refill towers/creeps) depend on deployed-stronghold structures that don't
exist yet.

- The four core actions are `#private` methods (`#reserveController` etc.) —
  they're NPC-only, called from the invader loop, not part of the player-facing
  structure API.
- NPC `'2'` is woken at the core's create/insert site (the same way
  `requestInvader` does for creeps) rather than from a `#beforeInsert` hook.
- Collapse expiry is a silent removal: no ruin and no `EVENT_OBJECT_DESTROYED`
  — those only emit from the damage-destroy path. The NPC reservation is left
  to run out on its own; the controller's existing tick processor releases it
  at expiry.
- `upgradeController` stores the controller invulnerability window in a new
  `StructureController` scalar `#upgradeInvulnerableUntil`, reported through
  the controller's `effects` getter (second producer on the #215 shape).
- `ticksToDeploy` uses nuke's `timeToLand` `+ 1 - 1` form so the getter clamps
  to `undefined` past expiry instead of throwing; the tick processor zeroes the
  stored field once it elapses.
- `rangedMassAttack`'s walk skips invulnerable targets so no `EVENT_ATTACK` is
  emitted for damage that wasn't applied, matching the intent-level gates in
  `attack` / `rangedAttack` / `dismantle`.
- A `checkSourceAlive` guard drops intents from a core destroyed earlier in
  the same intent pass; the four intent checks re-verify ownership through
  `checkMyStructure`, same shape as `checkTower`. Oversized `transferEnergy`
  amounts clamp at the processor instead of failing the check.
- One flagged divergence: the core's `attackController` logs an `attack`
  action; vanilla logs it under `reserveController` — a copy/paste slip in its
  core intent (the creep intent logs `attack`).

### Test harness

`simulate` now seeds rooms with an active NPC into the processor's active set,
mirroring the main-service boot that double-buffers every room once.
`updateUserRoomRelationships` only enqueues rooms via their non-NPC players, so
an NPC-only room never got a first processing tick under test — the
invader-core cases previously worked around this with a stand-in presence
creep, which this removes. In production the boot already seeds these rooms and
an active NPC keeps them awake (6873b3ca), so this only closes the harness gap.

## Validation

- `pnpm run build` clean; `npx xxscreeps test`: 426 passed. Ten invader-core
  cases cover deploy-phase damage blocking, deploy+collapse effects
  composition, reserve/attack/upgrade behaviors and reservation accumulation,
  tower refill, and collapse expiry vs damage-destroy (ruin, event, and
  reservation outcomes); the NPC-loop cases now run on the harness seed without
  a presence creep.
- `eslint` over the changed files: no new warnings.
- Live smoke on a dev server: injected a level-2 core into an empty
  controller room, observed deploy at the configured tick, `reserveController`
  action-logging with the reservation accumulating, then silent collapse with
  the reservation left ticking down.

### DIFF
--- a/packages/xxscreeps/mods/invader/invader-core.ts
+++ b/packages/xxscreeps/mods/invader/invader-core.ts
@@ -41,15 +41,8 @@ export class StructureInvaderCore extends withOverlay(OwnedStructure, shape) {
 	@enumerable get spawning(): null { return null; }
 
 	@enumerable get ticksToDeploy(): number | undefined {
-		// `+ 1 - 1` like nuke's `timeToLand`: yields `0` on the final invulnerable tick
-		// (`Game.time === #deployTime`) and clamps to `undefined` from the tick after, so reads at
-		// `#deployTime + 1` don't throw before the processor zeroes the field.
 		const deployTime = this['#deployTime'];
-		if (deployTime === 0) {
-			return undefined;
-		}
-		const ticks = RoomObject.requiredExpiryTime(Game, deployTime + 1) - 1;
-		return ticks < 0 ? undefined : ticks;
+		return deployTime === 0 ? undefined : RoomObject.requiredExpiryTime(Game, deployTime + 1) - 1;
 	}
 
 	override get hitsMax(): number {
@@ -127,6 +120,8 @@ export function create(pos: RoomPosition, level: number, deployTime: number) {
 
 // Block invader-core intents whose source was destroyed earlier in the same intent pass,
 // before `#flushObjects` removes the core from the object map.
+// TODO: Intents check is not the place to do this. Also, do other objects in the game have the same
+// condition?
 const checkSourceAlive = (core: StructureInvaderCore) =>
 	core.hits > 0 ? undefined : C.ERR_INVALID_TARGET;
 
--- a/packages/xxscreeps/mods/invader/invader-core.ts
+++ b/packages/xxscreeps/mods/invader/invader-core.ts
@@ -1,8 +1,13 @@
 import type { RoomPosition } from 'xxscreeps/game/position.js';
+import type { Room } from 'xxscreeps/game/room/index.js';
+import { chainIntentChecks, checkSameRoom, checkTarget } from 'xxscreeps/game/checks.js';
 import * as C from 'xxscreeps/game/constants/index.js';
-import { Game, registerGlobal } from 'xxscreeps/game/index.js';
+import { Game, intents, registerGlobal } from 'xxscreeps/game/index.js';
 import * as RoomObject from 'xxscreeps/game/object.js';
-import { OwnedStructure, ownedStructureFormat } from 'xxscreeps/mods/structure/structure.js';
+import { StructureController } from 'xxscreeps/mods/controller/controller.js';
+import { Creep } from 'xxscreeps/mods/creep/creep.js';
+import { StructureTower } from 'xxscreeps/mods/defense/tower.js';
+import { OwnedStructure, checkMyStructure, ownedStructureFormat } from 'xxscreeps/mods/structure/structure.js';
 import { compose, declare, struct, variant, withOverlay } from 'xxscreeps/schema/index.js';
 import { assign } from 'xxscreeps/utility/utility.js';
 
@@ -11,6 +16,8 @@ const shape = struct(ownedStructureFormat, {
 	...variant('invaderCore'),
 	hits: 'int32',
 	level: 'int8',
+	'#actionLog': RoomObject.actionLogFormat,
+	'#collapseTime': 'int32',
 	'#deployTime': 'int32',
 });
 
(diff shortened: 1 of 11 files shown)


packages/xxscreeps/mods/invader/invader-core.ts
```diff
@@ -43,12 +55,59 @@ export class StructureInvaderCore extends withOverlay(OwnedStructure, shape) {
 		return this.ticksToDeploy !== undefined;
 	}
 
+	/**
+	 * Block claim/reservation of the target controller.
+	 * @param target Neutral or invader-reserved controller in the same room.
+	 */
+	reserveController(target: StructureController) {
```
laverdet: These are internal intents only called from the invader npc loop? In that case can you make them `'#private'`

packages/xxscreeps/mods/invader/invader-core.ts
```diff
@@ -43,12 +55,59 @@ export class StructureInvaderCore extends withOverlay(OwnedStructure, shape) {
...
 			return;
 		}
 		super['#applyDamage'](power, type, source);
 	}
+
+	override '#beforeInsert'(room: Room) {
+		super['#beforeInsert'](room);
+		// Keep NPC '2' active in any room that hosts a core; inlined to avoid pulling the
+		// processor entry-point into the runtime sandbox.
+		room['#npcData'].users.add('2');
```
laverdet: Can we just do it when we create a core?

packages/xxscreeps/mods/invader/invader-core.ts
```diff
@@ -61,6 +120,87 @@ export function create(pos: RoomPosition, level: number, deployTime: number) {
 	return core;
 }
 
+// Block invader-core intents whose source was destroyed earlier in the same intent pass,
+// before `#flushObjects` removes the core from the object map.
+const checkSourceAlive = (core: StructureInvaderCore) =>
```
laverdet, line 125: Could you register the intent processor as `after: 'combat'` to sidestep this issue?

packages/xxscreeps/mods/invader/processor.ts
```diff
@@ -75,16 +82,130 @@ registerRoomTickProcessor((room, context) => {
...
+			if (endTime > Game.time + C.CONTROLLER_RESERVE_MAX) {
+				return;
+			}
+			reserve(context, controller, '2', endTime);
+			saveAction(core, 'reserveController', controller.pos);
+			appendEventLog(controller.room, {
+				event: C.EVENT_RESERVE_CONTROLLER,
+				objectId: core.id,
+				amount: power,
+			});
```
laverdet, line 111: Missing context.didUpdate?

packages/xxscreeps/mods/invader/processor.ts
```diff
@@ -75,16 +82,130 @@ registerRoomTickProcessor((room, context) => {
...
+registerObjectTickProcessor(StructureInvaderCore, (core, context) => {
+	const collapseTime = core['#collapseTime'];
+	if (collapseTime > 0 && collapseTime <= Game.time) {
+		// Collapse expiry is a silent removal: no ruin, no EVENT_OBJECT_DESTROYED. Those
+		// only emit from the damage-destroy path, which collapse doesn't traverse. The NPC
+		// reservation is left to run out; the controller's tick processor releases it at expiry.
+		// TODO: Reset an NPC-owned controller here (user/level/effects) once stronghold
+		// deployment can create one.
+		core.room['#removeObject'](core);
+		context.setActive();
```
laverdet: I think context.didUpdate would work?

packages/xxscreeps/mods/invader/processor.ts
```diff
@@ -75,16 +82,130 @@ registerRoomTickProcessor((room, context) => {
...
+		return;
+	}
+
+	flushActionLog(core['#actionLog'], context);
+
+	// Clear the deploy timer the tick after it elapses. `ticksToDeploy`/`effects` read `#deployTime`
+	// through `optionalExpiryTime`, which throws on a past time, so this branches on the raw field
+	// rather than the getter. Zeroing lands in the blob read at `deployTime + 1` — one tick past the
+	// core's last invulnerable tick (`Game.time === deployTime`, `ticksToDeploy === 0`).
+	const deployTime = core['#deployTime'];
```
laverdet, line 199: Can you do the + 1 - 1 trick I did for nuke's timeToLand?

packages/xxscreeps/mods/invader/test.ts
```diff
@@ -262,4 +269,221 @@ describe('Invader core', () => {
...
+			]);
+		});
+	}));
+
+	const findRoomCore = (room: Room) =>
+		room.find(C.FIND_STRUCTURES).find(
+			(structure): structure is StructureInvaderCore => structure.structureType === C.STRUCTURE_INVADER_CORE);
+
+	// A presence creep activates the room for processing; NPC '2' alone is filtered out by
+	// `updateUserRoomRelationships`, so without a human user the room never enters the process queue.
```
laverdet: Is this true? My work in 6873b3ca98beb9f251ee9f96c32a68321fb54acb was meant to fix rooms sleeping on active NPCs. If you load a default world and watch the source keeper rooms you'll see them spawn and die continuously. When the source keeper reaches its destination the NPC loop returns false which sleeps the room until either the creep death wakes it up, or a player enters.

packages/xxscreeps/mods/invader/invader-core.ts
misterwise: Good point.

packages/xxscreeps/mods/invader/invader-core.ts
misterwise: Unfortunately since the intents live with two different users, it wouldn't have the desired effect since initializeIntentConstraints cmp returns 0 for unrelated receivers (index.ts:104-110) and the object remains in the id-map until flushObjects (which is after all intents). checkMyStructure also doesn't check for hits so the intent would get through in the rare case where a core was killed in the same tick as it issued an intent.

packages/xxscreeps/mods/invader/processor.ts
misterwise: In this case it calls reserve() which has a context.didUpdate() in it. We could add it here explicitly to make it clearer, but it isn't necessary in this case.

packages/xxscreeps/mods/invader/test.ts
misterwise: I thought so at first as well, but this didn't hold for the test harness, only the live server. Simulate was missing the main-service boot pass that seeds NPC rooms into the active set so after removing them the tests were failing. I opted to resolve the harness gap, but could pull that out into a separate PR and leave the presence creep instead if you prefer.

PR #252, opened 2026-06-12
powerbank: add power bank structure mod

misterwise:
Adds `mods/powerbank` with `StructurePowerBank`:

- Plain `Structure` (no owner bytes) with `my === false` and an `owner` getter, so banks appear in `FIND_HOSTILE_STRUCTURES` and work as ranged-mass-attack targets with no combat-mod changes.
- `PowerBankStore`: single `power` resource, no capacity, withdraw and transfer both rejected with `ERR_INVALID_TARGET`.
- Damage from creep attackers reflects at `POWER_BANK_HIT_BACK` through `captureDamage`, including on the killing blow, with the `EVENT_ATTACK_TYPE_HIT_BACK` event entry. Non-creep attackers don't reflect — the reflected damage is a dead write upstream; we also skip the event-log entry (commented at the site).
- Natural decay after `POWER_BANK_DECAY` ticks removes the bank silently; violent destruction leaves a ruin carrying the power (`RUIN_DECAY_STRUCTURES.powerBank`).
- Backend renders map key `pb`, the store, and `decayTime`; the runtime global stub is replaced by the real class.
- `create(pos, power)` is the entry point for the spawn-policy PR stacked on this one.

## Validation
- `pnpm run build` + `npx xxscreeps test`: 424 passed, including 5 new PowerBank tests — hostile surface and withdraw/transfer rejection, silent decay (no ruin, no drop), melee and ranged hit-back with event-log entries, killing-blow ruin loot with reflect.
- Live smoke on a local world (~3200 ticks across two server runs): bank injected into an active bot's room survived blob round-trips and server restarts, sat as an obstacle while the room ticked, then decayed silently at its exact `#nextDecayTime` — no ruin, no dropped power, no processor errors or expiry throws.
- Client render verified in the web client: bank draws at its tile with store and hits, plus the `pb` map-view dot.
- `eslint` on the mod: zero new warnings except one `no-unsafe-return` in `backend.ts` from `renderStore`'s untyped return, identical to every existing consumer (container/ruin/tombstone); typing that helper is a separate cleanup.

### DIFF
--- a/packages/xxscreeps/mods/powerbank/powerbank.ts
+++ b/packages/xxscreeps/mods/powerbank/powerbank.ts
@@ -1,0 +1,59 @@
+import type { RoomPosition } from 'xxscreeps/game/position.js';
+import * as C from 'xxscreeps/game/constants/index.js';
+import { Game } from 'xxscreeps/game/index.js';
+import * as RoomObject from 'xxscreeps/game/object.js';
+import { captureDamage } from 'xxscreeps/game/processor.js';
+import { appendEventLog } from 'xxscreeps/game/room/event-log.js';
+import { Creep } from 'xxscreeps/mods/creep/creep.js';
+import { Structure, structureFormat } from 'xxscreeps/mods/structure/structure.js';
+import { compose, declare, struct, variant, withOverlay } from 'xxscreeps/schema/index.js';
+import { assign } from 'xxscreeps/utility/utility.js';
+import { PowerBankStore, powerBankStoreFormat } from './store.js';
+
+export const format = declare('PowerBank', () => compose(shape, StructurePowerBank));
+const shape = struct(structureFormat, {
+	...variant('powerBank'),
+	hits: 'int32',
+	store: powerBankStoreFormat,
+	'#nextDecayTime': 'int32',
+});
+
+export class StructurePowerBank extends withOverlay(Structure, shape) {
+	@enumerable get power() { return this.store[C.RESOURCE_POWER]; }
+	@enumerable get ticksToDecay() { return RoomObject.requiredExpiryTime(Game, this['#nextDecayTime']); }
+	@enumerable get owner() { return { username: 'Power Bank' }; }
+	@enumerable override get my() { return false; }
+	override get hitsMax() { return C.POWER_BANK_HITS; }
+	override get structureType() { return C.STRUCTURE_POWER_BANK; }
+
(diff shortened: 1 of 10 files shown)


packages/xxscreeps/mods/powerbank/powerbank.ts
```diff
@@ -0,0 +1,59 @@
...
+export class StructurePowerBank extends withOverlay(Structure, shape) {
+	@enumerable get power() { return this.store[C.RESOURCE_POWER]; }
+	@enumerable get ticksToDecay() { return RoomObject.requiredExpiryTime(Game, this['#nextDecayTime']); }
+	@enumerable get owner() { return { username: 'Power Bank' }; }
+	@enumerable override get my() { return false; }
+	override get hitsMax() { return C.POWER_BANK_HITS; }
+	override get structureType() { return C.STRUCTURE_POWER_BANK; }
+
+	override '#applyDamage'(power: number, type: number, source?: RoomObject.RoomObject) {
+		if (this.hits <= 0) {
```
laverdet, line 33: Is this needed? I didn't look too closely but it looks like Screeps performs a bulk update on destruction, so another creep attacking a soon-to-die power bank would also receive the attack back.

packages/xxscreeps/mods/powerbank/store.ts
```diff
@@ -0,0 +1,50 @@
+import type { ResourceType } from 'xxscreeps/mods/resource/resource.js';
+import type { BufferView } from 'xxscreeps/schema/index.js';
+import * as C from 'xxscreeps/game/constants/index.js';
+import { Store } from 'xxscreeps/mods/resource/store.js';
+import { compose, struct, withOverlay } from 'xxscreeps/schema/index.js';
+
+const shape = struct({
+	'#power': 'int32',
```
laverdet: Make this `#amount`. I think SingleStore should be refactored into this but I'm also annoyed that we store constant capacities as game state on that object. So that would be a bigger refactor. Anyway if you name it the same thing then upgrading to a common type will carry the amount forward to a new version.

packages/xxscreeps/mods/powerbank/powerbank.ts
misterwise: The in-memory objects are removed synchronously which prevents the second attack and the deferred part is the persistence (db) layer. This stops the second damage back.. but it reveals a potential parity issue around multiple attackers beyond just power banks, since I think i twould still emit an attack log entry. I think that's outside the scope of this though.

PR #255, opened 2026-06-13
client: render effect-icon tooltip via uib-tooltip directive

misterwise:
The effect icon's countdown lives in a native `title`; its text changes every
tick, so at this tick rate the browser keeps re-arming the hover delay and the
tooltip never shows. Rewrite it on serve to `uib-tooltip` (the directive this
client's ui-bootstrap 2.5.0 registers), which opens on mouseenter and stays live.

## Validation

Built `@xxscreeps/client` and ran the backend against a live shard. Hovered an
invader core's deploy effect icon: tooltip shows `EFFECT_INVULNERABILITY — Ticks
remaining: N / 5000`, counting down live, ticking and paused.
Screenshot 2026-06-13 at 4 27 48 PM

### DIFF
--- a/.changeset/client-effect-icon-tooltip.md
+++ b/.changeset/client-effect-icon-tooltip.md
@@ -1,0 +1,5 @@
+---
+"@xxscreeps/client": patch
+---
+
+Render the effect-icon tooltip via uib-tooltip so the live countdown shows.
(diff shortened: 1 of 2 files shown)


laverdet:
By the way Screepers have a bunch of other patches here https://github.com/screepers/steamless-client/tree/main/src/patches. This kind of thing is a rabbit hole.

misterwise:
> By the way Screepers have a bunch of other patches here https://github.com/screepers/steamless-client/tree/main/src/patches. This kind of thing is a rabbit hole.

Just fixing this one thing was frustrating enough to pin down, but I want my effects tooltip! I hope I don't get drawn in.

PR #262, opened 2026-06-18
powerbank: place power banks in highway rooms on a per-room timer

misterwise:
This adds the placement policy for power banks, on top of the structure mod merged in #252.

Every highway room runs an independent respawn timer. The next-due tick is stored on the room itself via `registerStruct('Room', { '#nextPowerBankTime' })` (the `#invaderEnergyTarget` pattern), so it survives a restart; the scratch sorted-set that drives the per-tick sweep is rebuilt from those room fields on shard init. When a room comes due, the shard-tick processor pushes a `placePowerBank` intent that places a bank on a wall position from live terrain and persists the room's next deadline. Keyval accessors live in a sibling `model.ts`.

The placement substrate is reused from deposit placement (`place.ts`, #249), with the simpler per-room policy banks use — no sector geometry, throughput target, or decay coupling. The supporting `isHighwayRoom` predicate lands in `game/room/sector.ts` as the dual of `isCentralRoom`.

Notes for review:
- **Game state stays off the persistent keyval.** Per review, the authoritative deadline lives on the room (not `shard.data`); scratch is a rebuildable index, repopulated from the rooms on init. Init reads `#nextPowerBankTime` from each accessible highway room's blob (object initialization skipped) — a bounded, one-time boot scan, the cost of keeping the deadline authoritative-on-room.
- **Tick-domain schedule** (divergence from vanilla's 5-minute cron poll): a bank appears on its exact `powerBankTime` deadline. The deadline is a game-time value, so scheduling stays tick-domain.
- **Bounded placement search** (divergence from vanilla): position selection caps its retry loop instead of vanilla's unbounded `do/while`, so a wall-less room can't hang the processor.
- A room that has never placed a bank re-rolls its *first* countdown at boot (there is no processing context to persist it before the first placement); once a bank places, the deadline persists on the room and restarts no longer shuffle it.

## Validation
- `npx xxscreeps test` green (full suite, 435 tests). Coverage:
  - `isHighwayRoom`: pins the predicate against the sector modulus, and agrees with `sectorsForRoom` across a swept coordinate range.
  - `PowerBank placement`: bootstrap seeds every highway room without placing; a due highway room places one valid bank (wall position with a non-wall neighbour, in-range power, correct hits/decay) and persists its next-due tick on the room; non-highway rooms are never seeded; a critical roll adds the max-capacity bonus; shard init rebuilds the schedule from a room's persisted deadline.
- `pnpm run build` clean; eslint clean on all changed files.


laverdet:
I'd like to keep game state out of the persistent shard keyval. That's why I pushed back on the very similar Deposit changes, this has the same problem.

You can `registerStruct('Room', { ... })` to store mod information directly on a room (see `'#invaderEnergyTarget'`). Use that to store the next powerbank spawn time. Then on shard initialization you can repopulate the scratch keyval. This wasn't necessary for Deposits because those can be shuffled each time the server starts without correctness violations.

Also please move data accessors to a model.ts

misterwise:
> I'd like to keep game state out of the persistent shard keyval. That's why I pushed back on the very similar Deposit changes, this has the same problem.
> 
> You can `registerStruct('Room', { ... })` to store mod information directly on a room (see `'#invaderEnergyTarget'`). Use that to store the next powerbank spawn time. Then on shard initialization you can repopulate the scratch keyval. This wasn't necessary for Deposits because those can be shuffled each time the server starts without correctness violations.
> 
> Also please move data accessors to a model.ts

I caught the pushback but not the intention, I've made a note of this to avoid repeating it in the future.

PR #263, opened 2026-06-18
Add `manage bot` verbs and register the `manage` subcommand

misterwise:
## Summary

Adds `manage bot` verbs — `bot add` / `bot update` / `bot remove` — alongside the existing
`user` verbs, and registers `manage` as an `xxscreeps` subcommand (previously run via
`node dist/scripts/manage.js`). `bot add` loads a flat directory of built modules (`main.js` CJS,
`main.mjs` ESM, or `main.wasm`/`*.wasm`), saves them under a named user (auto-assigning a random
badge to new users, matching the vanilla server's `genRandomBadge` so fresh bots aren't blank on
the map), and optionally places the bot's first spawn with `--spawn <room> [x,y]`.

Three non-obvious bits:

- **`--spawn` runs under the game mutex.** Instead of queueing a `placeSpawn` intent and waiting
  for the processor next tick, it acquires the `gameMutex` and applies the same claim/spawn
  mutation directly, between ticks — so it no longer requires a running server (a running server
  hands the mutex off cooperatively; with none it's uncontended). It reuses the engine's
  `create`/`claim` helpers and persists both double-buffer slots plus the user/room relationship,
  so the next processed tick picks the spawn up normally. Position validation still happens on a
  throwaway in-memory copy of the room; nothing is persisted until the mutation succeeds.

- **wasm keying matches the player upload format**: a `*.wasm` file is stored under its
  extension-stripped key (`require('foo')` resolves `foo`, not `foo.wasm`), while `main.wasm` keeps
  its name so the runtime's entry detection finds it.

- **The private transform now passes JSON imports through to nodejs**: `badge.ts` imports
  `badge.schema.json` `with { type: 'json' }`, which surfaced a latent loader bug — under
  `--private-transform` the ESM load hook babel-parsed every package module, JSON included, so it
  threw on the schema. The hook now skips `type: 'json'` imports.

Subdirectories are ignored (bots are a flat module map), so `bot add` warns on skipped entries —
point it at flattened build output.

## Validation

- Live-smoked against a running server: `manage bot add … --spawn <room>` acquired the game mutex
  between the server's ticks (observed the contention/handoff), placed the spawn directly, and the
  server processed the newly-claimed room without interruption — controller claimed, spawn present,
  room registered active. Re-validated with the server stopped (uncontended path) and with explicit
  `--spawn <room> x,y`. `bot update` / `bot remove` and the `user` verbs exercised.
- Full test suite green under `--private-transform=isolated-vm` (the CI command), including the
  `Badge.generateRandom` schema test.
- `tsc -b` and `eslint` clean on the changed files.

### DIFF
--- a/.changeset/bot-manage.md
+++ b/.changeset/bot-manage.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Add `manage bot` verbs (add/update/remove, first-spawn for JS/wasm bots) and register the `manage` subcommand.
--- a/README.md
+++ b/README.md
@@ -121,6 +121,29 @@ schema](src/config/config.ts).
 
 If you want to use your bot script, use `npx xxscreeps import --overwrite-code ../your-bot/dist` to replace builtin bots.
 
+### Adding bots
+
+`import --overwrite-code` replaces the built-in bots. To add a *named* bot — and optionally place
+its first spawn — without touching the built-ins, use the `manage` subcommand:
+
+```
+npx xxscreeps manage bot add <name> <dir> --spawn <room>
+```
+
+`<dir>` is a flat directory of a bot's built modules — `main.js` (CommonJS), `main.mjs` (ESM), or a
+`main.wasm`/`*.wasm` binary. Subdirectories are ignored, so point it at the build output (often `src`
+or `dist`), not the source tree of a bot that still needs compiling. `--spawn` also places the bot's
+first spawn, whether or not a server is running; drop it to just save the code. Open-source bots
+published for vanilla servers work as-is:
+
+```
+npm install screeps-bot-tooangel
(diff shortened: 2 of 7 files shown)


laverdet:
If you acquire the `gameMutex` you could run the updates directly in between game ticks without involving the processor. Then you wouldn't need the invariants around the server being started or not. It's not a well-tested API though so if it doesn't work out let me know and we can push this as-is.

misterwise:
> If you acquire the `gameMutex` you could run the updates directly in between game ticks without involving the processor. Then you wouldn't need the invariants around the server being started or not. It's not a well-tested API though so if it doesn't work out let me know and we can push this as-is.

Because it's acting on sleeping rooms, it works, at least in my tests. I think it's worth proceeding with and we can see if it surfaces any issues once it gets some use down the line.

PR #274, opened 2026-06-25
Add StructureInvaderCore defender spawning

misterwise:
Continues the invader-core arc (after #215 and #251's NPC actions): gives
`StructureInvaderCore` the `createCreep` action and a spawning lifecycle. The
core incubates a defender on its own tile and materializes it on an open
adjacent tile once `INVADER_CORE_CREEP_SPAWN_TIME[level] × body.length` ticks
elapse.

`StructureSpawn` and the invader core finish a spawn the same way — find an
open adjacent tile (preferring requested directions, stomping a lone blocker,
else retrying next tick). That completion logic moves out of the spawn
processor into a shared `hatchSpawningCreep`; the two callers differ only in
how the hatched creep is aged. The core also reuses `StructureSpawn.Spawning`
for the record, so `invader-core` now depends on `mods/spawn`.

As with #251, `createCreep` lands ahead of its driver: the stronghold
population behaviors that select bodies/boosts, boost handling, and the
collapse-tied creep lifetime are later slices.

## Validation

- `mods/invader/test.ts` — added `Invader core spawning` (5 tests): records a
  spawning defender on the core tile; materializes it adjacent on completion;
  holds the spawn pending while every neighbour is blocked, then hatches when a
  tile frees; rejects a second `createCreep` while busy (`ERR_BUSY`); rejects a
  non-spawning level (`ERR_INVALID_TARGET`).
- Exercised end-to-end on a running multi-service server: the queued
  `createCreep` persisted `spawning` onto the room blob and the defender
  hatched onto an adjacent tile across a real save/load cycle.

### DIFF
--- a/packages/xxscreeps/mods/invader/invader-core.ts
+++ b/packages/xxscreeps/mods/invader/invader-core.ts
@@ -1,4 +1,5 @@
 import type { RoomPosition } from 'xxscreeps/game/position.js';
+import type { PartType } from 'xxscreeps/mods/creep/creep.js';
 import { chainIntentChecks, checkSameRoom, checkTarget } from 'xxscreeps/game/checks.js';
 import * as C from 'xxscreeps/game/constants/index.js';
 import { Game, intents, registerGlobal } from 'xxscreeps/game/index.js';
@@ -6,8 +7,9 @@ import * as RoomObject from 'xxscreeps/game/object.js';
 import { StructureController } from 'xxscreeps/mods/controller/controller.js';
 import { Creep } from 'xxscreeps/mods/creep/creep.js';
 import { StructureTower } from 'xxscreeps/mods/defense/tower.js';
+import { Spawning, spawningFormat } from 'xxscreeps/mods/spawn/spawn.js';
 import { OwnedStructure, checkMyStructure, ownedStructureFormat } from 'xxscreeps/mods/structure/structure.js';
-import { compose, declare, struct, variant, withOverlay } from 'xxscreeps/schema/index.js';
+import { compose, declare, optional, struct, variant, withOverlay } from 'xxscreeps/schema/index.js';
 import { assign } from 'xxscreeps/utility/utility.js';
 
 export const format = declare('InvaderCore', () => compose(shape, StructureInvaderCore));
@@ -15,6 +17,7 @@ const shape = struct(ownedStructureFormat, {
 	...variant('invaderCore'),
 	hits: 'int32',
 	level: 'int8',
+	spawning: optional(compose(spawningFormat, Spawning), null),
 	'#actionLog': RoomObject.actionLogFormat,
 	'#collapseTime': 'int32',
 	'#deployTime': 'int32',
@@ -35,10 +38,6 @@ export class StructureInvaderCore extends withOverlay(OwnedStructure, shape) {
 		];
 		return effects.length === 0 ? undefined : effects;
 	}
-
-	// TODO: stronghold spawning
-	// eslint-disable-next-line @typescript-eslint/class-literal-property-style
-	@enumerable get spawning(): null { return null; }
 
 	@enumerable get ticksToDeploy(): number | undefined {
 		const deployTime = this['#deployTime'];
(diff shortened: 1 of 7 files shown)


packages/xxscreeps/mods/invader/invader-core.ts
```diff
@@ -184,6 +195,26 @@ export function checkUpgradeController(core: StructureInvaderCore, target: Struc
...
+	return chainIntentChecks(
+		() => checkMyStructure(core, StructureInvaderCore),
+		() => checkSourceAlive(core),
+		() => checkString(name, 100, true),
+		() => core.spawning ? C.ERR_BUSY : C.OK,
+		// Only levels with a configured spawn time (2+) field defenders.
+		() => (C.INVADER_CORE_CREEP_SPAWN_TIME[core.level] ?? 0) > 0 ? undefined : C.ERR_INVALID_TARGET,
+		() => {
+			if (
+				!Array.isArray(body) ||
```
laverdet: These checks are too aggressive against internal intents

packages/xxscreeps/mods/invader/processor.ts
```diff
@@ -205,6 +224,17 @@ registerObjectTickProcessor(StructureInvaderCore, (core, context) => {
 			context.wakeAt(deployTime + 1);
 		}
 	}
+
+	// Advance an in-progress defender spawn. Read the raw spawn time rather than `remainingTime`
+	// (which throws once overdue) so a coalesced/late wake in an unobserved room still hatches.
```
laverdet: Late wakes are an invariant. If it's good enough for the StructureSpawn processor then it's good enough for this one.

packages/xxscreeps/mods/invader/test.ts
```diff
@@ -472,3 +474,105 @@ describe('Invader core', () => {
...
 });
+
+describe('Invader core spawning', () => {
+	const corePos = new RoomPosition(25, 25, 'W1N1');
+	const findRoomCore = (room: Room) => lookForStructures(room, C.STRUCTURE_INVADER_CORE)[0]!;
+	const findCreep = (room: Room, name: string) => room.find(C.FIND_CREEPS).find(creep => creep.name === name);
+	// Level 5 spawns one tick per body part (`INVADER_CORE_CREEP_SPAWN_TIME[5] === 1`), so a 2-part
+	// body needs 2 ticks — the shortest countdown the mechanism still exercises.
+	const body = [ C.MOVE, C.ATTACK ];
+	const allNeighbors = [ -1, 0, 1 ].flatMap(dx => [ -1, 0, 1 ].map(dy => [ dx, dy ] as const)).filter(([ dx, dy ]) => dx !== 0 || dy !== 0);
```
laverdet: `iterateNeighbors`

packages/xxscreeps/mods/spawn/processor.ts
```diff
@@ -192,64 +193,81 @@ const intents = [
...
-					if (!Fn.some(otherDirections, dir => check(getPositionInDirection(creep.pos, dir)))) {
-						spawnPos = hostileCreep.pos;
-						buryCreep(hostileCreep);
-					}
-				}
+// Complete a pending spawn whose timer has elapsed: place the incubating creep on the first open
+// adjacent tile (preferring `spawning.directions`, stomping a lone blocking hostile as a last
+// resort), stamp its real age, and clear the record. A fully blocked owner retries next tick. Shared
+// by `StructureSpawn` and `StructureInvaderCore`, which differ only in how the spawned creep ages.
+export function hatchSpawningCreep(
```
laverdet: Nitpicky but can you rename to `birthSpawnCreep` since they are born / spawned not hatched.

packages/xxscreeps/mods/spawn/processor.ts
```diff
@@ -192,64 +193,81 @@ const intents = [
...
+			}
+		}
 
-				// Place the creep
-				const hasClaim = creep.body.some(part => part.type === C.CLAIM);
-				creep['#ageTime'] = Game.time + (hasClaim ? C.CREEP_CLAIM_LIFE_TIME : C.CREEP_LIFE_TIME) - 1;
-				creep.room['#moveObject'](creep, spawnPos);
+		// All preferred directions blocked — only stomp if non-preferred are also blocked
+		if (!spawnPos && hostileCreep) {
+			const otherDirections = Fn.reject(ALL_DIRECTIONS, dir => directions.has(dir));
```
laverdet: There will be a merge conflict here with the directional changes I made recently

packages/xxscreeps/mods/invader/invader-core.ts
misterwise: Fair, got into the habit of guarding intents against malicious calls and neglected the fact that we are the caller.

packages/xxscreeps/mods/spawn/processor.ts
misterwise: folded your getPositionInDirection-returns-undefined guard (if (!pos) continue) and the Fn.pipe(… Fn.filter …) stomp into the shared birthSpawnCreep, so both callers pick up the directional changes.

PR #275, opened 2026-06-25
Add power creep account layer

misterwise:
Adds the account-scoped power creep layer as a new `mods/powercreep` (off `powerspawn`), serving the six `/api/game/power-creeps/*` routes the client drives the garage with — `list`/`create`/`upgrade`/`rename`/`delete`/`cancel-delete` — and materializing `Game.powerCreeps` (including unspawned creeps) through the runner payload, replacing the old `Game.powerCreeps`/`PowerCreep` stubs.

Account ops are backend routes, not intents: there's no global/non-room intent substrate to push them through, so they're pure user-keyspace mutations with GPL gating and wall-clock delete cooldowns checked on read. No room presence yet — spawn/renew and the room verbs are the next slice. `promo`/`experimentation` (power-era onboarding + respec) are deferred; they're button-only, never on the garage load path.

Also surfaces the account's `power` in `/api/auth/me` (via `powerspawn`'s `sendUserInfo`) so the client shows the Global Power Level — without it the garage reads GPL 0. The roster stores each creep as a JSON-encoded hash value (dense; the nested `powers` map makes per-field normalization awkward).

## Validation

- 9 `PowerCreep account` harness tests (GPL gating, reachability, budget/downgrade rejects, rename, delete round-trip, runtime view); full suite 437/437, eslint clean on changed files.
- Live against the official Steam/web client (headless Chrome + CDP): garage loads at `#!/overview/power`, full create→upgrade(`GENERATE_OPS`)→rename→delete→cancel-delete round-trip through the real routes, creep renders, and the GPL tool shows "2/2" for a GPL-2 account.


packages/xxscreeps/mods/powercreep/backend.ts
```diff
@@ -0,0 +1,71 @@
...
+import * as Model from './model.js';
+
+// The official client drives the power-creep screen through these routes (`/api/game/power-creeps/*`).
+// Each mutates account keyspace directly — there is no global-intent stage and no room presence yet.
+// The model is the sole validator; routes are glue that maps a thrown rejection to `{ error }` and a
+// success to `{ ok: 1 }`, the envelope the client expects.
+type RouteContext = RouterContext<State, Context>;
+
+interface Body {
+	name?: unknown;
```
laverdet: Take a look at `makeValidatedPayloadRoute` which is a pattern for validating frontend requests.

packages/xxscreeps/mods/powercreep/driver.ts
```diff
@@ -0,0 +1,18 @@
...
+	interface TickPayload {
+		powerCreeps: PowerCreepRecord[];
+	}
+}
+
+hooks.register('runnerConnector', player => {
+	const { shard, userId } = player;
+	return [ undefined, {
+		async refresh(payload) {
+			payload.powerCreeps = await loadRoster(shard.db, userId);
```
laverdet: There needs to be a channel here which listens for updates and only pushes a new payload to the runtime when these are mutated.

laverdet (changes requested):
For the power creeps global database records they should be RoomObject blobs. Since they are not spawned, for the position you can use `RoomPosition(0, 0, 'E0S0')` which is signed `[0,0,0,0]`. Then when it's time to spawn them it's just a matter of copying them into a room. This also means you won't have two different object formats. Also this follows the idea that the runner machinery should generally not be handling parsed object structures. We prefer to pass around (Shared)ArrayBuffer instances to instantly transfer into the runtime.

For the mutation operations you can use keyval compare and swap (if: EQ) to avoid race conditions entirely.

packages/xxscreeps/mods/powercreep/model.ts
```diff
@@ -0,0 +1,234 @@
...
+// Apply `fn` to the live roster and commit it with a compare-and-swap, retrying if a concurrent
+// mutation slipped in between the read and the write. `fn` validates and throws on rejection, in
+// which case nothing is written.
+async function mutate<Type>(db: Database, userId: string, fn: (roster: PowerCreep[]) => Type | Promise<Type>) {
+	const key = powerCreepsKey(userId);
+	for (let attempt = 0; ; ++attempt) {
+		const raw = await db.data.get(key);
+		const roster = parseRoster(raw).filter(isLive);
+		const result = await fn(roster);
+		const stored = await db.data.set(key, typedArrayToString(write(roster)), {
```
laverdet: These should be read with `.get({blob:true})`. Reading as a string and converting into a blob defeats the purpose. `deleteTime` could live on the blob. You will need to add CAS support to blob `set`

packages/xxscreeps/mods/powercreep/model.ts
```diff
@@ -0,0 +1,234 @@
...
+	return Math.floor((power / C.POWER_LEVEL_MULTIPLY) ** (1 / C.POWER_LEVEL_POW));
+}
+
+function freeLevels(power: number, roster: PowerCreep[]) {
+	const used = roster.length + Fn.accumulate(roster, creep => creep.level);
+	return gplLevel(power) - used;
+}
+
+function userPower(db: Database, userId: string) {
+	return db.data.hGet(User.infoKey(userId), 'power').then(power => Number(power) || 0);
```
laverdet: Don't use `then` unless you have a very good reason

packages/xxscreeps/mods/powercreep/powercreep.ts
```diff
@@ -0,0 +1,74 @@
...
+import { compose, declare, optional, struct, vector, withOverlay } from 'xxscreeps/schema/index.js';
+import { instantiate } from 'xxscreeps/utility/utility.js';
+
+// A power creep is a `RoomObject` whether it is sitting in the account roster or spawned into a room,
+// so it gets a single serialized format. Unspawned creeps live at `RoomPosition(0, 0, 'E0S0')` (the
+// all-zero signed position); spawning is then just a matter of copying the object into a room.
+export const format = declare('PowerCreep', () => compose(shape, PowerCreep));
+const shape = struct(baseFormat, {
+	name: 'string',
+	className: 'string',
```
laverdet: Should be an enumeration

packages/xxscreeps/mods/powercreep/powercreep.ts
```diff
@@ -0,0 +1,74 @@
...
+	// Flat `{ [PWR]: level }`; the public `powers` getter expands it to `{ [PWR]: { level } }`.
+	'#powers': compose(vector(struct({ power: 'int8', level: 'int8' })), {
```
laverdet: This is a really lousy compose format. You go through all this trouble to convert it into this internal format and then on the other you constantly have to invoke `Object.values` and `Object.keys` to consume it. As a general rule: JavaScript object literals are not a good choice for dynamic keys.

Anyway I would recommend just leaving this directly as a vector.

packages/xxscreeps/mods/powercreep/powercreep.ts
```diff
@@ -0,0 +1,74 @@
...
+		compose: (powers: { power: number; level: number }[]) =>
+			Fn.fromEntries(powers, ({ power, level }) => [ power, level ]),
+		decompose: (powers: Record<number, number>) =>
+			Object.entries(powers).map(([ power, level ]) => ({ power: Number(power), level })),
+	}),
+	// Wall-clock ms; `spawnCooldownTime` is set when a spawned creep dies (slice 4), `0` while idle.
+	spawnCooldownTime: 'double',
+	// Wall-clock ms; deletion scheduled this far out, cancellable until it elapses.
+	deleteTime: optional('double'),
```
laverdet: Just do `double` and 0 means it won't be deleted.

misterwise:
Round 2 pushed — all five inline comments addressed (two commits on top of the reviewed tip).

- **model.ts:53 (blob + CAS)** — `loadRoster`, `loadPowerCreepsBlob`, and `mutate`
  now read/write the raw blob via `.get({ blob: true })` and `set(…, Uint8Array)`;
  the latin1 string round-trip is gone and `deleteTime` stays on the blob. Adding
  `if` support to blob `set` was the bulk of it and touched two layers worth flagging
  (it's the separate first commit):
  - `@xxscreeps/redis`: `keyval.ts` maps `EQ`/`NE` to `SET … IFEQ/IFNE`, and a
    `Uint8Array` `matchValue` runs through `recv()` so it encodes to the same bytes
    as the stored value. Server-atomic — nothing else needed.
  - local provider: `conditionalSet` reads with `get` (which yields on a disk read)
    before storing, and the in-process responder invokes methods directly with no
    serialization — so two concurrent swaps would both read the prior and both commit
    (lost write). Fixed with a per-`BlobStorage` promise chain that serializes
    *conditional* writes only; the hot get/unconditional-set paths are untouched.
- **powercreep.ts:18** — `#powers` is a plain `vector(struct({ power, level }))` now;
  the compose round-trip and the `Object.keys`/`Object.values` consumers are gone.
- **powercreep.ts:16** — `className` is `enumerated(...Object.values(C.POWER_CLASS))`.
- **powercreep.ts:27** — `deleteTime` is `double`, `0` = not scheduled; threaded
  through `isLive` / schedule / cancel / render.
- **model.ts:76** — `userPower` is `async`/`await`.

The blob-CAS capability stands on its own, so it's a separate commit + changeset
(bumps `@xxscreeps/redis` too) — happy to fold it into the feature changeset if you'd
prefer. New tests cover the CAS conditions and a concurrent-create race; suite green.

laverdet:
In think you'll have to refactor these model checks into intent checks, since they can be invoked from the player runtime. What you have here is kind of off-meta, I think because you got caught up with the backend route validation which ended up being moot (for example all the typeof checks). Compare to the spawn placement backend which just invokes `checkCreateConstructionSite`.

Also I have no idea what Claude is talking about when it says "slice 4". A lot of these comments are just indecipherable technobabble.

PR #277, opened 2026-06-25
Overwrite existing branch on clone-branch instead of erroring

misterwise:
Caveat, it may be an intentional divergence that xxscreeps doesn't clobber an existing branch on clone, if so, I will update #1 accordingly.

`POST /api/user/clone-branch` returned 500 (`Branch already exists`) when `newName`
matched an existing branch. The official `@screeps/backend` route upserts instead
and skips the branch-count check when overwriting. Dropped the guard and gated
`Too many branches` on the name not already existing — the `copy`/`saveContent`
paths already replace the target (the redis provider passes `REPLACE`). Overwriting
a branch can now hit the *active* branch, so the handler also publishes a code-channel
`update` (like `saveContent`) to mark a running runner stale.

## Validation

Ran the redis-backed backend live: register → save `main`=v1 → clone→`alpha` →
save `main`=v2 → clone→`alpha` again. Pre-fix the second clone 500'd and `alpha`
kept v1; post-fix it returns 200 and `GET code?branch=alpha` shows v2. eslint on
the changed file: no new warnings.

Refs #1.

### DIFF
--- a/.changeset/clone-branch-overwrite.md
+++ b/.changeset/clone-branch-overwrite.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Fix clone-branch to overwrite an existing branch instead of erroring.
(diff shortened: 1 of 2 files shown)


laverdet:
Looking at the Screeps code it seems like they throw on max branches, even when clobbering an existing branch. It just looks to be a buggy endpoint on their end. I don't think we need to copy this exact behavior unless there's a greater compatibility issue with some tool.

PR #281, opened 2026-06-26
market: record terminal transfers in Game.market transactions

misterwise:
`Game.market.incomingTransactions` / `outgoingTransactions` were stubbed to `[]`, though `StructureTerminal.send`'s docs already promise both parties can track transfers. Each `send` now records a transfer both parties can read back.

A transfer is stored once as an immutable schema blob in its own namespace (`market/transaction/<id>`); each party's per-direction list holds references (ids), and the runner caches blobs by transaction id so both parties' runtimes are handed the same `SharedArrayBuffer`. The blob keeps the parties as user ids (resolved to usernames at read time) and the description raw (HTML-escaped by its getter, the way `creep.saying` is hidden). Each list returns the last 24 hours or 100 transfers, whichever is fewer; a reference count frees a blob once both parties have evicted it. Deal transactions wait for the order book.

Reworked from the first pass per review: schema blobs in their own namespace referenced by id, shared unparsed buffers cached by transaction id, getter-side description hiding and escaping, user-id (not username) storage, no escaped HTML in the database, and a 24h/100 bound instead of a permanent 100.

## Validation
- `npx xxscreeps test --private-transform=isolated-vm` — the engine suite passes under the CI transform. New `Market transactions` cases assert the shared blob (raw description + stored user ids), the 24h read window, and a self-send round-tripping the processor → driver → runtime read path with resolved usernames and an escaped description; the existing `Market › send` cooldown cases still hold.
- Rebased onto `main` to pick up #280's terminal send cooldown.

### DIFF
--- a/packages/xxscreeps/mods/market/driver.ts
+++ b/packages/xxscreeps/mods/market/driver.ts
@@ -1,6 +1,7 @@
 import type { TransactionPayload } from './transaction.js';
 import type { Shard } from 'xxscreeps/engine/db/index.js';
 import { hooks } from 'xxscreeps/engine/runner/index.js';
+import { getOrSet } from 'xxscreeps/utility/utility.js';
 import { getTransactionChannel, kTransactionWindow, loadTransactionBlob, loadTransactionEntries, selectRecent } from './model.js';
 import { read } from './transaction.js';
 
@@ -12,18 +13,13 @@ declare module 'xxscreeps/engine/runner/index.js' {
 
 // Transaction blobs are immutable and their ids are globally unique, so cache them by id: a transfer
 // referenced by both parties' runtimes is read once and handed out as the same SharedArrayBuffer.
-const blobCache = new Map<string, Promise<Readonly<Uint8Array>>>();
-function loadBlob(shard: Shard, id: string) {
-	let blob = blobCache.get(id);
-	if (blob === undefined) {
-		blob = loadTransactionBlob(shard, id);
-		blobCache.set(id, blob);
-	}
-	return blob;
-}
+const loadBlob = function() {
+	const cache = new Map<string, Promise<Readonly<Uint8Array>>>();
+	return (shard: Shard, id: string) =>
+		getOrSet(cache, id, () => loadTransactionBlob(shard, id));
+}();
 
 hooks.register('runnerConnector', async player => {
-	// Reload the user's transactions only when the processor signals a new transfer, not every tick.
 	const channel = await getTransactionChannel(player.shard, player.userId).subscribe();
 	let dirty = true;
 	channel.listen(() => { dirty = true; });
--- a/packages/xxscreeps/mods/market/driver.ts
+++ b/packages/xxscreeps/mods/market/driver.ts
@@ -20,31 +20,31 @@ const loadBlob = function() {
 }();
 
 hooks.register('runnerConnector', async player => {
+	// The processor publishes here whenever a transfer touches the user. Reload and resend only then;
+	// on quiet ticks the runtime keeps the last payload it was sent, so aged-out transfers linger
+	// until the next transfer rewindows the list.
 	const channel = await getTransactionChannel(player.shard, player.userId).subscribe();
 	let dirty = true;
 	channel.listen(() => { dirty = true; });
-	let entries: ReturnType<typeof loadTransactionEntries> | undefined;
 	return [ () => channel.disconnect(), {
+		// A fresh sandbox (first tick, or after a code reset) holds no transactions; resend next tick.
+		initialize() { dirty = true; },
 		async refresh(payload) {
-			if (dirty) {
-				dirty = false;
-				entries = loadTransactionEntries(player.shard, player.userId);
+			if (!dirty) {
+				return;
 			}
-			const { incoming, outgoing } = await entries!;
-			// Re-apply the window every tick so transfers age out without a reload.
+			dirty = false;
+			const { incoming, outgoing } = await loadTransactionEntries(player.shard, player.userId);
 			const cutoff = Date.now() - kTransactionWindow;
 			const incomingIds = selectRecent(incoming, cutoff);
 			const outgoingIds = selectRecent(outgoing, cutoff);
 			const ids = [ ...new Set([ ...incomingIds, ...outgoingIds ]) ];
 			const blobList = await Promise.all(ids.map(id => loadBlob(player.shard, id)));
-			// Ask the runner to resolve the parties' usernames for the read-time getters.
-			const parties = new Set<string>();
+			// Contribute both parties of every transfer; the runner resolves and dedupes the ids.
 			for (const blob of blobList) {
 				const transaction = read(blob);
-				parties.add(transaction['#sender']);
-				parties.add(transaction['#recipient']);
+				(payload.userIds ??= []).push(transaction['#sender'], transaction['#recipient']);
 			}
-			(payload.userIds ??= []).push(...parties);
 			const blobs = Object.fromEntries(ids.map((id, index) => [ id, blobList[index]! ]));
 			payload.transactions = { incoming: incomingIds, outgoing: outgoingIds, blobs };
 		},
--- a/packages/xxscreeps/mods/market/driver.ts
+++ b/packages/xxscreeps/mods/market/driver.ts
@@ -1,8 +1,7 @@
 import type { TransactionPayload } from './transaction.js';
 import type { Shard } from 'xxscreeps/engine/db/index.js';
-import * as User from 'xxscreeps/engine/db/user/index.js';
 import { hooks } from 'xxscreeps/engine/runner/index.js';
-import { loadTransactionBlob, loadTransactionRefs } from './model.js';
+import { getTransactionChannel, kTransactionWindow, loadTransactionBlob, loadTransactionEntries, selectRecent } from './model.js';
 import { read } from './transaction.js';
 
 declare module 'xxscreeps/engine/runner/index.js' {
@@ -11,43 +10,47 @@ declare module 'xxscreeps/engine/runner/index.js' {
 	}
 }
 
-interface BlobCacheEntry {
-	time: number;
-	blobs: Map<string, Promise<Readonly<Uint8Array>>>;
-}
-
-// Transaction blobs are immutable, so cache them by id per (shard, tick): a transfer referenced by
-// both parties' runtimes is read once and handed out as the same SharedArrayBuffer.
-const blobCache = new WeakMap<Shard, BlobCacheEntry>();
-function loadBlobOnce(shard: Shard, time: number, id: string) {
-	let cache = blobCache.get(shard);
-	if (cache?.time !== time) {
-		cache = { time, blobs: new Map() };
-		blobCache.set(shard, cache);
-	}
-	let blob = cache.blobs.get(id);
+// Transaction blobs are immutable and their ids are globally unique, so cache them by id: a transfer
+// referenced by both parties' runtimes is read once and handed out as the same SharedArrayBuffer.
+const blobCache = new Map<string, Promise<Readonly<Uint8Array>>>();
+function loadBlob(shard: Shard, id: string) {
+	let blob = blobCache.get(id);
 	if (blob === undefined) {
 		blob = loadTransactionBlob(shard, id);
-		cache.blobs.set(id, blob);
+		blobCache.set(id, blob);
 	}
 	return blob;
 }
 
-hooks.register('runnerConnector', player => [ undefined, {
-	async refresh(payload) {
-		const { incoming, outgoing } = await loadTransactionRefs(player.shard, player.userId);
-		const ids = [ ...new Set([ ...incoming, ...outgoing ]) ];
-		const blobList = await Promise.all(ids.map(id => loadBlobOnce(player.shard, payload.time, id)));
-		// Resolve every party so the read-time `sender` / `recipient` getters find a username.
-		const parties = new Set<string>();
-		for (const blob of blobList) {
-			const transaction = read(blob);
-			parties.add(transaction['#sender']);
-			parties.add(transaction['#recipient']);
-		}
-		const usernames = Object.fromEntries(await Promise.all([ ...parties ].map(async userId =>
-			[ userId, (await player.shard.db.data.hGet(User.infoKey(userId), 'username'))! ] as const)));
-		const blobs = Object.fromEntries(ids.map((id, index) => [ id, blobList[index]! ]));
-		payload.transactions = { incoming, outgoing, blobs, usernames };
-	},
-} ]);
+hooks.register('runnerConnector', async player => {
+	// Reload the user's transactions only when the processor signals a new transfer, not every tick.
+	const channel = await getTransactionChannel(player.shard, player.userId).subscribe();
+	let dirty = true;
+	channel.listen(() => { dirty = true; });
+	let entries: ReturnType<typeof loadTransactionEntries> | undefined;
+	return [ () => channel.disconnect(), {
+		async refresh(payload) {
+			if (dirty) {
+				dirty = false;
+				entries = loadTransactionEntries(player.shard, player.userId);
+			}
+			const { incoming, outgoing } = await entries!;
+			// Re-apply the window every tick so transfers age out without a reload.
+			const cutoff = Date.now() - kTransactionWindow;
+			const incomingIds = selectRecent(incoming, cutoff);
+			const outgoingIds = selectRecent(outgoing, cutoff);
+			const ids = [ ...new Set([ ...incomingIds, ...outgoingIds ]) ];
+			const blobList = await Promise.all(ids.map(id => loadBlob(player.shard, id)));
+			// Ask the runner to resolve the parties' usernames for the read-time getters.
+			const parties = new Set<string>();
+			for (const blob of blobList) {
+				const transaction = read(blob);
+				parties.add(transaction['#sender']);
+				parties.add(transaction['#recipient']);
+			}
+			(payload.userIds ??= []).push(...parties);
+			const blobs = Object.fromEntries(ids.map((id, index) => [ id, blobList[index]! ]));
+			payload.transactions = { incoming: incomingIds, outgoing: outgoingIds, blobs };
+		},
+	} ];
+});
--- a/packages/xxscreeps/mods/market/driver.ts
+++ b/packages/xxscreeps/mods/market/driver.ts
@@ -11,9 +11,14 @@ declare module 'xxscreeps/engine/runner/index.js' {
 	}
 }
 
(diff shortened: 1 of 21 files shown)


packages/xxscreeps/mods/market/driver.ts
```diff
@@ -0,0 +1,15 @@
...
+
+declare module 'xxscreeps/engine/runner/index.js' {
+	interface TickPayload {
+		transactions?: Transactions;
+	}
+}
+
+hooks.register('runnerConnector', player => [ undefined, {
+	async refresh(payload) {
+		payload.transactions = await loadTransactions(player.shard, player.userId);
```
laverdet: Should not refresh this payload every tick. It needs a channel.

packages/xxscreeps/mods/market/model.ts
```diff
@@ -0,0 +1,83 @@
...
+export function loadUsername(shard: Shard, userId: string) {
+	return shard.db.data.hGet(User.infoKey(userId), 'username');
+}
+
+async function pushTransaction(shard: Shard, userId: string, direction: Direction, payload: string) {
+	const key = transactionsKey(userId, direction);
+	const length = await shard.data.rPush(key, [ payload ]);
+	// Trim oldest entries (list head) so storage stays bounded near the read limit.
+	for (let excess = length - kMaxTransactions; excess > 0; --excess) {
+		await shard.data.lPop(key);
```
laverdet: `lTrim`. Claude probably refused to do that because a wrapper doesn't exist but that's a bad call.

packages/xxscreeps/mods/market/model.ts
```diff
@@ -0,0 +1,83 @@
...
+		...stored,
+		sender: stored.sender == null ? undefined : { username: stored.sender },
+		recipient: stored.recipient == null ? undefined : { username: stored.recipient },
+	};
+}
+
+async function loadDirection(shard: Shard, userId: string, direction: Direction) {
+	// `lRange` returns oldest-first; the API lists transactions newest-first.
+	const entries = await shard.data.lRange(transactionsKey(userId, direction), -kMaxTransactions, -1);
+	return entries.reverse().map(entry => view(JSON.parse(entry) as StoredTransaction));
```
laverdet: The `reverse()` call can be fixed by correct use of rpush vs lpush?

packages/xxscreeps/mods/market/processor.ts
```diff
@@ -18,27 +20,48 @@ const intents = [
...
+							transactionId: Id.generateId(),
+							// The processor runs one tick ahead of the code that issued the send.
+							time: Game.time - 1,
+							sender: senderName ?? undefined,
+							recipient: recipientName ?? undefined,
+							resourceType,
+							amount: sent,
+							from: terminal.room.name,
+							to: destination,
+							...description == null ? undefined : { description: description.replace(/</g, '&lt;') },
```
laverdet: Escaped data does not belong in the database. This would be a getter on the object itself if the user's sandbox is not safe.

laverdet (commented):
- Archived orders should be schema blobs. They also should not be stored directly in lists like these, normalize them into their own namespace and the player-scoped lists will reference the normalized structure
- Share unparsed payloads globally. The runner should cache by transaction id and give the same sharedarraybuffer to multiple runtimes.
- `description` can be hidden from within the runtime in the same way creep.saying is
- Don't store usernames in the database in this manner, store userids
- Don't store escaped HTML data in the database.  This can be escaped in a getter.
- I want to deviate from Screeps here on the number of transactions returned. It's bad that participating in the market permanently costs your runtime ~50kb of memory overhead. Instead it should just be the last 24 hours or 100 whichever is less.

misterwise:
Pushed a small follow-up (36f7393f) on top of the review revision — no change to the documented API:

- **Transaction list eviction:** `pushReference` now pops a single oldest entry per push instead of trimming to `length - kMaxTransactions`. The trim over-evicted when two same-direction records for one user raced — each saw a post-append length over the cap and trimmed independently. One pop per push keeps each list at the cap regardless of interleaving, and the refcounts stay balanced (one `dropReference` per evicted entry).
- **Surface tidy-up:** demoted `format` and `kMaxTransactions` to file-private (no external consumer) and named the driver's blob-cache entry type.

Full suite green (447 passed).

packages/xxscreeps/mods/market/driver.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+interface BlobCacheEntry {
+	time: number;
+	blobs: Map<string, Promise<Readonly<Uint8Array>>>;
+}
+
+// Transaction blobs are immutable, so cache them by id per (shard, tick): a transfer referenced by
+// both parties' runtimes is read once and handed out as the same SharedArrayBuffer.
+const blobCache = new WeakMap<Shard, BlobCacheEntry>();
```
laverdet: Why are you keying on time & shard? Transactions last forever and time doesn't have any bearing on their validity. There is also no reason to further scope on shard.

packages/xxscreeps/mods/market/driver.ts
```diff
@@ -0,0 +1,53 @@
...
+	if (blob === undefined) {
+		blob = loadTransactionBlob(shard, id);
+		cache.blobs.set(id, blob);
+	}
+	return blob;
+}
+
+hooks.register('runnerConnector', player => [ undefined, {
+	async refresh(payload) {
+		const { incoming, outgoing } = await loadTransactionRefs(player.shard, player.userId);
```
laverdet: Should not refresh this payload every tick. It needs a channel.

packages/xxscreeps/mods/market/driver.ts
```diff
@@ -0,0 +1,53 @@
...
+		const ids = [ ...new Set([ ...incoming, ...outgoing ]) ];
+		const blobList = await Promise.all(ids.map(id => loadBlobOnce(player.shard, payload.time, id)));
+		// Resolve every party so the read-time `sender` / `recipient` getters find a username.
+		const parties = new Set<string>();
+		for (const blob of blobList) {
+			const transaction = read(blob);
+			parties.add(transaction['#sender']);
+			parties.add(transaction['#recipient']);
+		}
+		const usernames = Object.fromEntries(await Promise.all([ ...parties ].map(async userId =>
```
laverdet: Integrate into the "// Load unseen users" code in the runner. You'd have `userIds?: string[] | undefined` defined on `TickPayload`. This function would add to the existing array, and then you'd expect the runner code to dedupe and pass usernames to the runtime.

packages/xxscreeps/mods/market/transaction.ts
```diff
@@ -0,0 +1,59 @@
...
+	incoming: string[];
+	outgoing: string[];
+	blobs: Record<string, Readonly<Uint8Array>>;
+	usernames: Record<string, string>;
+}
+
+// Register each party so the getters resolve, then overlay each referenced blob.
+export function readTransactions(payload: TransactionPayload): Transactions {
+	for (const [ userId, username ] of Object.entries(payload.usernames)) {
+		userInfo.set(userId, { username });
```
laverdet: Yeah, we really don't want every mod that needs to handle a username do this. Usernames are a feature of the runtime.

packages/xxscreeps/mods/market/transaction.ts
```diff
@@ -0,0 +1,59 @@
...
+ * `outgoingTransactions`. It is stored once as an immutable blob and referenced from each party's
+ * list, so the runner hands both parties' runtimes the same buffer. The parties are kept as user ids
+ * and resolved to usernames at read time; the description is kept raw and HTML-escaped by its getter.
+ */
+export class Transaction extends withOverlay(BufferObject, shape) {
+	@enumerable get sender() { return userInfo.get(this['#sender']); }
+	@enumerable get recipient() { return userInfo.get(this['#recipient']); }
+	@enumerable get description() {
+		const description = this['#description'];
+		return description == null ? undefined : description.replace(/</g, '&lt;');
```
laverdet: `replaceAll('<', '&lt;')`

packages/xxscreeps/mods/market/transaction.ts
```diff
@@ -0,0 +1,59 @@
...
+import { userInfo } from 'xxscreeps/game/index.js';
+import { resourceEnumFormat } from 'xxscreeps/mods/resource/resource.js';
+import { BufferObject, compose, declare, optional, struct, withOverlay } from 'xxscreeps/schema/index.js';
+
+const shape = struct({
+	transactionId: Id.format,
+	time: 'int32',
+	resourceType: resourceEnumFormat,
+	amount: 'int32',
+	from: 'string',
```
laverdet, line 13: I'm not saying you need to do this, but I think we've long passed the threshold where passing room names along as strings in format blobs is acceptable.

packages/xxscreeps/mods/market/model.ts
```diff
@@ -0,0 +1,99 @@
...
+			shard.data.hDel(refsKey, [ id ]),
+			shard.data.del(blobKey(id)),
+		]);
+	}
+}
+
+async function pushReference(shard: Shard, userId: string, direction: Direction, entry: string) {
+	const key = listKey(userId, direction);
+	// One push adds one entry, so at most one is over the cap. Pop a single oldest; trimming by
+	// `length - kMaxTransactions` over-evicts when concurrent pushes each observe a racing length.
```
laverdet: I'm not sure about this comment. lTrim accepts an offset from the end boundary. So you'd do `lTrim(0, -max)` or whatever and there is no race.

packages/xxscreeps/mods/market/model.ts
```diff
@@ -0,0 +1,99 @@
+import type { Shard } from 'xxscreeps/engine/db/index.js';
+import type { ResourceType } from 'xxscreeps/mods/resource/resource.js';
+import * as Id from 'xxscreeps/engine/schema/id.js';
+import { assign } from 'xxscreeps/utility/utility.js';
+import { Transaction, write } from './transaction.js';
+
+// Terminal transfers are normalized: each is stored once as an immutable blob at
+// `market/transaction/<id>` and referenced from each party's per-direction list. A list entry is
+// `<createdMs>|<id>` — the timestamp gates the 24h read window, the id dereferences the blob. A list
```
laverdet: I'm not crazy about this scheme where you parse numeric data from a string key name.

What would happen if you put them in a sorted set? You could also merge incoming and outgoing transactions into the same set this way by using negative timestamps to represent one direction and positive the other. Trimming the set in that case becomes a range trim by [-now, now].

packages/xxscreeps/mods/market/model.ts
```diff
@@ -0,0 +1,99 @@
...
+	description?: string | undefined;
+}
+
+export function loadTransactionBlob(shard: Shard, id: string) {
+	return shard.data.req(blobKey(id), { blob: true });
+}
+
+async function dropReference(shard: Shard, id: string) {
+	// One list no longer references this transfer; free the blob once neither party does.
+	if (await shard.data.hincrBy(refsKey, id, -1) <= 0) {
```
laverdet: I don't think you need this ref counting thing. Just use `px` and expire them after the timeout. Leave a TODO on the blob provider to implement `px`.

misterwise:
Round-2 revision pushed. Mapped to your comments:

**Storage (`model.ts`)**
- **Sorted sets, no string-key parsing** — each direction is a sorted set
  (`user/<id>/market/transactions/{incoming,outgoing}`), score = wall-clock ms, member = id.
  The `<ms>|<id>` parsing is gone.
- **Dropped refcounting for `px`** — the ref-count hash is gone; the blob is written with
  `set(…, { px: kTransactionWindow })` and there's a TODO on `BlobStorage.set` (it ignores `px`
  today). Set entries are score-trimmed to the window on write (`zRemRange`), so the sets stay
  bounded regardless. Until the provider honors `px` the blobs leak past the window — the
  tradeoff you signed off on.
- The "pop-one-per-push avoids a race" comment went out with the list it described.

**Runner (`driver.ts` + `runner/{index,instance}.ts`)**
- **Channel, not per-tick reload** — the connector subscribes to a per-user channel; the
  processor publishes on record. It reloads from the DB only on a dirty signal, and re-applies
  the 24h window in memory each tick.
- **Blob cache keyed by id only** — module-level `Map<id, Promise<blob>>`; no time, no shard.
- **Usernames are the runtime's job** — added `TickPayload.userIds`; the connector pushes the
  parties onto it and the runner's "load unseen users" path merges + resolves them alongside
  room `#users`. The mod no longer carries a usernames map or touches `userInfo`.

**`transaction.ts`**
- `replaceAll('<', '&lt;')`.
- Room names as strings: left as-is (you flagged it as optional).

New test: a cross-user send where the sender shares no room with the recipient, so its username
can only resolve through `payload.userIds` (the existing self-send test didn't exercise that).

On merging the two directions into one signed-score set: it doesn't survive self-sends. A
terminal transfer to your own terminal in another room is both incoming and outgoing for you
(vanilla matches the one record in both queries, and the existing self-send test asserts both),
but a single sorted set can't hold one transaction id at two scores — the member is unique, so
the `+time` and `-time` writes collide and one direction is lost. Disambiguating by prefixing
direction onto the member just moves the string-parsing you flagged out of the score and into
the member. So I kept a set per direction: same per-record cost (a transfer touches one
direction for one user either way), each a clean id-at-time set.

packages/xxscreeps/mods/market/driver.ts
```diff
@@ -0,0 +1,52 @@
...
+			}
+			const { incoming, outgoing } = await entries!;
+			// Re-apply the window every tick so transfers age out without a reload.
+			const cutoff = Date.now() - kTransactionWindow;
+			const incomingIds = selectRecent(incoming, cutoff);
+			const outgoingIds = selectRecent(outgoing, cutoff);
+			const ids = [ ...new Set([ ...incomingIds, ...outgoingIds ]) ];
+			const blobList = await Promise.all(ids.map(id => loadBlob(player.shard, id)));
+			// Ask the runner to resolve the parties' usernames for the read-time getters.
+			const parties = new Set<string>();
```
laverdet: Deduping userids here doesn't make sense because there could be dupes elsewhere in the runtime. I added a dedupe to the runner instance, so this part can be removed.

Also we don't need to do all this stuff every tick. If a player's transactions haven't changed then no work should be done. I think it's fine if the transactions fall out inconsistently when new ones push the channel. If it ends up not being fine then it would be better to have another solution that tells the runtime to drop certain transactions when they time out.

I'd like it if only new transactions were sent to the runtime, instead of rebuilding the whole list each time but the market probably isn't super active so it's not a big deal either way.

misterwise:
Round-3 pushed (4eb94095) on top of your `tweak` — the instance dedupe + `getOrSet` blob cache are yours; this commit does the two driver-side items:

- **No work on quiet ticks.** `refresh` now returns immediately unless the channel signalled a transfer. On a dirty tick it reloads, rewindows, and ships the list; otherwise it does nothing and the runtime keeps the last payload it was sent (`game.ts` holds `transactions` at module scope, like the memory mod's `foreignSegment`). So aged-out transfers linger until the next transfer rewindows the list — the tradeoff you called fine.
- **Resend after a reset.** Connectors outlive sandbox resets but the runtime's retained state doesn't, and transactions aren't re-requested by player code the way foreign segments are — so a bare skip-when-clean would blank a trader's history on every code push. An `initialize() { dirty = true }` hook covers it (fires on first sandbox and after every reset, forcing one resend next tick).
- **Dropped the driver-side dedupe.** Both parties' ids are pushed raw onto `payload.userIds` now that your `new Set` in the runner dedupes them.

Left the optional "only send new transactions" out, per your note. Full suite green (449).

laverdet:
I pushed some non-trivial changes here.

- Offload trimming and sorting to keyval. `selectRecent` is the job of the database layer.
- Transaction blobs are instantiated in the runtime lazily
- `@cached` decorator for the above
- `--test-redis` flag on the test runner to test against a local redis instance

I'm not positive I got the sorting correct. It seems like the most recent transaction should be `market.incoming[0]` based on what I'm seeing in Screeps' source, so that's what I did. The API documentation doesn't clearly specify it either way.

laverdet:
I made a thing in 05be3b2e0dc25731adeacdced841977bdab9a491 that I think you'd want pretty soon. You can use it to tamper with schema blobs directly in keyval stores. I imagine for market orders where the cost can change frequently it would be useful. Demonstrated below with `amount` on the otherwise immutable transaction schema.

```
$ git dc
diff --git a/packages/xxscreeps/mods/market/transaction.ts b/packages/xxscreeps/mods/market/transaction.ts
index 9ee6dcac..4117d06c 100644
--- a/packages/xxscreeps/mods/market/transaction.ts
+++ b/packages/xxscreeps/mods/market/transaction.ts
@@ -34,7 +34,9 @@ export class Transaction extends withOverlay(BufferObject, shape) {
     }
 }
 
-export const { read, write } = makeReaderAndWriter(format);
+export const { offsetOf, read, version, write } = makeReaderAndWriter(format);
+export const offsetOfAmount = offsetOf('MarketTransaction', 'amount');
+export const transactionSchemaVersion = version;
 
 export class Transactions {
     readonly payload;
```

```ts
// $ cat  packages/xxscreeps/scripts/arst.ts 
import * as assert from 'node:assert/strict';
import { Database } from 'xxscreeps/engine/db/index.js';
import { UpdateSchemaBlob } from 'xxscreeps/engine/schema/keyval.js';
import { Transaction, offsetOfAmount, read, transactionSchemaVersion, write } from 'xxscreeps/mods/market/transaction.js';
import { assign } from 'xxscreeps/utility/utility.js';
import 'xxscreeps:mods/game';

await using db = await Database.connect();
const key = 'foo';

const transaction = assign(new Transaction(), {
  transactionId: 'badc0de',
  time: 123,
  resourceType: 'energy',
  amount: 100,
  from: 'W0N0',
  to: 'W0N1',
});
await db.data.set(key, write(transaction));

const result = await db.data.eval(UpdateSchemaBlob, [ key ], [ transactionSchemaVersion, offsetOfAmount, 'int32', 123, 'incr' ]);
const blob = await db.data.get(key, { blob: true });
assert.ok(blob);
console.log(typeof result, result);
const next = read(blob);
console.log(next.amount);
```

```
$ node --enable-source-maps --import xxscreeps/loader packages/xxscreeps/dist/scripts/arst.js
number 223
223
```

laverdet:
Note it only handles the "happy" path where the correct schema version is in the keyval store. So in the failure case you'd still need to read + write in order to upgrade the schema. I don't think the market order will ever need to change though so it's probably not a problem. You might want to tweak the return value of `UpdateSchemaBlob` to differentiate between a value not found vs schema version mismatch.

laverdet:
Oh yeah it would only work for statically known field offsets. So for example it wouldn't work on the power creep roster since that is a vector, unless you wanted to do a blob per power creep.

PR #284, opened 2026-06-27
Add game pause controls

bastianh:
This is a PR that makes it possible to pause/unpause the game. Not sure if it makes sense for you, or if you have something different in mind ;)

### DIFF
--- a/packages/xxscreeps/engine/service/control.ts
+++ b/packages/xxscreeps/engine/service/control.ts
@@ -1,0 +1,23 @@
+import type { Shard } from 'xxscreeps/engine/db/index.js';
+import { Channel } from 'xxscreeps/engine/db/channel.js';
+
+const gamePausedKey = 'game/paused';
+
+export function getGameControlChannel(shard: Shard) {
+	type Message =
+		{ type: 'pauseChanged'; paused: boolean };
+	return new Channel<Message>(shard.pubsub, 'channel/game-control');
+}
+
+export async function isGamePaused(shard: Shard) {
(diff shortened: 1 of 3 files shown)


misterwise:
As a heads up, I was independently adding a pause verb to the `manage` script too, just via the `game` mutex rather than engine state, so there's some overlap here. Context worth surfacing: on my CLI PR (#79), laverdet's steer for pause/resume was to avoid main-loop special-casing and instead hold the mutex to suspend the loop:

> I don't think you need special-casing here in the otherwise very lean main loop. The CLI pause/resume thing could just acquire a gameMutex lock which would suspend the main loop.

That said, the durable-flag approach here has upsides a mutex hold lacks, decoupled from the operator process, survives a crash, has `status`, so just flagging the tradeoff.

(I have also pursued single-step ticks, though only to ±1 since the mutex yields at tick boundaries. Happy to share.)

bastianh:
I think a cli is fine but at some point there will probably be mods to add an admin api to do maintanence tasks, add newbie zones, restrict areas, update users and stuff like that. I also like having the flag in the storage to be able to start/stop the server easily without ticking
I think I tried a way to implement it as a mod without changes to the core .. but I think in that case the server would always hold the mutex .. which would make running commands that would require the mutex impossible while the server is running

laverdet:
I took a different approach here ef701d9bccf356bfe118670095249f28412d7e3e. I don't think it really needs to be a database flag. I always found it annoying in Screeps that a paused loop would stay paused on restart. I think if you want to start the service in a paused state it would make more sense to have a --paused flag on the launcher.

PR #285, opened 2026-06-28
Deploy stronghold structures from a deployed invader core

misterwise:
Continues the invader-core arc (after #215, #251, #274): when a core's deploy
timer elapses it deploys its stronghold, dropping invulnerability, starting a
collapse timer, and spawning the surrounding template structures.

Collapse rides a `#collapseTime` scalar on the core; when it elapses the core's
pre-tick silently removes every NPC-owned structure in the room, after first
releasing a controller the core took over (resolving the reset TODO in
`mods/invader/processor.ts`). Decaying peers are pinned to the collapse time
(`#nextDecayTime`) so they don't decay while the room sleeps.

The layout is a stub; canonical bunker templates, reward containers, world
placement, and defender behaviors are later slices.

## Validation

- Added `Invader core stronghold deployment` (4 tests) to `mods/invader/test.ts`:
  deploy swaps the deploy timer for a collapse timer and spawns the template
  (owned tower/rampart, decay pinned to collapse); collapse sweeps the NPC-owned
  structures silently and spares other owners; a controller-owning core releases
  it to neutral.
- Exercised on a running multi-service server: deploy spawned the template live,
  a restart's boot pass reprocessed the deployed stronghold without error, and
  shortening the timer removed the core and peers together, silently, with no
  ruins.


packages/xxscreeps/mods/invader/processor.ts
```diff
@@ -237,6 +233,63 @@ registerObjectTickProcessor(StructureInvaderCore, (core, context) => {
...
+// collapse time so they don't decay (and read a past expiry) while the room sleeps until collapse.
+function *strongholdTemplate(pos: RoomPosition, collapseTime: number) {
+	const tower = createTower(new RoomPosition(pos.x, pos.y - 1, pos.roomName), '2');
+	const rampart = createRampart(new RoomPosition(pos.x + 1, pos.y, pos.roomName), '2');
+	const container = createContainer(new RoomPosition(pos.x - 1, pos.y, pos.roomName));
+	const road = createRoad(new RoomPosition(pos.x, pos.y + 1, pos.roomName));
+	rampart['#nextDecayTime'] = collapseTime;
+	container['#nextDecayTime'] = collapseTime;
+	road['#nextDecayTime'] = collapseTime;
+	yield* [ tower, rampart, container, road ];
```
laverdet: Ii know I like iterables and generators but if you're going to materialize an array here you might as well just return it.

packages/xxscreeps/mods/structure/structure.ts
```diff
@@ -32,6 +32,9 @@ const attackNotificationHandlers: AttackNotificationHandler[] = [];
 export const structureFormat = declare('Structure', () => compose(shape, Structure));
 const shape = struct(objectFormat, {
 	'#noAttackNotify': 'bool',
+	// End time of a stronghold collapse timer; while set the structure is removed silently once it
+	// elapses. Default 0 means no timer. Set on an invader core and its deployed peers.
+	'#collapseTime': 'int32',
```
laverdet: Can we do this in the invader core processor? When the core collapses it could just remove all non-keeper lair structures belonging to the source keeper

packages/xxscreeps/mods/invader/processor.ts
```diff
@@ -237,6 +233,63 @@ registerObjectTickProcessor(StructureInvaderCore, (core, context) => {
 	}
 });
 
+const isCollapsing = (structure: Structure) => {
+	const collapseTime = structure['#collapseTime'];
+	return collapseTime > 0 && collapseTime <= Game.time;
```
laverdet: `optionalExpiryTime(...) == 0` to enforce those invariants I care about.

packages/xxscreeps/mods/invader/processor.ts
misterwise: Done, it returns the array now.

packages/xxscreeps/mods/structure/structure.ts
misterwise: Done: `#collapseTime` is back on the core's format and the core's pre-tick now sweeps every structure owned by '2' at collapse (keeper lairs sit on '3', so ownership alone excludes them). The stub's neutral container/road are invisible to an ownership sweep, so the template is down to tower + rampart; the canonical-templates slice will need an answer for neutral reward containers.

One divergence to flag: a core destroyed before the timer elapses no longer leaves peers that self-remove at the original collapse time (vanilla sweeps each object through its own collapse effect). The pinned rampart decays away on its own; a tower would persist. Happy to address that in the templates slice.

packages/xxscreeps/mods/invader/processor.ts
misterwise: Done.

laverdet:
38d15edad1dc8abb028ed3b66f67c257ba7f59a9 improves mod support for fields on core game objects. So I more or less reverted 26e24c7a1224c02d56c65efb6e1a6c9ceed38470 and used the new thing to implement this. The stronghold mod adds `#collapseTime` to all structures and handles the collapsing behavior.

PR #287, opened 2026-06-29
notifications: add /api/user/notify-prefs endpoint and expose prefs on /api/auth/me

bastianh:
Add /api/user/notify-prefs endpoint and surface prefs on /api/auth/me

@misterwise  maybe you can also check since I think that code is from you .)

## Summary

The notifications mod already had the storage helpers and the runtime worker for notification preferences, but there was no API surface for them, so the official Angular client's account screen couldn't read or change notification settings. This PR wires that up to match the original screeps-server behavior.

#### 1. POST /api/user/notify-prefs (new mods/notifications/backend.ts)

Mirrors the original server's handler:
- Auth-gated on the logged-in user.
- Validates the JSON body and maps disabled / disabledOnMessages / sendOnline (booleans) and interval / errorsInterval (numbers).
- interval / errorsInterval are checked against the original's allowed value sets (5,10,30,60,180,360,720,1440,4320 and 0,…,100000); out-of-range values are silently ignored, as in vanilla.
- Persists via setNotifyPrefs and returns { ok: 1 }.

▎ Note: the Angular client ignores the response body (it just re-fetches the user via /api/auth/me on success), so the return shape doesn't need to echo the original's raw DB-update result.

#### 2. Prefs surfaced on /api/auth/me (a sendUserInfo hook)

The account UI computes each toggle from Me.notifyPrefs.* (e.g. it posts { disabled: !Me.notifyPrefs.disabled }). Nothing populated notifyPrefs on the me payload, so the toggles read undefined → always-falsy and couldn't reflect or correctly flip saved state. The new hook adds userInfo.notifyPrefs for the caller's own profile.

#### 3. Prefs moved from per-shard storage to the global DB

getNotifyPrefs / setNotifyPrefs now take a Database and read/write db.data instead of shard.data. Notification prefs are a user-level setting (like the original db.users.notifyPrefs), so they belong in the shared global store — not least because the sendUserInfo hook only receives a Database, and the global store and a shard's store are physically separate. The notification queue and throttle state stay in shard.data, since those are genuinely per-world runtime state.

Callers updated accordingly:
- main.ts notification worker → getNotifyPrefs(shard.db, …)
- tests in mods/notifications/test.ts and mods/controller/test.ts → setNotifyPrefs(shard.db, …)

Behavioral differences from the original (intentional)

- Booleans are validated as real JSON booleans rather than coerced with !!, so the original's "false" → true quirk doesn't carry over. The official client always sends real booleans/numbers, so this is compatible.

### DIFF
--- a/packages/xxscreeps/mods/notifications/backend.ts
+++ b/packages/xxscreeps/mods/notifications/backend.ts
@@ -1,7 +1,6 @@
-import type { JSONSchemaType } from 'ajv';
 import type { NotifyPrefs } from './prefs.js';
-import { Ajv } from 'ajv';
-import { hooks } from 'xxscreeps/backend/index.js';
+import type { JSONSchemaType } from 'ajv';
+import { hooks, makeValidatedPayloadRoute } from 'xxscreeps/backend/index.js';
 import { getNotifyPrefs, setNotifyPrefs } from './prefs.js';
 
 // Mirrors the allowed values from the original screeps-server `/api/user/notify-prefs` handler.
@@ -18,9 +17,8 @@ interface NotifyPrefsRequest {
 }
 
 // Note: unlike vanilla (which coerces booleans with `!!`, so "false" becomes true), these are
-// validated as real JSON booleans. coerceTypes handles numeric fields sent as strings (e.g. the
+// validated as real JSON booleans. `coerceTypes` handles numeric fields sent as strings (e.g. the
 // official client sends interval as "180" rather than 180).
-const ajv = new Ajv({ coerceTypes: true });
 const notifyPrefsRequestSchema: JSONSchemaType<NotifyPrefsRequest> = {
 	type: 'object',
 	properties: {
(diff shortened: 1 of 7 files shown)


packages/xxscreeps/mods/notifications/backend.ts
```diff
@@ -0,0 +1,72 @@
...
+const notifyPrefsRequestSchema: JSONSchemaType<NotifyPrefsRequest> = {
+	type: 'object',
+	properties: {
+		disabled: { type: 'boolean', nullable: true },
+		disabledOnMessages: { type: 'boolean', nullable: true },
+		sendOnline: { type: 'boolean', nullable: true },
+		interval: { type: 'number', nullable: true },
+		errorsInterval: { type: 'number', nullable: true },
+	},
+	additionalProperties: false,
```
misterwise, line 31: Relatively minor, but this would be a behavioural divergence from vanilla which accepts (silently) unknown body fields. This would cause an error instead. https://github.com/screeps/backend-local/blob/9d079282303ec04e577ac2bb97f64312b25e4ccd/lib/game/api/user.js#L389

PR #288, opened 2026-06-29
messages: implement private messaging feature

bastianh:
This is probably a bit work in progress, but already would like to collect feedback ,)

Add a first-party messages mod (loaded via classic) providing the full private-message API the official client expects: send, list, index, mark-read and unread-count, plus the live newMessage and read-receipt socket channels. Messages are stored account-level in db.data as linked in/out documents with a strictly-increasing sequence for deterministic ordering. Adds a User.remove hook so a deleted user's messages are torn down, replacing the old unread-count stub.


packages/xxscreeps/mods/messages/backend.ts
```diff
@@ -0,0 +1,184 @@
...
+import * as User from 'xxscreeps/engine/db/user/index.js';
+import { sendNotification } from 'xxscreeps/mods/notifications/model.js';
+import { getNotifyPrefs } from 'xxscreeps/mods/notifications/prefs.js';
+import {
+	getConversation, getConversationIndex, getMessageChannel, getNewMessageChannel,
+	getUnreadCount, markRead, sendMessage,
+} from './model.js';
+
+// Mirrors the original server's cap; longer payloads are rejected rather than truncated.
+const kMaxMessageLength = 2000;
```
laverdet: Is the limit on MMO different? In backend-local it's 100kb and allows empty messages:
https://github.com/screeps/backend-local/blob/9d079282303ec04e577ac2bb97f64312b25e4ccd/lib/game/api/user-messages.js#L40-L42

packages/xxscreeps/mods/messages/model.ts
```diff
@@ -0,0 +1,262 @@
...
+const messageKey = (messageId: string) => `messages/${messageId}`;
+// zset: score = sequence of the last message, member = respondentId. Drives the conversation index.
+const conversationsKey = (userId: string) => `user/${userId}/messages/conversations`;
+// zset: score = global send sequence, member = messageId. One conversation thread.
+const threadKey = (userId: string, respondentId: string) => `user/${userId}/messages/with/${respondentId}`;
+// set of unread incoming messageIds. unread-count is an O(1) `sCard` over this.
+const unreadKey = (userId: string) => `user/${userId}/messages/unread`;
+// Strictly-increasing send counter, used as the zset score so message ordering is deterministic
+// even when several messages land in the same millisecond (which `date` cannot disambiguate).
+const sequenceKey = 'messages/sequence';
```
laverdet: My career is in social networking and I can tell you that no one cares if same-ms messages are reordered. The transmission of light over fairly small distances is easily in >1ms range so at a very physical level it is not even clear which message was "first".

packages/xxscreeps/mods/messages/backend.ts
bastianh: that looks like a hallucination .. I saw it but did not really thought about it .. I found some differences for example in map-stats between pserver and mmo ( the private server always sends owner0 + mineral0 ) but the mmo server only sends one, depending on what you ask because there are other stats like energy and power usage possible .. but thats a different topic .)

packages/xxscreeps/mods/messages/model.ts
```diff
@@ -0,0 +1,262 @@
...
+			return false;
+		}
+		await Promise.all([
+			db.data.hSet(messageKey(messageId), 'unread', '0'),
+			db.data.sRem(unreadKey(userId), [ messageId ]),
+		]);
+		// Flip the sender's read receipt and notify their open sessions.
+		const peer = await readMessage(db, row.peer);
+		if (peer?.unread === '1') {
+			await db.data.hSet(messageKey(row.peer), 'unread', '0');
```
laverdet: If you delete 'unread' when the message is read then you can use hsetnx to mark it as unread, publishing to the channel if it returned as set. Then you don't have to do the read + update thing.

packages/xxscreeps/mods/messages/model.ts
```diff
@@ -0,0 +1,261 @@
...
+//
+// All of this is the *default* store, registered as the fallback of `messageStore`. A mod can call
+// `messageStore.register(...)` to replace the persistence wholesale with a different backend (e.g. a
+// SQL store with native queries) without forcing every other account feature off the keyval store.
+// An override owns the full contract, including publishing on the channels below for live updates.
+
+export type MessageType = 'in' | 'out';
+
+export interface Message {
+	_id: string;
```
laverdet: `_id` fields are a weird artifact of screep's architecture. I prefer to just call these `id` and then inject `_id` into the backend payload later if the screeps client requires it.

packages/xxscreeps/mods/messages/model.ts
```diff
@@ -0,0 +1,261 @@
...
+		const date = String(now);
+		const inId = generateId(24);
+		const outId = generateId(24);
+
+		// `in` copy: owned by the recipient, unread until they read it.
+		// `out` copy: owned by the sender, `unread` here means "not yet read by the recipient" (the
+		// read receipt), flipped by `markRead`.
+		await Promise.all([
+			db.data.hmset(messageKey(inId), {
+				user: respondentId, respondent: senderId, date, type: 'in', text, unread: '1', peer: outId,
```
laverdet: Is in/out doing anything for us here? Can't this be inferred from which set it belongs to and/or the actor's userid?

packages/xxscreeps/mods/messages/test.ts
```diff
@@ -0,0 +1,126 @@
...
+	});
+
+	test('index lists latest message per respondent, newest first', async () => {
+		await using testShard = await instantiateTestShard();
+		const { db } = testShard;
+
+		await sendMessage(db, alice, bob, 'first');
+		// Thread order is scored by wall-clock ms with no tiebreaker, so give these two sends
+		// distinguishable timestamps (real messages are human-paced and never collide like this).
+		await new Promise(resolve => setTimeout(resolve, 2));
```
laverdet: There's `deterministicClockForTesting` in main now which will increment the clock by 1 each time `Date.now()` is invoked so you can remove the timeout.

packages/xxscreeps/mods/messages/test.ts
```diff
@@ -0,0 +1,126 @@
...
+		await sendMessage(db, alice, bob, 'hello bob');
+
+		const aliceThread = await getConversation(db, alice, bob);
+		assert.strictEqual(aliceThread.length, 1);
+		assert.strictEqual(aliceThread[0]!.type, 'out');
+		assert.strictEqual(aliceThread[0]!.text, 'hello bob');
+
+		const bobThread = await getConversation(db, bob, alice);
+		assert.strictEqual(bobThread.length, 1);
+		assert.strictEqual(bobThread[0]!.type, 'in');
```
laverdet: Prefer `?.` when `!` is not actually needed (all over the test)

packages/xxscreeps/mods/messages/model.ts
```diff
@@ -0,0 +1,261 @@
...
+// SQL store with native queries) without forcing every other account feature off the keyval store.
+// An override owns the full contract, including publishing on the channels below for live updates.
+
+export type MessageType = 'in' | 'out';
+
+export interface Message {
+	_id: string;
+	type: MessageType;
+	text: string;
+	// ISO-8601 string, matching the official server's JSON-serialized Mongo date.
```
laverdet: I know it's what Screeps does but is there a reason not to just store these as numeric timestamps? ISO strings don't have timestamp information so it seems like unneeded information for us. Also we already store the timestamp in the sorted set so that information could just be extracted from the keyval.

laverdet (commented):
This is mostly pretty good. I like the hooks for user removal and implementation details. Those could already be used for power creeps (user removal) and notifications (implementation swap), probably others too.

I slept on it for a little while and I'd like the messages to be deduplicated. So instead of each user getting their own copy of the message they would both reference the same message object. A 100kb message payload is a lot to double on small redis distributions. This also simplifies the peer unread state.

Architecturally I think most of the issues are a product of too much influence from the existing Screeps implementation. My approach to building this project is to look at a feature blind to the Screeps codebase. I build out something from scratch and really I'm only looking at the Screeps repository for messy implementation details that are undocumented.

packages/xxscreeps/mods/messages/test.ts
bastianh: hmm..  under this repo's noUncheckedIndexedAccess: true + no-unnecessary-condition combo, ?. on array-index access gets flagged as unnecessary while tsc still requires non-null handling — so ! is the only lint-clean option and matches the repo's own convention

PR #290, opened 2026-06-30
Add offline single-room generation

misterwise:
Stacked on #293 — this PR's own delta is `scripts/room-gen.ts`, the
`generate-room` command, and a neutral-owner default on `extractor.create`;
the commits below `room-gen:` are that PR.

Adds `generateRoom(shard, name, options)` — an offline engine helper that
procedurally generates a single room of any type (normal, source-keeper,
center, highway) into a stopped shard's terrain blob and `rooms` set — plus a
thin `generate-room` script that wraps it. Like `import`, it's an offline
mutator: stop the server first, or the backend/processor/runner workers run on
stale world state. Existing rooms in the target footprint are reused and exits
are matched to neighbors, so adjacent generated rooms connect.

The generator deals in #293's authored geometry: `buildRoom` composes the
room's `meta` (template classification plus authored `roomType`/`centers`
overrides), generates from it, and returns the same value for stamping into
the World schema — the generation input and the stored geometry are one thing.
No room-name arithmetic survives here: a highway's wall-mass orientation is
read from its ring memberships (the axis with the greater delta to a ring's
center is the edge it rides; tying deltas mark a crossing). The comparison
carries no ring radius, so authored mini/mega-sector rings orient the same way
the vanilla ±5 rings do.

Terrain is cellular-automata generated with neighbor-matched exits; sources,
controller, mineral + extractor, and keeper lairs are placed per type. There
is no public vanilla highway generator — `@screeps/backend`'s `cli/map.js`
only CA-fills isolated rooms — so highway corridor terrain is a from-scratch
noise model tuned against the live world. A `--type {normal|keeper|center|highway}`
preset applies a canonical loadout over the lower-level shape flags.

## Validation

No unit tests: this is offline tooling and mirrors `scrape-world.ts`
(`import`), which ships none. Ran `pnpm build` and eslint clean, and
smoke-tested `generate-room` against a local shard for each `--type`,
confirming the generated rooms load and render. The centers-derived highway
orientation was swept against the old mod-10 arithmetic across an 80×80 world
(1216 highway rooms): zero mismatches, and only highways carry ring
memberships. Highway terrain was validated separately against a 14,884-room
shard3 capture (wall-density, swamp, and lane distributions within KS<0.1 of
real); that statistical tooling lives on a private branch.


packages/xxscreeps/scripts/room-gen.ts
```diff
@@ -0,0 +1,1010 @@
...
+}
+
+interface TerrainTypeParams {
+	fill: number;
+	smooth: number;
+	factor: number;
+}
+
+const wallTypes: Record<number, TerrainTypeParams> = {
+	1: { fill: 0.4, smooth: 10, factor: 5 },
```
laverdet, line 57: Aren't these just arrays with more steps?

packages/xxscreeps/scripts/room-gen.ts
```diff
@@ -0,0 +1,1010 @@
...
+import { create as createKeeperLair } from 'xxscreeps/mods/source/keeper-lair.js';
+import { Source } from 'xxscreeps/mods/source/source.js';
+import { makeWriter } from 'xxscreeps/schema/write.js';
+
+type ExitSide = 'top' | 'right' | 'bottom' | 'left';
+type ExitMap = Record<ExitSide, number[]>;
+
+// The world's per-room terrain map, as returned by `shard.loadWorld()`. The generation entry points
+// accumulate freshly built rooms into one of these and serialize it once, rather than once per room.
+type WorldTerrain = Awaited<ReturnType<Shard['loadWorld']>>['terrain'];
```
laverdet, line 27: This is already defined as `TerrainByRoom`

laverdet (commented):
Do you think this is thinking too small about what makes a Screeps MMO map? The endless grid of 9x9 sectors is kind of tired. I always thought that all shards following the exact same repeating geometry was a missed opportunity.

I've considered room metadata that lives outside the room payload, as part of the "world" geometry. You could use it store dispositions [has: minerals, keepers, forbid nukes, etc] or relationships [sector center (if any), sector members]. That could open the door for mini-sectors, mega-sectors, or no sectors at all.

Also what if not all rooms were 50x50? I could imagine 200x200 contestable Arena-style rooms connected to the world but living in the same shard.

Anyway, I think at a minimum before working on terraformation I'd like to see a generic metadata structure along those lines. Basically `sector.ts` would be blasted and that information would be part of the world's data. I don't have much guidance to go on there though.

misterwise:
> Do you think this is thinking too small about what makes a Screeps MMO map? The endless grid of 9x9 sectors is kind of tired. I always thought that all shards following the exact same repeating geometry was a missed opportunity.
> 
> I've considered room metadata that lives outside the room payload, as part of the "world" geometry. You could use it store dispositions [has: minerals, keepers, forbid nukes, etc] or relationships [sector center (if any), sector members]. That could open the door for mini-sectors, mega-sectors, or no sectors at all.
> 
> Also what if not all rooms were 50x50? I could imagine 200x200 contestable Arena-style rooms connected to the world but living in the same shard.
> 
> Anyway, I think at a minimum before working on terraformation I'd like to see a generic metadata structure along those lines. Basically `sector.ts` would be blasted and that information would be part of the world's data. I don't have much guidance to go on there though.

Honestly, I hadn't considered it. I was more focused (once again) on parity, and trying to generate novel, but vanilla style worlds. I think I can put together a foundational PR with a start on the metadata you are proposing and then rebase this PR on that. I'll start with metadata that supports 'vanilla' generation but once you mention it, I would love to have more flexibility in world generation without hardcoding vanilla maps permanently. I'll mark this as a draft until you're happy with the initial metadata format.

misterwise:
If #293 works, then this is ready to go and built on top of that foundation.

laverdet:
Could you break off the room terrain generation logic into its own PR? I think that is a mostly straightforward adaption of the existing logic. We can work out how mods affect that later.

misterwise:
> Could you break off the room terrain generation logic into its own PR? I think that is a mostly straightforward adaption of the existing logic. We can work out how mods affect that later.

Ok, I've got a split out worktree of the terrain generation ready to go once #293 is confirmed.

misterwise:
Superseded: the terrain half landed as #308, the world-metadata foundation as #293, and the object-placement half is now #321 (rebuilt fresh on main rather than rebasing this branch). The highway/sector work will follow separately once the metadata direction settles.

PR #293, opened 2026-07-01
Author sector geometry into the World schema

misterwise:
World geometry — which rooms make up a sector — was implicit in room-name arithmetic: `game/room/sector.ts` decoded coordinates at runtime to answer `isHighwayRoom` / `isCentralRoom` / `sectorsForRoom` / `sectorEdgeRooms`. This moves that geometry into the World schema as authored data, in the shape from review: a sector is a record carried by its center room — `meta.sector = { edges, members }` on the extensible `RoomMeta` path — stamped wherever terrain is authored (import, scrape) and read through `GameMap` (`getSector`, `sectors()`, `getSectorCenters`). One room is in charge of the sector; every other room pays a single absent-`optional` slot. Room names inside the record are stored packed through a reusable `roomNameFormat` (`uint16` ↔ name, `game/room/name.ts`).

`edges` is the 40-room highway ring at range 5, shared between adjacent sectors (mid-edge rooms register to 2, corners to 4, seam rooms to 1). `members` are the 81 rooms at range ≤ 4 — the center included — registered to exactly this sector; the two lists are disjoint. `members` has no runtime reader yet: it's the sector's exclusive territory roster, authored for the generation PR stacked on this one.

There is no runtime fallback to the mod-10 template. An absent record means the room anchors no sector — which is what keeps mini/mega/no-sector worlds authorable. The template math survives only as the authoring-time populator (`computeRoomMeta`), which clips each record to the rooms actually present. `roomType` is gone per review: highway-ness is the edge lists, keeper-ness is a generation-time input, and gameplay shakes out from placed objects. The room → owning-centers query the deposit decay path needs (`getSectorCenters`) is a lazily-built inverse of the stored records.

Consequences worth naming: a clipped world containing no center rooms seeds no power banks (previously any mod-10 highway room seeded; deposits were already sector-keyed, so the two policies now agree). As before, the World terrain blob has no upgrader, so existing worlds need re-import/re-scrape. Dispositions and external `RoomMeta` registrants stay as seams — the World reader is built at module load, so the deferred-build treatment (the Room schema's thunked shape) lands with the first mod that extends the path.

## Validation
- `tsc -b` clean; eslint clean on changed files (the empty `Schema {}` carries the same pre-existing `no-empty-object-type` warning as `game/room/index.ts`).
- Full suite (459) passes — deposit and powerbank placement read the sector records end-to-end. New tests pin the record shape (40 edges / 81 members), sharing multiplicity (corner = 4 sectors, mid-edge = 2, member = exactly 1), authoring-time clipping, and the stored-record ↔ inverse-index round-trip.

### DIFF
--- a/packages/xxscreeps/game/map.ts
+++ b/packages/xxscreeps/game/map.ts
@@ -1,17 +1,20 @@
 import type { ExitType } from './room/find.js';
 import type { Room } from './room/index.js';
+import type { RoomType } from './room/sector.js';
 import type { TypeOf } from 'xxscreeps/schema/index.js';
 import type { Adapter } from 'xxscreeps/utility/astar.js';
 
 import { build } from 'xxscreeps/engine/schema/index.js';
 import { primitiveComparator } from 'xxscreeps/functional/comparator.js';
 import { Fn } from 'xxscreeps/functional/fn.js';
-import { makeRoomName, parseRoomName } from 'xxscreeps/game/room/name.js';
-import { compose, declare, makeReader, struct, vector } from 'xxscreeps/schema/index.js';
+import { makeRoomName, makeRoomNameFromId, parseRoomName, parseRoomNameToId } from 'xxscreeps/game/room/name.js';
+import { compose, declare, enumerated, makeReader, struct, vector } from 'xxscreeps/schema/index.js';
 import { astar } from 'xxscreeps/utility/astar.js';
+import { getOrSet } from 'xxscreeps/utility/utility.js';
 import * as C from './constants/index.js';
 import { getDirection, getOffsetsFromDirection } from './direction.js';
 import { RoomPosition } from './position.js';
+import { computeRoomMeta } from './room/sector.js';
 import * as Terrain from './terrain.js';
 
 // Schema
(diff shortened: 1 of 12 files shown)


packages/xxscreeps/game/map.ts
```diff
@@ -1,31 +1,54 @@
...
 
 // Schema
 const roomTerrain = () => struct({
 	exits: 'uint8',
 	terrain: Terrain.format,
 });
+// The sector centers a room rings, stored as packed room ids (`uint16`) rather than name strings to
+// keep the full-shard terrain blob compact; composed to/from names so callers deal in room names.
+const sectorCenters = () => compose(vector('uint16'), {
+	compose: (ids: number[]) => ids.map(makeRoomNameFromId),
```
laverdet: This should be broken out into its own reusable `roomNameFormat`.

packages/xxscreeps/game/map.ts
```diff
@@ -1,31 +1,54 @@
...
+	decompose: (names: string[]) => names.map(parseRoomNameToId),
+});
+// Authored geometry, one record per room. `roomType` reserves enum index 0 for `undefined`, so a
+// room an authoring path leaves unstamped (or a hand-built world) reads back as unauthored and
+// `GameMap` recomputes it from the template instead of reporting it as `normal`.
+const roomMeta = () => struct({
+	roomType: enumerated(undefined, 'normal', 'highway', 'sourceKeeper', 'center'),
```
laverdet: This doesn't really capture the room traits all that well. highway / non-highway is a trait of the room geometry/terrain which is mostly distinct from the other traits that mainly affect entity placement. sourceKeeper is similar in that it only affects map generation, and gameplay naturally shakes out from the placed objects.

I think the way to start with this is a struct of something like:
```
const roomMeta = registerStruct('RoomMeta', {
	sector: optional(struct({
		edges: vector(roomNameFormat), // highways, registered to more than one sector
		members: vector(roomNameFormat),
	})),
});
```

I like putting sector metadata on one room instead of all rooms, and having that one room be in charge of things.

misterwise: Reworked to this shape: `sector: optional(struct({ edges, members }))` carried by the center room only, on the `RoomMeta` path (the core field is inlined in the base like the Room schema; external registrants go through `registerStruct`). Dropping `roomType` also dropped the template fallback — an absent record just means the room anchors no sector, so no-sector worlds stay authorable, and the populator clips each record to the rooms actually present at authoring time. `members` is the 9x9 interior including the center, exclusive to the sector (no runtime reader yet — authored for the generator to consume); `edges` is the shared ring. The room → owning-centers query the deposit decay path needs is a lazy inverse of the stored records.

laverdet:
- Merged RoomEntry and RoomMeta into RoomTraits
- TODO's about mods. Nothing about sectors are intrinsic to rooms; all this logic should be a mod
- GameMap is a public interface, no unscoped internal methods should leak out
- Sector control -> edge/member relationships are stored bidirectionally.

Note: The previous test where it puts an orphaned deposit in a corner room outside the extent radius is now incorrect. Since there is only one sector in the test shard, all coordinates in those corner rooms belong to W5N5. Breaking this test case actually validates our approach here.

PR #295, opened 2026-07-02
Add spawned power creep room entity

misterwise:
Adds the spawned power creep room form on top of the account layer. The spawn intent carries only the roster id; the processor atomically claims the authoritative unspawned entry and seeds the on-tile room object from its stored identity and powers. Spawn, movement, carry/store verbs, say, renew, suicide, aging, tombstones, and account cooldown writeback are included; applying learned powers remains a follow-up.

Power creeps share the classic creep processor arms for movement and carry/store behavior, including border crossing, road wear-out, hostile construction-site removal, and safe-mode movement priority.

## Scope and design notes

- Combat targeting and healing are follow-up work. `AttackTarget` and the heal verbs still exclude power creeps, apart from ranged mass attack reaching `#applyDamage`; the post-heal `hitsMax` clamp belongs with that integration.
- `notifyWhenAttacked` was deferred instead of leaving a write-only surface. It should return with combat integration, wired through the notifications mod's attack-notify path.
- The `classic/creep/processor.ts` extraction is behavior-identical substrate used by the power creep verbs. It can be split into a preliminary PR if that makes the review easier.
- The `PowerCreep.memory` setter still mirrors Creep's `??=` behavior. Vanilla assigns unconditionally, but changing only one class would fork the two APIs; both should change together in a separate PR, if at all.
- Vanilla power creep tombstones expose `powerCreep*` fields. `extend()` in `utility/utility.ts` is the established mechanism for adding public API members to a foreign class, so the getter half is solved. The remaining issue is structural: the creep-owned tombstone occupant schema is fixed, and `bindRenderer` has a single owner. This PR therefore keeps the current creep-shaped occupant pending a design that also composes cleanly with #301.

## Validation

- `pnpm run build` passed.
- The default suite passed 498/498 tests, and `npm test -- --private-transform=isolated-vm` passed 499/499 tests.
- ESLint reported zero errors and zero new warnings across all 11 changed TypeScript files.
- The rebased worktree was live-smoked on a worktree-native imported world against Redis 8.4. The same roster id was submitted to two active power spawns; exactly one on-tile entity materialized with the roster-stored name, class, level, and powers and a claimed `#ageTime`. It crossed W7N7 to W8N7, then suicide left one tombstone and wrote back a future cooldown while clearing `#ageTime` and the pending deletion marker. The server shut down cleanly.

### DIFF
--- a/.changeset/power-creep-entity.md
+++ b/.changeset/power-creep-entity.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Spawn power creeps into rooms with the shared creep carry verbs, aging, renew, and death.
--- a/packages/xxscreeps/mods/classic/creep/creep.ts
+++ b/packages/xxscreeps/mods/classic/creep/creep.ts
@@ -609,6 +609,31 @@ registerObstacleChecker(params => {
 
 //
 // Intent checks
+
+/**
+ * The shared carry/store verb surface. `Creep` and `PowerCreep` both satisfy it, so the
+ * transfer/withdraw/pickup/drop checks (and their processor arms) operate on this type rather than
+ * `Creep` — power creeps carry resources the same way but have no body or fatigue.
+ */
+export interface Carrier extends RoomObject {
+	my: boolean;
+	spawning?: boolean;
+	store: OpenStore;
+	'#user': string;
+}
+
+/** Room-presence preconditions every carry verb shares, with no body part involved. */
+export function checkCarrier(creep: Carrier) {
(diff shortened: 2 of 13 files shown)


laverdet:
The way I see operator's abilities working is that each operable structure would have some schema flag which adds the effect. The base mods would only declare and read the flag. Operator would then just set the flag on the game object. So all game mods shouldn't really need to know about powercreep. Really similar to how Screeps does it. My worry was that Operator would require a ton of hooks into each mod but I think just a handful of schema flags will do it.

PR #297, opened 2026-07-02
market: order book with credits and createOrder

misterwise:
Adds the `Game.market` order book on top of the transaction log (#281): `credits`, `createOrder`, `orders` / `getAllOrders` / `getOrderById`, with each order's state maintained by its owning terminal's room pass.

- An order is a mutable schema blob at `market/order/<id>` (owner in a hidden `#user`, price in millicredits behind a `÷1000` getter), referenced by id from the active book, the owner's order set, a creation-time zset, and the owning terminal's `'#orderIds'`.
- `createOrder` is an intent on the terminal. The runtime method and the processor share `checkOrderParams`; same-tick orders accumulate into one batched intent (the slot is unique per object/action, and vanilla allows several creates from one room per tick); the listing fee debits up front and refunds on a negative balance or a full order set, so concurrent room processors can't overdraw. Divergence: a non-positive or non-finite `totalAmount` returns `ERR_INVALID_ARGS` instead of vanilla's silent OK that never creates the order.
- The terminal's tick processor maintains its orders from local state: age expiry (refunding the unspent fee share), activation, and advertised volume — terminal stock for a sell; for a buy, the owner's affordable volume, tracked even while the order is inactive, bounded by the terminal's free space. A terminal holding orders keeps its room ticking, the same standing heartbeat a spawn provides. Amount-only changes — the common case as stock and credits fluctuate — are patched in place with `UpdateSchemaBlob`; activity flips rewrite the blob and move the active-book index. The shard pass loads no rooms: it's an expiry backstop over the creation-time zset for orders whose room never processes.
- Order writes publish their ids on a `market/orders` channel. The runner keeps a cross-tick blob cache invalidated by the channel plus a per-runtime dirty set, ships each runtime only blobs it hasn't seen, and the runtime folds the deltas into its retained book — the active book's blobs stay shared across every player on the process.

Known gaps, deliberate for this slice:

- A destroyed terminal orphans its orders: they keep their last `amount`/`active` state until the expiry backstop ages them out. Nothing can fill them before `deal` lands.
- `credits` is a per-tick `hGet` in the connector rather than a channel; happy to move it onto the transactions/orders shape if you'd rather.

`deal`, the order-manage verbs (`cancelOrder` / `extendOrder` / `changeOrderPrice`), and `getHistory` are follow-ups.

## Validation

- Full suite passed on the local provider (492/492), with `--test-redis` (492/492 — both `UpdateSchemaBlob` implementations), and with `--private-transform=isolated-vm` (493/493).
- screeps-ok behavior suite: 2435 passed, with the same 50 known failures as upstream main — no regressions.
- Live `xxscreeps start` smoke in the branch worktree (with the #310 sandbox fix cherry-picked — current main can't boot an isolated-vm runner without it): 423 clean ticks, graceful shutdown.
- Added tests: order creation / fee / refund / activation / expiry, two same-tick orders on one terminal both materializing, the in-place amount patch (asserts the blob is not rewritten), the runtime read getters and delta folding over written blobs, an end-to-end sandbox `createOrder` → book pipeline, numeric coercion, the `MARKET_MAX_ORDERS` cap, and dropping an order stored under a stale schema version.
- eslint on changed files: no new warnings.


packages/xxscreeps/mods/market/driver.ts
```diff
@@ -49,3 +54,43 @@ hooks.register('runnerConnector', async player => {
...
+// writer and reads are tick-synchronized), so the active book's id list and every order blob are
+// read once per (shard, tick) and shared across this runner process's player instances: the book is
+// identical for every player, and an active order that is also one of yours is read once.
+interface OrderCache {
+	time: number;
+	active: Promise<string[]>;
+	blobs: Map<string, Promise<Readonly<Uint8Array> | null>>;
+}
+
+const orderCaches = new WeakMap<Shard, OrderCache>();
```
laverdet: Each worker will only handle one shard, so you never need to scope on shard.

packages/xxscreeps/mods/market/driver.ts
```diff
@@ -49,3 +54,43 @@ hooks.register('runnerConnector', async player => {
...
+		const [ active, mine, money ] = await Promise.all([
+			cache.active,
+			loadUserOrderIds(player.shard, player.userId),
+			loadMoney(player.shard, player.userId),
+		]);
+		const ids = [ ...new Set([ ...active, ...mine ]) ];
+		// An id can outlive its blob (the shard pass removes orders between the id-set read and this
+		// fetch), so missing blobs are dropped here and the runtime filters the id lists against them.
+		const entries = await Fn.mapAwait(ids, async id => {
+			const blob = await getOrSet(cache.blobs, id, () => loadOrderBlob(player.shard, id));
```
laverdet: Most transactions don't change every tick. An idea is to put the ids in a zset by game time of last update. You can use that to populate the cache from the previous tick. Or you could do a channel which publishes updates.

packages/xxscreeps/mods/market/driver.ts
```diff
@@ -49,3 +54,43 @@ hooks.register('runnerConnector', async player => {
...
+			return blob && ([ id, blob ] as const);
+		});
+		payload.orders = { active, mine, blobs: Fn.fromEntries(Fn.filter(entries)) };
```
laverdet: It goes without saying but this should only send updated blobs, not the whole market book each tick to each runtime.

packages/xxscreeps/mods/market/main.ts
```diff
@@ -0,0 +1,26 @@
...
+	}
+}
+
+// The market's global pass. It runs once per tick in `main` (so order/credit mutations are
+// serialized): drain this tick's player-global market intents, apply them, then recompute every
+// order's activity. The book is snapshotted before the intents run, so an order created this tick is
+// not maintained until next tick (it is written inactive and first activates then).
+registerShardTickProcessor(async (shard, time) => {
+	const orders = await loadOrders(shard);
+	for (const { userId, named } of await acquireNamedIntentsForTick(shard, time)) {
```
laverdet: Shouldn't this be promise.all?

packages/xxscreeps/mods/market/game.ts
```diff
@@ -43,5 +44,9 @@ hooks.register('gameInitializer', (game, data) => {
 	if (data?.transactions) {
 		transactions = new Transactions(data.transactions);
 	}
-	game.market = new Market(game, transactions);
+	game.market = new Market(
+		game,
+		transactions,
+		data?.orders && new Orders(data.orders),
```
laverdet: Maybe just make Orders accept `undefined` and remove the cascading `?.` everywhere.

packages/xxscreeps/mods/market/market.ts
```diff
@@ -1,27 +1,98 @@
...
 	get incomingTransactions() { return this.#transactions?.incoming ?? []; }
 	get outgoingTransactions() { return this.#transactions?.outgoing ?? []; }
+
+	getAllOrders(filter?: ((order: Order) => boolean) | Partial<Order>) {
+		const active = this.#orders?.active ?? [];
+		if (filter === undefined) {
+			return [ ...active ];
+		}
+		const predicate = typeof filter === 'function' ? filter :
+			(order: Order) => Object.entries(filter).every(([ key, value ]) => order[key as keyof Order] === value);
```
laverdet: You can do a lot better by folding a predicate from Object.entries(). Certainly, don't invoke Object.entries once for each element that's wild.

packages/xxscreeps/mods/market/market.ts
```diff
@@ -1,27 +1,98 @@
...
+	 */
+	createOrder(options: CreateOrderOptions) {
+		const { type, resourceType, price, roomName } = options;
+		const totalAmount = Math.trunc(options.totalAmount);
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+		if (type !== C.ORDER_BUY && type !== C.ORDER_SELL) {
+			return C.ERR_INVALID_ARGS;
+		}
+		// A negative `totalAmount` passes these checks and returns OK; the shard pass rejects it and
+		// no order is created.
```
laverdet: Is there a reason for this behavior or is it some whacky screeps behavior that got copy/pasted? I think it'd be better just to reject the intent. I'd rather if you can combine the checks as much as possible between the game runtime and the processor.

packages/xxscreeps/mods/market/model.ts
```diff
@@ -84,3 +91,187 @@ export async function recordTransaction(shard: Shard, senderId: string, recipien
...
+		shard.data.del(orderBlobKey(id)),
+		shard.data.sRem(allOrdersKey, [ id ]),
+		shard.data.sRem(activeOrdersKey, [ id ]),
+		shard.data.sRem(userOrdersKey(order['#user']), [ id ]),
+	]);
+}
+
+// User credit balance in millicredits; `Game.market.credits` divides by 1000.
+export function loadMoney(shard: Shard, userId: string) {
+	return shard.db.data.hGet(User.infoKey(userId), moneyField).then(money => Number(money) || 0);
```
laverdet: Don't use `then`

packages/xxscreeps/mods/market/model.ts
```diff
@@ -84,3 +91,187 @@ export async function recordTransaction(shard: Shard, senderId: string, recipien
...
+	if (type !== C.ORDER_BUY && type !== C.ORDER_SELL) {
+		return;
+	}
+	if (!(C.RESOURCES_ALL as string[]).includes(resourceType) || !(price > 0) || !(totalAmount > 0)) {
+		return;
+	}
+	// A non-intershard order requires a terminal the player owns in `roomName`. Full initialization
+	// is needed so the room's objects are instantiated and `terminal` resolves. `roomName` is
+	// unvalidated, so the room may simply not exist — a failed load rejects the intent like any other.
+	const room = await shard.loadRoom(roomName).catch(() => undefined);
```
laverdet: I'm getting the sense here that this means that this would be better as an intent on the terminal itself. Obviously pixels, cpu, etc would still need a named intent.

packages/xxscreeps/mods/market/model.ts
```diff
@@ -84,3 +91,187 @@ export async function recordTransaction(shard: Shard, senderId: string, recipien
...
+// Only orders whose state actually changed are rewritten.
+export async function runOrderMaintenance(shard: Shard, orders: Order[]) {
+	const now = Date.now();
+	const rooms = new Map<string, Promise<Room>>();
+	const loadRoomOnce = (roomName: string) => getOrSet(rooms, roomName, () => shard.loadRoom(roomName));
+	for (const order of orders) {
+		if (now - order.createdTimestamp > C.MARKET_ORDER_LIFE_TIME) {
+			const refund = Math.floor(order.remainingAmount * order['#price'] * C.MARKET_FEE);
+			if (refund > 0) {
+				await shard.db.data.hincrBy(User.infoKey(order['#user']), moneyField, refund);
```
laverdet: Promise.all

packages/xxscreeps/mods/market/model.ts
```diff
@@ -84,3 +91,187 @@ export async function recordTransaction(shard: Shard, senderId: string, recipien
...
+			}
+			await removeOrder(shard, order);
+			continue;
+		}
+		const { terminal } = await loadRoomOnce(order.roomName);
```
laverdet: I don't want to indiscriminately load rooms in this phase. What about putting the order ids on the terminal and then running this stuff in the tick processor?

misterwise:
Reworked along both structural lines and rebased onto current main:

- `createOrder` is now an intent on the terminal. The runtime method and the processor share `checkOrderParams`, and the vanilla negative-amount quirk now rejects with `ERR_INVALID_ARGS`. The listing fee is a debit-then-refund (`chargeListingFee`), so concurrent room processors can't overdraw a balance. Since the intent slot is unique per (object, action), same-tick orders accumulate into one batched intent — vanilla allows several `createOrder` calls from one room in a tick.
- Order ids anchor on the terminal (a `'#orderIds'` schema field), and the terminal's tick processor maintains activation/amount/expiry from local state, keeping its room ticking while orders exist. The shard pass no longer loads rooms — it's an expiry backstop over `market/orders`, now a zset scored by creation time so it reads only due ids.
- The book ships as deltas. Writers publish changed ids on a `market/orders` channel; the runner keeps a cross-tick blob cache invalidated by it plus a per-runtime dirty set, and each runtime folds only blobs it hasn't seen into its retained book. Shard scoping is dropped, `Orders` takes `undefined`, and the `Object.entries` fold / `.then` / `Promise.all` points are fixed.
- The named-intent transport is reverted out of the net diff — `createOrder` was its only consumer once it moved onto the terminal.

The rebase also surfaced a module-evaluation-order hazard: building a schema reader at module scope inside the runtime graph can close the `resourceEnumFormat` extension path before another mod registers into it. `order.ts` builds its runtime reader lazily and `model.ts` builds eagerly behind `xxscreeps:mods/game`; upstream `transaction.ts` has the same latent pattern and survives by import order.

Two gaps called out in the body: a destroyed terminal's orders stop being maintained (stale `amount`/`active`) until the age-out backstop removes them, and `credits` still rides a per-tick `hGet` in the connector — happy to move it onto a channel like transactions/orders if you'd rather.

laverdet:
I've been working on paring this down. My first pass is 0bf9351054c09e6b8acd2bbf4404d4fc2fcbd7a6. I think the scope of this PR was too large and complex so I couldn't really review it effectively before things got out of control.

The first problem I see happening a lot is Claude treats existing code as a matter of fact. I've made a bunch of tools in this project but if the tools aren't effective then they should always be revisited.

I added `hIncrByEx` (based on [incrByEx](https://redis.io/docs/latest/commands/increx/)) in my patch. This missing feature resulted in the two-pass `chargeListingFee` implementation (ok, fine) and the serial processing of orders (not fine). I'm really picky here because on a live server the market has the potential to slow down the whole game.

Another example is a missing feature in `deterministicClockForTesting` which led to a finnicky test case which manually updates `createdTimestamp` but doesn't update the zSet. The correct solution there is just to add an increment feature to that testing harness.

On comments: for Screeps player API functions I'm not writing them by myself, they're all copy/pasted from https://docs.screeps.com/api/. The intent is that they can be exported for consumption by the player. I had this working a long time ago but it's since been removed. I think this could be more clear with a `@public` tag and URL ref.

On too much influence from Screeps: Order's `active` field is simply a view on `amount > 0`, it does not need to be its own schema field. `roomName` should not be a string, we have a format for that. `type` should also not be a string since it can only be two states and the API type is a view on top of that. The terminal order processing is maybe another example. The insight here is that the room's processor is only concerned with what's happening in the room. So for game resources the room processor should only update `amount`, but for expiry that is actually a concern for the shard processor.

I wasn't super happy with the way transaction blobs ended up in that previous PR but I didn't bother because the impact is pretty low. But another thing I've noticed is that there's no "disgust factor" when tokenmaxxing. Like, if a human is committing lousy code we know it's lousy but "it's only this one time" so it's ok. But with AI, lousy code tends beget more lousy code.

Anyway the next pass will be for the game-visible order book. I think that'll end up being a smaller diff. Thanks for your work here, it's still helpful despite the problems.

misterwise:
> I've been working on paring this down. My first pass is [0bf9351](https://github.com/laverdet/xxscreeps/commit/0bf9351054c09e6b8acd2bbf4404d4fc2fcbd7a6). I think the scope of this PR was too large and complex so I couldn't really review it effectively before things got out of control.
> 
> The first problem I see happening a lot is Claude treats existing code as a matter of fact. I've made a bunch of tools in this project but if the tools aren't effective then they should always be revisited.
> 
> I added `hIncrByEx` (based on [incrByEx](https://redis.io/docs/latest/commands/increx/)) in my patch. This missing feature resulted in the two-pass `chargeListingFee` implementation (ok, fine) and the serial processing of orders (not fine). I'm really picky here because on a live server the market has the potential to slow down the whole game.
> 
> Another example is a missing feature in `deterministicClockForTesting` which led to a finnicky test case which manually updates `createdTimestamp` but doesn't update the zSet. The correct solution there is just to add an increment feature to that testing harness.
> 
> On comments: for Screeps player API functions I'm not writing them by myself, they're all copy/pasted from https://docs.screeps.com/api/. The intent is that they can be exported for consumption by the player. I had this working a long time ago but it's since been removed. I think this could be more clear with a `@public` tag and URL ref.
> 
> On too much influence from Screeps: Order's `active` field is simply a view on `amount > 0`, it does not need to be its own schema field. `roomName` should not be a string, we have a format for that. `type` should also not be a string since it can only be two states and the API type is a view on top of that. The terminal order processing is maybe another example. The insight here is that the room's processor is only concerned with what's happening in the room. So for game resources the room processor should only update `amount`, but for expiry that is actually a concern for the shard processor.
> 
> I wasn't super happy with the way transaction blobs ended up in that previous PR but I didn't bother because the impact is pretty low. But another thing I've noticed is that there's no "disgust factor" when tokenmaxxing. Like, if a human is committing lousy code we know it's lousy but "it's only this one time" so it's ok. But with AI, lousy code tends beget more lousy code.
> 
> Anyway the next pass will be for the game-visible order book. I think that'll end up being a smaller diff. Thanks for your work here, it's still helpful despite the problems.

This is very helpful feedback, thank you. It's not just Claude, I've also tended to work from the mindset of existing code being matter of fact, so I engineered some of that rigidity into my own framework for evaluation. I'll try to take the same lens to my other open PRs and see what I can do about improving the quality of the code.

PR #301, opened 2026-07-03
creep: complete the Tombstone.creep deceased-creep surface

misterwise:
## Summary
`Tombstone.creep` exposed only a partial creep — `body` as raw part-type strings and a `store` that aliased the tombstone's own store — and inherited getters misfire on the bufferless fake creep (`spawning` reads `#ageTime === 0` → true; `saying` is gated on `time === Game.time`). Widen the getter to `{ type, hits }` body parts, `spawning: false`, and an empty `store` sized to the body's carry capacity. `saying` is gated at bury time to the public death-tick message — a private or stale saying isn't memorialized — shrinking the stored `#creep` saying from a struct to a string; the getter and backend renderer forward it raw.

## Validation
screeps-ok tombstone parity (`18.1-tombstone.test.ts`, 18/18): closed TOMBSTONE-006/-010/-016/-017/-018; -012 (±1-tolerant ticksToLive) stays green.

### DIFF
--- a/packages/xxscreeps/mods/creep/tombstone.ts
+++ b/packages/xxscreeps/mods/creep/tombstone.ts
@@ -1,5 +1,6 @@
 import type { ResourceType } from 'xxscreeps/mods/resource/resource.js';
 import * as Id from 'xxscreeps/engine/schema/id.js';
+import { Fn } from 'xxscreeps/functional/fn.js';
 import * as C from 'xxscreeps/game/constants/index.js';
 import { Game } from 'xxscreeps/game/index.js';
 import { RoomObject, create as createObject, format as objectFormat, requiredExpiryTime } from 'xxscreeps/game/object.js';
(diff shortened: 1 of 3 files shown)


packages/xxscreeps/mods/creep/tombstone.ts
```diff
@@ -105,11 +105,12 @@ export function buryCreep(creep: Creep, rate = C.CREEP_CORPSE_RATE) {
 		for (const [ type, amount ] of creep.store['#entries']()) deposit(type, amount);
 	}
 
+	const saying = creep['#saying'];
 	tombstone['#creep'] = {
 		body: creep.body.map(bodyPart => bodyPart.type),
 		id: creep.id,
 		name: creep.name,
-		saying: creep['#saying'],
+		saying: saying?.isPublic && saying.time === Game.time ? saying.message : undefined,
```
laverdet, line 113: Should `saying.time == Game.time` be removed? I don't see why that should be there.

misterwise: We keep actions around for 10 ticks for the renderer, so this matches creep.ts:103-108 which ensures that only the actual tick of saying is preserved in a tombstone (avoiding a 10 tick window of saying stored)

PR #302, opened 2026-07-03
schema: store layout archives in the database

bastianh:
## What

Schema archives — the layout descriptions `makeUpgrader` needs to read blobs written by older schema versions — were written as files into `schemaArchive` at build time and read back lazily from disk. Deployments had to carry that directory around (e.g. bake it into images and seed it onto volumes).

They now live in the keyval provider itself, under `schema/{name}` hash keys (field = version, value = archive source). `Database.connect` / `Shard.connectWith` call `initializeSchemaArchive`, which preloads all archived versions into memory — keeping the upgrader synchronous — and stores the current version of each package with `hSet(..., { if: 'NX' })`, so concurrently starting services can race safely.

With this, every persistent store is self-describing: a database (redis or local) always carries the schemas needed to read its own blobs, which also makes backups and moving worlds between storage backends straightforward.

The generated `.ksy` Kaitai documentation files are no longer written.

## Migration (second commit — optional)

The second commit imports archives found in the legacy `schemaArchive` directory into the database on connect (idempotent), so existing worlds keep their upgrade history. Since the magic-number change in 05be3b2e already invalidated existing blobs and effectively required a world reset anyway, this commit can be dropped if you'd rather just expect fresh worlds.

## Testing

- `tsc -b` clean; eslint shows only the two warnings that already exist on main in this file.
- Two-process end-to-end test: process A declares a schema (v1), connects a fresh local store, and writes the archive plus a blob; process B declares a changed schema with the same name, connects, and verifies that upgrading fails before `initializeSchemaArchive`, succeeds after (existing field preserved, new field defaulted), that an upgraded blob is stable, and that both versions end up archived in the database. The same test was repeated with the old version's archive provided only as a file in the legacy directory (old filename encoding) to cover the seeding path.

### DIFF
--- a/.changeset/schema-archives-in-database.md
+++ b/.changeset/schema-archives-in-database.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Store schema archives in the database instead of the filesystem. Every persistent store is now self-describing: previous schema versions needed to upgrade old blobs are preloaded from `schema/{name}` hash keys on connect, and archives found in the legacy `schemaArchive` directory are imported automatically. The generated `.ksy` Kaitai documentation files are no longer written.
--- a/.changeset/schema-archives-in-database.md
+++ b/.changeset/schema-archives-in-database.md
@@ -2,4 +2,4 @@
 "xxscreeps": patch
 ---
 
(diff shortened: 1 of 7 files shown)


laverdet:
I took a different approach of loading them on demand instead of pulling everything when the service starts. I'm also stashing some metadata so we could purge old schemas as needed. With the writeback behavior added a while ago hopefully old blobs would be rare after a server upgrade itself.

PR #303, opened 2026-07-03
Update User stats / Map stats

bastianh:
# PR Summary: Update User/Map Statistics Collection and Endpoints

Implements tracking, storage, and API endpoints for 7 classic gameplay stats (*harvest, upgrade, build/repair, spawn/renew, creeps produced, creeps lost, power processed*) as bucketed time series.

---

### Commit 1: Stats Infrastructure & Base Endpoints
* **Storage:** Saves aggregated per-user totals in `db.data` (cross-shard) and per-room series in `shard.data`. Prunes old buckets deterministically on write.
* **Processing:** `addStat` coalesces room-tick contributions at processor sites and flushes them via a queued task. Added a teardown hook for removed users.
* **API:** Implements base endpoints:
  * `GET /api/user/stats` (aggregated profile totals)
  * `GET /api/user/overview` (totals + room punchcards)
  * `GET /api/game/room-overview` (owner, room punchcards, totals/maxima)

### Commit 2: User-Splitting & Map Layers
* **Segmentation:** Keys per-room series by `(room, user)` instead of aggregating, and tracks each room's contributors. This enables personalized overviews and world map stat layers.
* **API:** Adds `POST /api/game/map-stats` to return windowed contributor values per room (highest first, dropping aged-out users), plus layer max and user lookups.

### Commit 3: Client Compatibility Fixes
Aligns response structures with the legacy Angular client (`screeps-client`):
* **`user/overview`:** Nests stats under the shard level; defaults `statName` to `energyHarvested`.
* **`user/stats`:** Honors the `id` query parameter to allow viewing any user's profile.
* **`room-overview`:** Returns `owner.badge` as `null` instead of `{}` when missing so the client correctly hides it.


bastianh:
Rebased the PR on current main, did some style and bug fixing. Will also do some more testing...

bastianh:
@laverdet I really struggle a bit how to handle the stats in a performant way .. when zoomed out the map the screeps client requests like 4500 rooms in one call.. thats a lot of rooms to load .... in that zoom it only fetches minerals0 or the owner .. but with curl it also responds with the 7 other stats like `energyHarvested8`

Maybe the solution is to have a optional interface like the messages that allows to plug a database for stats .. maybe something like
```ts

export interface StatStore {
      writeStats(shard: Shard, userId: string, deltas: StatDeltas, now: number): Promise<void>;
      writeRoomStats(shard: Shard, room: string, userId: string, deltas: StatDeltas, now: number): Promise<void>;
      pruneRoomContributors(shard: Shard, room: string, now: number): Promise<void>;
      readTotals(shard: Shard, userId: string, interval: StatInterval, now: number): Promise<StatTotals>;
      readRoomTotals(shard: Shard, room: string, userId: string, interval: StatInterval, now: number): Promise<StatTotals>;
      readRoomPunchcard(shard: Shard, room: string, userId: string, interval: StatInterval, stat: StatName, now: number): Promise<number[]>;
      readRoomLayer(shard: Shard, room: string, interval: StatInterval, stat: StatName, now: number): Promise<RoomStatContribution[]>;
      removeAllForUser(shard: Shard, userId: string): Promise<void>;
}


```

laverdet:
I think you'll have to throw most of this away, but it's a good way to think about ups and downs.

Per-user & per-room stats is a good approach but I think you're doing it wrong? The client never asks for stats for users one at a time in a room (or does it, and I don't know?). The cool thing about per-user per-room stats would be that you could just write once and then aggregate on read. But you're writing both so I don't see the point?

I think the way to do this is to store per-user, per-room stats aggregated directly on the room blob. We're already writing the blob so this is essentially free. Then once per 5-15 minutes these stats are dumped into redis buckets.

Also the diff needs to be split. The scope here is too large.

### Diff 1:

`api/game/map-stats` mod hook. Map stats should remain a core backend endpoint, but mods need a way to hook into it. We already have backend hooks, so add one. Factor controller info, and deposit map-stats out into their respective mods.

### Diff 2:

Processor stats integration. Add a `incrementRoomStat?: (...whatever) => void` function on `ProcessorContext`. A new `mods/meta/stats` mod would define that. I'll be honest I don't remember why ProcessorContext and RoomProcessor are separate. So maybe just merge those again. Then you'd patch `RoomProcessor.prototype.incrementRoomStat` in the stats mod. You could imagine a different stats mod do the same for a different storage backend.

Anyway, this would just store the stats directly on the room blob. I imagine this looks something like:

```
`struct userRoomStats { wallTime: number; userId: id, stat: enum; amount: int32 }`
`struct room { roomStatBucket: userRoomStats[] }`
```

### Diff 3:

Redis and map-stats integration. All that remains is aggregation logic. Every so often you'd take the stats out of the room and dump them into redis. Essentially your first bucket at time T-0 lives only on the room blobs.

bastianh:
thanks for your feedback .. I'll look into it.

Actually there are per user/room stats calls in the map .. in center rooms multiple users could harvest energy in the same room in a time bucket .. then the map view shows multiple user badges for that room .. but I'm not sure if I ever saw that on the live map .)

bastianh:
created #324 and #325 for diff 1 and 2.

A few notes on how I'd build diff 3 on top of that:

- Account-level totals currently aggregate cross-shard via db.data (like GCL). I'd keep that: the periodic dump would hincrBy both the per-(room, user) buckets in shard storage and the per-user buckets in db.data.
- On the blob I coalesce to one entry per (user, stat) with a single bucket timestamp on the room, instead of wallTime per entry, otherwise the array grows with event count between dumps rather than with users × stats. Per-entry times would only buy precision if entries weren't coalesced at all.
- Flush rule: a room flushes when the current time has crossed into a different bucket of the finest interval (8 minutes) than the room's bucket timestamp, crediting the redis bucket the timestamp falls in. That bounds the attribution error to one bucket at the finest resolution, keeps totals exact, and naturally lands at your "once per 5–15 minutes" cadence — no extra scheduling needed, it piggybacks on the room tick.

laverdet:
Yes, one aggregate bucket per-user per-room. The main insight is that we're already writing a room, so writing some stats to the blob is basically free. Aggregating over 5 minutes (or 8 or whatever) saves thousands of stats writes. You should also add some jitter to the local room buckets to avoid the thundering herd problem after a tick's bucket expires.

PR #304, opened 2026-07-04
Fix double-unsubscribe crash in Subscription.disconnect

bastianh:
A subscription's disconnect could fire twice (explicit unsubscribe message plus socket-close cleanup), and the second call crashed RedisPubSubProvider.unsubscribe with an uncaught TypeError, killing the backend's socket handling while the engine kept ticking.

### DIFF
--- a/.changeset/pubsub-double-unsubscribe.md
+++ b/.changeset/pubsub-double-unsubscribe.md
@@ -2,4 +2,4 @@
 "xxscreeps": patch
 ---
 
-Fix a crash (`Cannot read properties of undefined (reading 'subscribers')`) in `RedisPubSubProvider.unsubscribe` that could take down the backend's socket handling. When several subscribers share one channel key (e.g. a room's shared and per-socket listeners), the previous size-based teardown could delete the whole delegate entry while a live subscriber remained — or race with an in-flight subscribe — leaving a later unsubscribe to dereference a missing entry. Unsubscribe now removes only its own subscriber, is a no-op if the entry or subscriber is already gone, and drops the shared Redis subscription only once the last subscriber leaves.
+Fix a backend socket crash
--- a/.changeset/pubsub-double-unsubscribe.md
+++ b/.changeset/pubsub-double-unsubscribe.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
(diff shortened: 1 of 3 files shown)


laverdet:
It's ok to add a check to the subscription but it should be a throw, not a fall through. If there's a double-disconnect then we should find the source and fix that. A subscription is a resource and there should be only one owner. If there's more than one disconnect it means that more than one branch of code thinks it owns the resource. So then one block of code would disconnect and another would start failing in different ways.

bastianh:
So the reason what happend according to claude was that the client sends an explicit unsubscribe message in backend/socket.ts and does a second unsubscribe because the connection was closed.

The second call reaches RedisPubSubProvider.unsubscribe (packages/redis/pubsub.ts:89) after the entry was already deleted from subscribersByKey, throwing TypeError: Cannot read properties of undefined (reading 'subscribers')

laverdet:
backend/socket.ts has a `subscriptions` map which stores client subscriptions by name. A client unsubscribing will unsubscribe the given "socket" (client resource) by name. Closing the socket closes all unclosed sockets. I'm not saying that there is not a bug but it needs to be understood before we plaster over it.

bastianh:
I will revert my changes in my branch and do dome more debugging

bastianh:
Ok .. I could reproduce the crash happening when reloading my client multiple times.. the description what happens from claude is now: 

Fixes a crash in RedisPubSubProvider.unsubscribe when multiple subscribers share one channel key (e.g. a room's shared listener plus per-socket ones). The old teardown deleted the whole delegate entry as soon as it saw size === 1, without checking the caller was actually the last live subscriber — so a fast client reconnect (unsubscribe + resubscribe racing the await sSubscribe gap) could drop the entry while a subscriber was still attached, and that orphan's later unsubscribe then threw Cannot read properties of undefined (reading 'subscribers'), taking down socket handling. Now unsubscribe removes only its own subscriber, no-ops if it's already gone, and drops the shared subscription only once the last subscriber leaves (mirrored in the local provider).



> Confirmed — deterministically. The harness drives the exact fast-reconnect sequence with no Redis, running both implementations:
> 
> OLD (size===1)             -> THREW: TypeError: Cannot read properties of undefined (reading 'subscribers')
> NEW (membership-checked)   -> OK, no crash
> 
> The old logic throws the exact production error; the fix survives the same sequence. That proves the mechanism.
> 
> What it did (faithful transcription of redis/pubsub.ts):
> 1. subscribe(K) — first subscriber, its sSubscribe held in-flight (the async gap the real code has).
> 2. subscribe(K) — second subscriber to the same key; entry already exists, so it resolves immediately and adds itself.
> 3. unsubscribe(B) — the reconnecting client's old socket leaves first. Old code sees size === 1 and deletes the whole delegate entry.
> 4. First subscribe's sSubscribe finally resolves → A adds itself to the now-orphaned set.
> 5. unsubscribe(A) → get(K) is undefined → TypeError. 💥
> 
> So: fast reconnect is the trigger, and the async await sSubscribe gap is the window — your 50ms ticks didn't cause it, they just widened how often that window got hit. The size === 1 teardown was never order-safe; the fix removes only its own subscriber and no-ops on a missing entry, which is why step 5 becomes harmless.

PR #309, opened 2026-07-07
Move sector geometry out of the game core into a sector mod

misterwise:
#293 landed the sector fields with `// TODO: mods/sector` ("Nothing about sectors are intrinsic to rooms; all this logic should be a mod"). This moves them: `mods/sector` registers `sectors`/`sectorControl` into `RoomIntrinsics` through the `Schema` extension point, `game/room/sector.ts` relocates whole, and `GameMap['#sectors']` becomes the mod-owned `iterateSectors(world)`. Deposit and powerbank pick the mod up as a dependency. The registration rides the `schema` provide, which every mod side-effect aggregate evaluates first, so it lands before the World reader is built on every entry path; `driver/pathfinder/profile.ts` (the one direct `World` consumer with no mod bootstrap) now imports it explicitly, keeping the pathfinder PGO run reading server-written terrain blobs.

The built World format is byte-identical — the schema version hash is `ee5f783c` on main and on this branch.

## Validation
- Full suite 484/484 on the rebased tip, and 485/485 with `--private-transform=isolated-vm`; the sector describes run relocated under `mods/sector`.
- Schema version hash compared on upstream/main vs branch builds (identical, re-checked after rebasing over the terminal/market split); `xxscreeps import` world ran ~340 live ticks with a bot on the isolated sandbox — no errors, deposit seeded its sector schedule at init.
- eslint on changed files: no new warnings.


laverdet:
I removed this line. It does not appear to be needed?

```js
// Mod schema registrants must evaluate before the World reader in 'game/map.js' is built.
import 'xxscreeps:mods/schema';
```

https://github.com/laverdet/xxscreeps/pull/309/changes/d8d0242e732db5405619de0efe5e6e138b50e7fe#diff-4469262a2a9ec0f91ae54ad4f195ced69c7f5ed794cb307c080a41545f8b142cR2

misterwise:
It is needed, but the failure only shows in the pathfinder workflow, which is path-filtered and didn't run on this merge. `pathfinder.yml` writes the terrain fixture with a full-mod `xxscreeps import`, so the blob's version hash includes the mod registrants; `profile.js` runs without a mod bootstrap, so its `World` reader is built from the base `RoomIntrinsics` only:

```
Error: Unexpected blob version [3999234108]. Expected 3789116416.
    at new World (dist/game/map.js)
```

Repro'd on `194f9c15` with the workflow's exact steps. Every other entry point gets this bootstrap implicitly — the generated aggregates all lead with `import "xxscreeps:mods/schema"`, and the sandboxes prepend it — profile.js is just the one bare-node entry that has to say it explicitly. Happy to send the one-liner restoring it, with a comment that names the actual failure this time. (The TODO you kept in map.ts — building the World reader lazily instead of at module load — is the version of this that removes the import for good.)

PR #311, opened 2026-07-08
invader: deploy canonical bunker templates with reward containers

misterwise:
Replaces the stub stronghold layout with the five bunker templates from `@screeps/common/lib/strongholds.js` (ISC). On deploy the core clears each template tile the way vanilla does — creeps die, construction sites drop half their progress as energy, player-buildable structures are destroyed — then spawns its template's peers: ramparts at `STRONGHOLD_RAMPART_HITS[rewardLevel]`, towers at full energy, and containers rolling loot from the weighted reward table.

- `#templateName` is stored on the core at placement rather than derived from its level, so layout selection stays independent of core level and deploy just reads it.
- `#strongholdId` groups a stronghold's structures under one id. Deploy only writes it today; the defense/behavior slice (next up) queries by it, so it rides this migration alongside `#collapseTime`. That slice also wants vanilla's rampart `hitsTarget` (stamped at deploy, retuned by antinuke), which will ratchet the Structure archive again regardless — happy to move `#strongholdId` into that PR instead if you'd rather the field land with its first reader.
- Deployed peers surface `EFFECT_COLLAPSE_TIMER` via an `extend(Structure, …)` effects getter in the invader mod, which owns `#collapseTime`; the core's and controller's own getters shadow it.
- Deployed ramparts carry hits above `hitsMax`, which reads 0 in a room the invader doesn't control; live servers show the same.
- Templates reach ±3 tiles from the core, so deploy assumes the core sits at least 3 tiles inside the room edge; placement will enforce that margin when it lands.

## Validation

- 479/479 unit tests; eslint clean on changed files.
- Live smoke in the worktree: bunker3 deployed through the real processor (35 structures, container loot summing to the 10000 target density), server restarted with the stronghold intact, and a shortened collapse timer removed every peer.

### DIFF
--- a/packages/xxscreeps/mods/invader/strongholds.ts
+++ b/packages/xxscreeps/mods/invader/strongholds.ts
@@ -8,9 +8,9 @@ import {
 	RESOURCE_PURIFIER, RESOURCE_REDUCTANT, RESOURCE_UTRIUM_BAR, RESOURCE_ZYNTHIUM_BAR,
 } from 'xxscreeps/mods/modern/factory/constants.js';
 
-// Stronghold layout templates and loot tables ported from @screeps/common (lib/strongholds.js),
-// which is ISC-licensed. Each template's `rewardLevel` matches its bunker number. The core's own
-// template entry is omitted — deploy spawns peers around an existing core.
+// Stronghold layout templates and loot tables ported from @screeps/common (lib/strongholds.js).
+// Each template's `rewardLevel` matches its bunker number. The core's own template entry is omitted
+// — deploy spawns peers around an existing core.
 
 export interface StrongholdStructure {
 	type: typeof C.STRUCTURE_RAMPART | typeof C.STRUCTURE_TOWER | typeof C.STRUCTURE_CONTAINER | typeof C.STRUCTURE_ROAD;
--- a/packages/xxscreeps/mods/invader/strongholds.ts
+++ b/packages/xxscreeps/mods/invader/strongholds.ts
@@ -1,0 +1,309 @@
+import type { ResourceType } from 'xxscreeps/mods/classic/resource/resource.js';
+import { mappedNumericComparator } from 'xxscreeps/functional/comparator.js';
+import { Fn } from 'xxscreeps/functional/fn.js';
+import * as C from 'xxscreeps/game/constants/index.js';
+import {
+	RESOURCE_BATTERY, RESOURCE_COMPOSITE, RESOURCE_CRYSTAL, RESOURCE_GHODIUM_MELT,
+	RESOURCE_KEANIUM_BAR, RESOURCE_LEMERGIUM_BAR, RESOURCE_LIQUID, RESOURCE_OXIDANT,
+	RESOURCE_PURIFIER, RESOURCE_REDUCTANT, RESOURCE_UTRIUM_BAR, RESOURCE_ZYNTHIUM_BAR,
+} from 'xxscreeps/mods/modern/factory/constants.js';
+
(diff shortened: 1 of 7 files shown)


packages/xxscreeps/mods/invader/strongholds.ts
```diff
@@ -0,0 +1,309 @@
...
+			{ type: C.STRUCTURE_RAMPART, dx: 0, dy: 2 },
+			{ type: C.STRUCTURE_CONTAINER, dx: 0, dy: -2 },
+			{ type: C.STRUCTURE_ROAD, dx: 0, dy: -2 },
+			{ type: C.STRUCTURE_RAMPART, dx: 0, dy: -2 },
+		],
+	},
+	bunker5: {
+		rewardLevel: 5,
+		structures: [
+			{ type: C.STRUCTURE_RAMPART, dx: 0, dy: 0 },
```
laverdet: If this was something I was writing from scratch I'd probably just draw a little picture and parse it out. It'd be fewer lines and easier to understand.

```
  ..... 
 .c.t.c.
 ..t.t..
 ...x...
 ..t.t..
 .c.t.c.
  ..... 
```

misterwise: Love this idea, I'll work it in.

packages/xxscreeps/mods/invader/strongholds.ts
```diff
@@ -0,0 +1,135 @@
...
+	weights: Record<string, number>,
+	targetDensity: number,
+	itemsLimit: number,
+): Generator<[ ResourceType, number ]> {
+	const picks = Fn.pipe(
+		Object.entries(weights),
+		$$ => Fn.map($$, entry => ({ entry, order: Math.random() })),
+		$$ => [ ...$$ ],
+	);
+	picks.sort(mappedNumericComparator(pick => pick.order));
```
laverdet: Use lodash collection shuffle or do a fisher-yates.

packages/xxscreeps/mods/invader/strongholds.ts
```diff
@@ -0,0 +1,135 @@
...
+		...x...
+		..t.t..
+		.c.t.c.
+		 .....
+	`),
+} satisfies Record<string, StrongholdTemplate>;
+
+// Weighted resource table for the loot in a stronghold's containers, keyed by reward level into
+// `containerAmounts`.
+export const containerRewards: Record<string, number> = {
```
laverdet: This is exported and then only used once, when it's passed back as an argument to calcRewards? Why all the indirection?

laverdet:
On licensing: Screeps & Artem are already attributed in [LICENSE](https://github.com/laverdet/xxscreeps/blob/main/LICENSE). My understanding is that's all we need.

packages/xxscreeps/mods/invader/strongholds.ts
misterwise: This looks like it got through a rework without being properly audited and removed.

packages/xxscreeps/mods/invader/strongholds.ts
misterwise: Swapped in an in-place fisher-yates: 3bea7629

packages/xxscreeps/mods/invader/strongholds.ts
misterwise: Folded the tables and item limit into calcReward — it takes just the reward level now: 3bea7629

laverdet:
Looking at `strongholdId` in Screeps it seems like the plan here is to tag damaged roads with the core id? I think you should remove that from the structure format and store roads owned by the core on the core. `ownedNeutralStructureIds: vector(Id.format)`

PR #313, opened 2026-07-09
schema: upgrade persisted blobs to the current format on load

bastianh:
Is there a reason why only room storage blobs are updated?

Any schema change touching a blob's layout made terrain, flags, power creep rosters, market transactions, user code, and room visuals throw `Unexpected blob version` on load — they were read with a plain reader and never routed through the upgrader (rooms already were).

Add `upgrade` to `makeReaderAndWriter` and apply it host-side at each database read site — the raw-blob loaders that feed both the backend parse and the runtime payload — so old blobs migrate transparently and the sandbox keeps receiving an already-current blob. Power creep keeps the raw blob for its compare-and-swap and upgrades only for parsing.

### DIFF
--- a/.changeset/schema-blob-upgrade.md
+++ b/.changeset/schema-blob-upgrade.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Upgrade all persisted schema blobs to the current format version on load.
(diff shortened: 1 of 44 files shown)


laverdet:
- removed visual upgrader (no need)
- made these functions async
- write modified buffer back to database (possibility of race condition but probably not)

PR #314, opened 2026-07-11
backend: trust X-Forwarded-* headers via configurable proxy settings

bastianh:
For generating correct urls the server needs to know the external urls and whether https was used behind a reverse proxy 

Add backend.proxy, backend.proxyIpHeader, and backend.maxIpsCount config options so the Koa app can honor X-Forwarded-Proto/Host/For when running behind a trusted reverse proxy. This makes ctx.secure/ctx.protocol report https correctly when TLS is terminated upstream (e.g. for OAuth redirect URIs).

### DIFF
--- a/.changeset/backend-trusted-proxy-config.md
+++ b/.changeset/backend-trusted-proxy-config.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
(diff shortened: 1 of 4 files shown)


laverdet:
I took a different approach. `maxIpsCount` is a more strict `forwardedCount` which will fail if the proxy topology doesn't match what is configured. I didn't implement `x-real-ip` because there are better solutions.

PR #316, opened 2026-07-12
Fix local keyval/blob save-load regressions

bastianh:
Two regressions in the local (file-backed) storage provider, both only reachable on the on-disk save → reload path.

**Reviver drops plain objects.** A `JSON.parse` reviver visits nested values before their parents, so it sees a serialized map's `$` payload object (and any raw `null`) as an untagged object. The old branch had no fall-through for that case and returned `undefined`, which deletes the property — the wrapper's `$` was stripped before the `'map'` case ran, so `new Map(Object.entries(value.$))` threw on load. `null` separately hit `null['#']`. The reviver now passes untagged objects, `null`, arrays, and primitives through unchanged.

**Blob save can't unlink a never-flushed deletion.** A blob created and deleted between two saves never reaches disk, so `save()`'s `fs.unlink` threw `ENOENT`. It now tolerates `ENOENT` (the file is already absent — the delete's goal) and rethrows anything else.

## Validation
- `eslint --max-warnings 0`, `tsc -b`, and the full test suite (`xxscreeps test`, node 24) all pass.
- No automated regression test: both paths only manifest on the `file:` disk save→reload cycle, which can't be exercised without temporary files (the harness keeps fixtures in-memory). Reproduced and fixed manually.

### DIFF
--- a/packages/xxscreeps/engine/db/storage/local/blob.ts
+++ b/packages/xxscreeps/engine/db/storage/local/blob.ts
@@ -91,11 +91,11 @@ export class BlobStorage extends AsyncDisposableResource {
 			// Ensure it actually exists on disk
 			const path = Path.join(this.path, key);
 			try {
+				await fs.stat(path);
 				this.cache.set(key, {
 					saveId: this.saveId,
 					value: null,
 				});
-				await fs.stat(path);
 				return true;
 			} catch {
 				this.cache.set(key, {
@@ -266,6 +266,7 @@ export class BlobStorage extends AsyncDisposableResource {
 			return;
 		}
 		const entries = [ ...Fn.filter(this.cache, entry => entry[1].saveId === this.saveId) ];
+		console.log(entries);
 		++this.saveId;
 
 		// Save to disk
@@ -288,14 +289,7 @@ export class BlobStorage extends AsyncDisposableResource {
 				await fs.writeFile(tmp, value);
 				await fs.rename(tmp, path);
 			} else {
-				// The blob may have been created and deleted between two saves, in
-				// which case it never made it to disk
-				await fs.unlink(path).catch((err: unknown) => {
-					// @ts-expect-error
-					if (err.code !== 'ENOENT') {
-						throw err;
-					}
-				});
+				await fs.unlink(path);
 			}
 		});
 
--- a/packages/xxscreeps/engine/db/storage/local/blob.ts
+++ b/packages/xxscreeps/engine/db/storage/local/blob.ts
@@ -288,7 +288,14 @@ export class BlobStorage extends AsyncDisposableResource {
 				await fs.writeFile(tmp, value);
 				await fs.rename(tmp, path);
 			} else {
-				await fs.unlink(path);
+				// The blob may have been created and deleted between two saves, in
+				// which case it never made it to disk
+				await fs.unlink(path).catch((err: unknown) => {
+					// @ts-expect-error
+					if (err.code !== 'ENOENT') {
+						throw err;
+					}
+				});
 			}
 		});
 
--- a/packages/xxscreeps/engine/db/storage/local/keyval.ts
+++ b/packages/xxscreeps/engine/db/storage/local/keyval.ts
@@ -62,9 +62,10 @@ export class LocalKeyValResponder extends AsyncDisposableResource implements May
 		this.blob = blob;
 		if (payload !== undefined) {
 			const map = JSON.parse(payload, (key, value: SerializedValue) => {
-				if (typeof value !== 'object' || value instanceof Array) {
-					return value;
-				} else {
+				// `value` is typed as the serialized shape, but a JSON reviver also visits raw
+				// `null` and untagged plain objects that the type doesn't model
+				// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+				if (typeof value === 'object' && value !== null && !(value instanceof Array)) {
 					switch (value['#']) {
 						case 'map': return new Map(Object.entries(value.$));
 						case 'set': return new Set(value.$);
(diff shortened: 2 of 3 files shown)


packages/xxscreeps/engine/db/storage/local/blob.ts
```diff
@@ -288,7 +288,14 @@ export class BlobStorage extends AsyncDisposableResource {
 				await fs.writeFile(tmp, value);
 				await fs.rename(tmp, path);
 			} else {
-				await fs.unlink(path);
+				// The blob may have been created and deleted between two saves, in
```
laverdet: Did you observe this condition? The `saveId` stuff above is supposed to dispatch only files that should be updated or removed. I confirmed a naive `set`, `del`, `save` case doesn't attempt to write anything.

packages/xxscreeps/engine/db/storage/local/keyval.ts
```diff
@@ -62,16 +62,21 @@ export class LocalKeyValResponder extends AsyncDisposableResource implements May
 		this.blob = blob;
 		if (payload !== undefined) {
 			const map = JSON.parse(payload, (key, value: SerializedValue) => {
-				if (typeof value !== 'object' || value instanceof Array) {
-					return value;
-				} else {
+				// `value` is typed as the serialized shape, but a JSON reviver also visits raw
+				// `null` and untagged plain objects that the type doesn't model
```
laverdet: I don't think `null` can appear this file?

packages/xxscreeps/engine/db/storage/local/blob.ts
bastianh: I'm currently testing a mod that allow connecting chrome dev tools to the server and those two issues where observed during the end2end tests with attached debugger.. so maybe they are not generally happening without the mod.

here is the reasoning: 

You're right — the sequential `set → del → save` case is covered, by two mechanisms in `del()`:

- **Pending branch:** `set → del` before a save takes the `cached?.saveId === this.saveId` branch and writes the tombstone with `saveId: -1` (clean), so `save()` dispatches nothing.
- **Catch downgrade:** `del` of a nonexistent key stats the file and downgrades the tombstone to `{ saveId: -1 }` on ENOENT. Sequentially this self-heals.

The case that slips through is asynchronous, and it's a window opened by the reorder in 8237674d:

```ts
// before:                       // after (8237674d):
try {                            try {
  await fs.stat(path);             this.cache.set(key, { saveId: this.saveId, value: null });
} catch { /* clean */ }            await fs.stat(path);
this.cache.set(key, dirty);        return true;
                                 } catch { /* clean */ }
```

Before, a dirty tombstone (`saveId === current, value: null`) could only exist for a file that provably existed on disk. After, one exists for a never-written path for the entire duration of the `await fs.stat`. If `save()` snapshots the cache inside that window, it collects the tombstone, `++saveId` runs, and `fs.unlink` on the never-existing path throws ENOENT — rejecting the whole `spread(500, entries, …)` flush.

Two things make that window realistic to hit:

1. **`vDel` is fire-and-forget** (per the comment in `provider.ts`: "returns no response and may be reordered"; the local implementation is a floating `this.del(key)`), so nothing orders the stat window against `save()`.
2. **Threadpool saturation stretches the window.** The local keyval `del` falls through to `this.blob.del(key)` for any key not in the in-memory map — including blob keys that never existed. The code-push flow produces exactly that: `user/code.ts` calls `vDel(Code.buffersKey(…))` even for branches whose buffers blob was never created. With a save flushing hundreds of `writeFile` jobs (concurrency 500 on 4 libuv threads), the del's `fs.stat` can sit in the queue for seconds, so the next periodic save lands inside the window.

So the commit message's "created and deleted between two saves" was too loose — read sequentially that case is indeed covered. Tolerating ENOENT on unlink still seems like the right fix to me: it makes the tombstone idempotent (same semantics as a redis `DEL` of a missing key) and closes this whole class of interleavings without reworking the del/save ordering.

packages/xxscreeps/engine/db/storage/local/keyval.ts
bastianh: For this one the reasoning was: 

Serialization format (keyval.ts:768–781): maps are persisted as {'#':'map', $: Fn.fromEntries(...)} — a map's $ payload is a plain object, not an array. The top level of the file is itself such a map.

Before 8237674d the reviver was total:

```
switch (value?.['#']) {
    case 'map': ...
    default: return value;   // plain objects & null: pass through untouched
}
```
After 8237674d it became:
```
if (typeof value !== 'object' || value instanceof Array) return value;
else switch (value['#']) { case 'map': ... /* no default! */ }
```
A plain object (no # tag) falls through the switch → the reviver function returns undefined → per JSON.parse semantics, returning undefined deletes the property from its parent. Revivers run bottom-up, so the $ object is stripped out of `{'#':'map', $:…}` before the parent is revived → `new Map(Object.entries(value.$))` runs with `value.$ === undefined` → exactly the observed Cannot convert undefined or null to object. Since the top-level structure is itself a map, every persisted file is affected.

On the maintainer's question "I don't think null can appear in this file?" — He's right. The serializer only writes strings, numbers, tagged objects, and arrays of those; a well-formed data.json contains no null. The value === null guard in my fix doesn't fix the bug — only the default: return value branch does. The guard merely restores the tolerance the old code had: before the refactor the switch head was value?.['#'] (optional chaining), which implicitly absorbed null; after the refactor, null falls into the switch because typeof null === 'object' and throws when reading value['#'].

laverdet:
On the topic of devtools in xxscreeps, unless I'm misunderstanding what you were trying to do, you do not need a mod for it. I connect all the time for troubleshooting: `node --inspect --enable-source-maps --experimental-vm-modules --import xxscreeps/loader packages/xxscreeps/dist/engine/service/launcher.js`

PR #319, opened 2026-07-15
scrape: remove dependency on @screeps/launcher

ShuP1:
Add a `@xxscreeps/loki-init-db` package which extract `init_data/db.json` from `@screeps/launcher` without requiring it as dependency

Skipping 4 deprecated and 3 vulnerable sub-dependencies


laverdet:
Without the dependency `xxscreeps import` will fail. Like we talked about on Discord I'd rather dust off the import/export pipeline. Mod hooks for #309 & #321 will need to be added, but otherwise the work is pretty much already done.

PR #321, opened 2026-07-16
room-gen: place sources, controller, mineral, and keeper lairs

misterwise:
Follows #308 with the placement half of `@screeps/backend`'s `cli/map.js`: sources (random 1–2 by default), controller, mineral with density roll, and keeper lairs 3–5 tiles from each source and the mineral (vanilla's `map.js` guards only sources; live SK rooms guard the mineral too). New `GenerateRoomOptions` fields — `sources`, `mineral`, `controller`, `keeperLairs` — plus `--sources`/`--mineral` flags on `generate-room`.

Controller-less rooms (keeper/center) bake `SOURCE_ENERGY_KEEPER_CAPACITY` at generation — `Source` computes the same value in `'#roomStatusDidChange'`, but the controller processor that drives that hook never runs without a controller — and ship an unowned extractor so the mineral is harvestable. That surfaced a latent gap: `checkActiveStructures` never ported vanilla's owner-less-structures-are-always-active rule, so any unowned quota-typed structure was permanently inactive. The first commit fixes that with a regression test; `extractor.create` widens to `owner: string | null`.

The third commit integrates with mods: a new `terrain` provide slot, with `source`, `controller`, and `mineral` registering `roomGenerator` hooks on `scripts/symbols.ts` (the future home of import/export hooks). Each attempt builds CA terrain, then runs the hooks against a placement context (`place`/`isPlaceable`/`tagsAt`/`findRandomTile`); any hook returning false rerolls the terrain. Hooks carry an explicit `order` because mod resolution order (mineral → source → controller) can't express the mineral's stay-clear-of-sources-and-controller constraint. `room['#level']`/`'#user'` moved into the controller hook — controller-less rooms now leave `'#user'` `undefined` per the schema's tri-state — and mod options (`sources`, `mineral`, …) ride `GenerateRoomOptions` interface augmentation. The source and mineral hooks read `options.controller` for vanilla's keeper-room semantics (4000-capacity sources, prebuilt extractor) — a deliberate option-level leak rather than a cross-mod import. Two behavior notes: placeability is now "untagged", closing a latent edge where a keeper lair could land on the mineral tile; and source count / mineral type re-roll on each placement attempt (swamp layout still rolls once per room). A config that omits one of these mods generates rooms without those objects.

## Validation

- New test: an owner-less extractor below required RCL harvests `OK` — fails (`-14`) with the structure fix reverted. Full suite 485/485 after the mods rework.
- Offline smoke on fresh local shards (16 runs) re-run through the hook path: normal / keeper / center / custom-option / no-mineral rooms, asserting object counts, 4000 vs 1500 source capacity, mineral ≥ 5 from sources and controller, a lair near each source and the mineral on distinct tiles, an unowned extractor on the mineral tile, `#user` tri-state, all placements on wall tiles, and a `loadRoom` round-trip. `generate-room` CLI runs with valid and invalid `--mineral` (validation now reads `C.MINERAL_MIN_AMOUNT` keys, so the CLI no longer loads the mineral mod's registration module).
- `xxscreeps types` emits cleanly; the `terrain` provide never enters the player types program.
- Build and eslint clean. No tests for the generation script itself, matching #308.

### DIFF
--- a/.changeset/room-object-generation.md
+++ b/.changeset/room-object-generation.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Place sources, controller, mineral, and keeper lairs in generated rooms; owner-less structures are always active.
--- a/.changeset/room-object-generation.md
+++ b/.changeset/room-object-generation.md
@@ -2,4 +2,4 @@
 "xxscreeps": patch
 ---
 
-Place sources, controller, mineral, and keeper lairs in generated rooms; owner-less structures are always active.
+Place sources, controller, mineral, keeper lairs via mod terrain hooks; owner-less structures always active.
--- a/README.md
+++ b/README.md
@@ -167,14 +167,21 @@ await using shard = await Shard.connect(db, 'shard0');
 
 await generateRoom(shard, 'W12N5');
 await generateRoom(shard, 'W11N5', { terrainType: 3, swampType: 2 });
+await generateRoom(shard, 'W10N5', { sources: 1, mineral: 'H' });
 await Promise.all([ db.save(), shard.save() ]);
 ```
 
-`generateRoom` adds a room with procedurally generated terrain to the shard's
-terrain blob and `rooms` set. Like `npx xxscreeps import`, it is an offline
-operation — stop the server before running it so cached world state in the
-backend, processor, and runner workers doesn't go stale. Exits are read from
-any already-generated neighbor, so adjacent generated rooms connect.
+`generateRoom` adds a room with procedurally generated terrain and objects to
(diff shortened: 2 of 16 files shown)


laverdet:
This is a good first pass. I think the things that could change later:

- `grid` is an artifact of the terrain generation procedure. By the time these hooks come into play I think it's fine to flatten that into a `Terrain` object.
- `RoomPosition` should be used instead of an abstract grid. Then you can reuse those helpers like `iterateNeighbors`, `iterateWithRangeTo` and so on.
- "Random placement" logic is now repeated for deposits, powerbanks, and terrain generation with slightly different characteristics each time. You could do a feistel network random iteration if you want to have some fun.

PR #322, opened 2026-07-17
pf: cmake 4.4 'import std' UUID

misterwise:
CMake 4.4 rotates the experimental `import std` opt-in UUID, and the macos-15 arm64 runner image picked up cmake 4.4.0 on 2026-07-15, so the generate step now hard-fails on `CXX_MODULE_STD`. This adds the 4.4 UUID to the version ladder.

## Validation

- Reproduced the CI generate failure on darwin-arm64 with cmake 4.4.0 and the workflow's toolchain env; with the fix, configure/generate passed and a full `ninja -C build` linked both modules.
- The Windows job failure in the same runs is a separate pnpm 11 module-resolution issue, not addressed here.


laverdet:
My [PR upstream](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/12053) landed in this release so we'll soon run into issues with the patch logic 
https://github.com/laverdet/xxscreeps/blob/b88d6af3cfff04849d23aee6aaf1bb62c6cd0ddd/.github/workflows/pathfinder.yml#L155-L168

misterwise:
Nice, so once msys2 ships 4.4 the hunk stops applying. Want me to gate the patch step on `cmake --version` < 4.4 in this PR, so it skips once your upstream fix arrives? (Current Windows red looks like the separate loaderkit resolution failure, not the patch.)

PR #327, opened 2026-07-19
(Stats diff #3) backend: aggregate room stat buckets into redis and serve them

bastianh:
Rooms flush their accumulated stat bucket once wall-clock time crosses the next boundary of the finest interval (8 minutes), jittered per room to spread the writes; the batch is credited to the bucket the room's timestamp falls in. The dump feeds one hash per (room, interval) in shard storage - a single hGetAll enumerates a room's contributors for the world-map layer - and per-user account hashes in db.data so totals aggregate across shards like GCL. Expired fields are reclaimed by the dump, and the not-yet-flushed blob bucket is merged into reads wherever the room is already loaded.

This serves the classic client's stat surfaces: /api/user/stats totals, /api/game/room-overview owner punchcards, per-room punchcards on /api/user/overview, and per-user stat layers with statsMax on /api/game/map-stats via the mapStats hook, which grows a `response` object for top-level fields.


laverdet:
What do you think about dropping the per-user per-room stats? If I'm understanding correctly, and this explains what I'm seeing on World, you can only see your own stats in a given room? It seems kind of boring, like I'm browsing around World and all the stats pages are empty because I don't have any activity. So we'd store room stats (for all users in the room) and user stats (aggregated per shard). It would simplify the implementation and generally be more exciting. Win/win?

Actually though do you think it's a good idea or not

bastianh:
I think the only location where it is used is the energy harvest map view. In rooms where multiple users harvested energy you can see all user badges sized by how much energy they harvested in that room. 

But I'll check with the client again.

laverdet:
Are you seeing something different on World? I know this guy has been harvesting energy for a whole decade but I don't see any of it on the stats page.

image

bastianh:
you can see it in the map overlay where you can select the stats .. and I think it can also show multiple badges on a room or entries in the tooltip by user

the map view can also fetch a lot of room stats at the same time depending on the zoom level. I think the client only fetches the mineral/owner for zoomed out view .. but the api has no such restriction and can get thousands rooms at once

Screenshot 2026-07-21 at 20 46 23

laverdet:
- Room overview aggregates all users, instead of just the logged in user. I think only showing your own stats here is wrong and boring, even though I guess it's what World does.
- loadBackendUserInfo utility function for loading username + badge
- More aggressive 404's on missing parameters
- Functional convention fixes
- Claude continues to try and reinvent the wheel when it comes to seeded randoms. I used the existing random utilities here for the jitter function
- `peekRoom` preferred over `player` for observing game state

image

PR #329, opened 2026-07-20
memory: skip the save when Memory fails to serialize

misterwise:
Storing an unserializable value in `Memory` — a circular reference being the easy accident — errored every tick: `crunch` walked the structure outside the `try`, so the failure escaped `flush` instead of being contained to the save. `crunch` now runs inside the `try`, and the catch also drops `previousJson`, so the next read re-parses the last saved string rather than serving cached mutations the save never captured — without that, the save fails silently-in-effect while cached `Memory` keeps working, and the failure only surfaces as a rollback at the next isolate reset.

This degrades more gently than vanilla, where the post-tick stringify is unwrapped ([runtime.js L246-248](https://github.com/screeps/driver/blob/cf63d8adf902663e2ebddd7f8c5b7baa425dc928/lib/runtime/runtime.js#L246-L248)) and a circular `Memory` costs the whole out-message, intents included. Here it's a console error and a skipped save; the rest of the tick's payload still flushes.

Previous revision of this PR also extended `crunch`'s coercions (`NaN`/`Infinity`/functions) and invalidated the cache on a deliberate `_parsed` discard — both withdrawn per review: no observed script breakage, and the bots shipping `delete RawMemory._parsed` (ZeSwarm, wiki MemHack) pair it with a heap-cached `Memory` or `RawMemory.set`, so they never touch the cache path.

## Validation

- screeps-ok `UNDOC-MEMJSON-005` (circular `Memory` per-tick error, the `memory-circular-ref-crash` gap) — passed; the `-001`/`-003`/`-004`/`MEMHACK-011` IDs from the previous revision return to expected-fail
- No local sandbox test: the expected `console.error` trips the harness's player-error assert; coverage lives in screeps-ok 005
- `npx xxscreeps test` — 489 passed
- `pnpm run build`

### DIFF
--- a/.changeset/spicy-donuts-refuse.md
+++ b/.changeset/spicy-donuts-refuse.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Unserializable `Memory` now skips the save instead of erroring every tick
--- a/packages/xxscreeps/mods/meta/memory/memory.ts
+++ b/packages/xxscreeps/mods/meta/memory/memory.ts
@@ -267,12 +267,14 @@ export function flush() {
 	if (_parsed) {
 		// Typical case: user accessed `Memory`, so we simulate vanilla reconstruction and save the
 		// string.
-		crunch(previousJson = _parsed as MemoryRecord);
 		try {
+			crunch(previousJson = _parsed as MemoryRecord);
 			string = JSON.stringify(previousJson);
 			isBufferOutOfDate = true;
 		} catch (err) {
 			console.error(err);
+			// The save was skipped, so the object may hold mutations the saved string never received
+			previousJson = undefined;
 			return { size: memoryLength };
 		}
 	} else if (!isBufferOutOfDate || string === undefined) {
--- a/packages/xxscreeps/mods/meta/memory/memory.ts
+++ b/packages/xxscreeps/mods/meta/memory/memory.ts
@@ -273,7 +273,10 @@ export function flush() {
 			isBufferOutOfDate = true;
 		} catch (err) {
 			console.error(err);
-			// The save was skipped, so the object may hold mutations the saved string never received
+			// The user's memory object cannot be serialized. Probably, it is circular or contains some
+			// other `toJSON` error. It will be reparsed from `string` or `memory` next tick.
+			// TODO: Does it make sense to do a sandbox reset here instead of attempt to continue
+			// gracefully?
 			previousJson = undefined;
 			return { size: memoryLength };
 		}


laverdet:
Have you observed these values (NaN, Infinity) causing problems with user scripts? I'd like to document which scripts and how.

There's a thousand other ways in which `crunch` differs from JSON.stringify / parse. Via JSON: Dates will become ISO strings. Circular objects are flattened. Sparse arrays are materialized. That's just off the top of my head.

So if this is just a matter of chasing a spec then I don't want to do it. If it's a matter of fixing something that actually broke then we can figure it out.

misterwise:
Ok, I wasn't able to find a case in public bots of the value issue, so will drop it. Generally I use that as the filter for any undoc issues now and will flag the screeps-ok parity cases as won't fix (unless we find an example).

The stale cache issue would also require someone to delete parsed without fully doing memhack, so I think it goes as well, it's chasing parity too much.

There is a crash bug that this resolves, so I'll trim down to just that.

laverdet:
What error does `crunch` throw that this diff now catches? Is it a RangeError?

misterwise:
Yes, it would be RangeError for circular reference (that's the most common one in my experience) but anything that throws really. Here's Claude's summary 

> For a circular (or absurdly deep) Memory it's RangeError: Maximum call stack size exceeded — crunch recurses with no cycle guard and overflows before stringify can throw its circular-structure TypeError. crunch can also propagate a user getter's error (Object.entries invokes getters, same as stringify would), and — strict-mode wrinkle — a TypeError when it tries to delete/assign an undefined member of a frozen object, a case vanilla's stringify handles by just omitting the field. All of those escaped flush before this diff; the circular RangeError is the one that's an easy accident.

bastianh:
since you are currently working on the memory maybe it's a good time to also think about implementing the api endpoints that allows changing memory values via the api?

laverdet:
The way the code is right now though the RangeError would bubble up to the sandbox and the sandbox marks itself as stale. Then we instantiate a new sandbox. Screeps is able to recover from this condition because it sends the active memory payload blob every tick, which would get reparsed. In our case the player sandbox manages its own memory payload and never receives updates after initialization. So it seems like letting the error hard reset is the only solution. If anything the try/catch should be removed entirely because we can't recover the user's memory once they've corrupted their memory state.

misterwise:
I traced the mechanics and you're right: the RangeError escapes `flush()`, `tick()` returns `error`, the runner marks the instance stale, and the rebuild reloads memory from the db (plus a skipped tick from the "player's fault" path).

But I don't think the sandbox needs a driver update to recover. `string` in memory.ts is only overwritten on a successful stringify (or a validated `RawMemory.set`), so after a failed flush it still holds the last-good payload — the same bytes the db has. That's what the `previousJson = undefined` line in the catch is for: next tick the `Memory` getter falls through to reparsing `string`, which is the same reparse-last-good recovery vanilla gets from re-sending the blob, just sourced locally.

So both paths converge on the same rollback; the difference is cost. The hard reset discards the erroring tick's intents (the throw happens in the send phase, after they're collected), skips a second tick, and takes a global reset — and if the user's code recreates the cycle every tick, that's a teardown/rebuild loop every two ticks. With the catch, the tick's intents survive, saves skip with a console error, and memory rolls back one tick later.

Removing the try/catch entirely would also drop the existing stringify catch, which currently keeps `previousJson` on failure — that lets the user keep running on an object that never saves, so sandbox and db diverge silently until some unrelated reset rolls back everything. Clearing `previousJson` bounds that divergence to one tick.

laverdet:
> Removing the try/catch entirely would also drop the existing stringify catch, which currently keeps previousJson on failure

Removing the try/catch would do a hard reset, as discussed above.

Anyway this is enough of an edge case where maybe it's not super important. I like the hard reset because if you've created a circular reference this tick then you'll probably make one next tick too. The `RangeError` from crunch especially is an expensive one, and requires burning a lot of CPU and stack space before we even get there. But we can catch the error for now and come back to the note later.

misterwise:
Post-merge observation for the validation record: consuming d2268ac5 in screeps-ok showed that UNDOC-MEMJSON-005 is still environment-sensitive. A fresh macOS arm64 build passes, but the clean Ubuntu 24.04 x64 CI runner on Node 24.14.1 times out during the circular Memory flush: https://github.com/misterwise/screeps-ok/actions/runs/29882027764/job/88804745711

The sandbox tick limit there is 500 ms. On that runner, crunch remains in recursive traversal until the isolate timeout fires, so it does not reach the RangeError catch and recovery path. The same test passes locally because the recursion reaches a catchable RangeError before the timeout.

No request to revisit this edge case; I am only recording the remaining platform-dependent limitation so the merged PR validation note is not read as universal.

PR #330, opened 2026-07-22
invader: drive stronghold defense with per-bunker behaviors

misterwise:
Deployed strongholds now defend themselves. Behaviors dispatch off the core's template name (vanilla keys them off `strongholdBehavior`, which always equals the template name): the controller driver moved from the NPC loop into `loop/stronghold.ts`, joined by `refillTowers` (chance-gated by core level), `focusClosest` (every tower and defender in reach attacks the hostile nearest the core), and a per-bunker defender population whose simple-melee creeps path only over the stronghold's ramparts. bunker4/5's specialized behaviors (focusMax, assignDefenders, antinuke, towersMaintenance, refillCreeps) fall back to the basic set for now.

System-user structures are exempted from controller activity limits so NPC towers can fire; they no longer count against a player's allowance either. Divergence from Screeps, whose equivalent runtime check reports them inactive while its processor, which never checks, fires them anyway.

The second commit picks up the #311 merge note: the per-structure `#strongholdId` is removed, and the core instead records its unowned peers (roads, reward containers) in an `ownedNeutralStructureIds` vector. Still write-only — reads arrive with stronghold maintenance/loot.

Known limitations, resolved by the expansion/placement slice:

- The loop gates raid-vs-defense at room level (any core present → defense only), where vanilla gates per creep (`!strongholdId` → raid AI). The two only coexist once expansion cores sit in raid-eligible rooms; the per-creep split needs a defender marker and lands with that slice.
- `FIND_HOSTILE_CREEPS` misses hostile power creeps, which vanilla's stronghold targeting includes; moot until power creeps can act, and targeting them raises an invader→powercreep mod-boundary question worth settling then.

## Validation

- Full suite 515/515. New coverage: refill target selection (emptiest rampart-protected tower wins, unprotected never refilled), focus fire from towers + melee + ranged + point-blank mass attack, population spawn order and slot-skip, rampart-confined pursuit and hold, a deploying core running no defense behaviors, and the core recording exactly its neutral peers.
- Live smoke on a fresh `xxscreeps import` world: injected a bunker2 core and watched deploy populate `ownedNeutralStructureIds` (6 roads + 2 containers), the controller reservation climb, `defender0` spawn and hatch, and both towers one-shot an injected hostile (energy 1000→990 each). Ticks stayed clean across graceful restarts.


laverdet:
Can you split out raiders into mods/classic/invader and strongholds into mods/modern/stronghold please.

For the NPC loop you can either do a monkey patch from mods/modern, or just use a different userId for invaderCore invaders.

misterwise:
> Can you split out raiders into mods/classic/invader and strongholds into mods/modern/stronghold please.
> 
> 
> 
> For the NPC loop you can either do a monkey patch from mods/modern, or just use a different userId for invaderCore invaders.

Can do, I'm also thinking maybe a config option to disable invaders entirely as the speedrunners seem to want a simple way to achieve this.

laverdet:
Can't they just remove the mod?

misterwise:
Yep, I had gotten so used to the mods not being treated like mods I hadn't even considered that..

PR #331, opened 2026-07-22
utility: iterate random placement through a feistel permutation

misterwise:
Adds `utility/random.ts` — a feistel permutation over `[0, count)` (cycle-walking non-power-of-4 domains) with lazy `shuffle` and `shuffledSquare` views — and converts the hand-rolled randomness sites onto it: room-gen's `findRandomTile`, deposit and powerbank `findPlacement`, and the stronghold loot shuffle (retiring its fisher-yates TODO). Follows the #321 merge comment's suggestion.

Semantics change worth noting: exhaustive random-order iteration replaces capped rejection sampling, so placement fails only when no position qualifies — never spuriously after 1000 misses. Distribution is unchanged: the first accepted element of a uniform permutation is uniform over the accepted set.

## Validation
- Added a pure-math spec test: the permutation yields each index exactly once (including 0, 1, and power-of-4 domains) plus a coarse first-index uniformity smoke, under the deterministic RNG fixture.
- Full suite passed (511 tests); eslint clean on changed files.

### DIFF
--- a/.changeset/random-placement-primitive.md
+++ b/.changeset/random-placement-primitive.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Unified random placement on a feistel permutation; placement fails only when no position qualifies.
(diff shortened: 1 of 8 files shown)


laverdet:
Very cool

PR #335, opened 2026-07-23
powercreep: enable rooms and use powers, starting with PWR_GENERATE_OPS

misterwise:
Adds `PowerCreep.enableRoom` and `usePower`, with per-power cooldowns stored as `cooldownTime: 'int32'` on the `#powers` vector — the public `powers` getter returns `{ level, cooldown }` and the room renderer sends the absolute tick. `PWR_GENERATE_OPS` is the first working power: it credits `effect[level - 1]` ops, spills overflow to the ground, stamps the cooldown, and appends an `EVENT_POWER` event-log entry plus a `power` action-log entry. Per-power dispatch is the processor switch, per the #295 merge comment — no hook substrate.

Deferred, flagged rather than silent:

- Vanilla's stronger-active-effect `ERR_FULL` re-application check is deferred until the first targeted power lands — today no object ever carries an effect to re-apply against.
- The `power` action-log entry can't carry vanilla's power id (fixed `{type, x, y}` entry shape) — client animation nicety.
- A learnable-but-unimplemented power (e.g. `PWR_OPERATE_SPAWN` today) no-ops without burning cooldown or ops.
- Upgrading a spawned creep takes effect at next spawn; vanilla applies it to the live creep immediately. Syncing account upgrades onto the spawned room object is left for a follow-up PR.

## Validation

- Eight new harness tests over the distinguishing branches (unlearned power, disabled room, enable-then-use flow, hostile safe mode on both verbs, cooldown/`ERR_TIRED` with the event-log assert via `peekRoom`, overflow spill, ops-before-target check ordering); 526/526 total, zero new lint warnings.
- Live-smoked on a worktree server: spawn → pre-enable `ERR_INVALID_ARGS` → `enableRoom` → repeated `GENERATE_OPS` at the exact 50-tick cooldown cadence with `ERR_TIRED` between uses and ops accruing; the roster endpoint returns `{cooldown, level}` and the room socket renders the absolute `cooldownTime`.


misterwise:
Rebased onto the `xx.d.ts` restructuring — the runtime side ports cleanly (registrations in `schema.ts`, enum/`RoomSchema` merges in `xx.d.ts`). Four type errors remain that look like the mmo project isolation closing these channels to mmo mods — this seems to be the first mmo mod to register an action-log action or event-log variant:

- `tsconfig.mmo.json` consumes the main project's emitted declarations, where the action-log member type and the `Room.eventLog` union are already baked as closed literal unions (`dist/game/schema.d.ts` has `"attack" | … | "produce"`), so an mmo-side `enum ActionLogSchema` / `RoomSchema` merge can't widen them — `saveAction(creep, 'power')` and `appendEventLog(room, { event: EVENT_POWER, … })` don't typecheck.
- `Room.getEventLog`'s augmentation (`game/room/event-log.ts`) is also invisible to mmo sources, even with the module explicitly imported.

Happy to implement whichever shape you prefer for keeping these open across the project boundary, or to adapt if you already have a plan here.

laverdet:
- `usePower` intent processor was unintelligible. I had to ask the AI what it meant, I had no idea. I converted it to the same pattern we use in harvest.
- The sneaky ternary in `.powers` trying to avoid invoking `cooldownTime`.
- `@public` comments should retain their docs from the Screeps API. Unless it's "with the additional properties" where they're defined on the returned interface.

PR #342, opened 2026-07-25
leaderboard: rank players on monthly boards

bastianh:
Based on #341 — it needs the `zRank` added there, so this branch contains that commit too and its diff will shrink to the leaderboard commit alone once #341 merges.

Replaces the `/api/leaderboard/*` stubs with real monthly boards for the two modes the client ranks: `world` (control points) and `power` (power processed). A season is a UTC calendar month; each `(mode, season)` is one sorted set in `db.data`, so standings aggregate across shards like GCL and the account stat totals.

Both scores are already recorded as stat contributions, so nothing new is measured — `mods/meta/stats` grows a `flush` hook and the leaderboard accumulates the same batch the per-interval series are credited with. That keeps the tick free of extra writes and keeps the two mods from knowing about each other beyond the declared dependency.

`makeValidatedQueryRoute` gains the `coerceTypes` option `makeValidatedPayloadRoute` already had; the paging query has numeric fields and query strings arrive as strings.

## Validation

`tsc -b` and `eslint --max-warnings 0` clean; 538/538 tests pass. `mods/meta/leaderboard/test.ts` covers season boundaries, per-mode accumulation, ranking and tie order, paging windows, the season list, per-player history, teardown through `User.remove`, and a bucket flushed through `writeRoomBucket` landing on the right month.


packages/xxscreeps/mods/meta/leaderboard/model.ts
```diff
@@ -0,0 +1,157 @@
...
+		return entry && { ...entry, season };
+	});
+	return [ ...Fn.filter(entries) ];
+}
+
+/**
+ * Drop a removed user from every leaderboard. Their absence renumbers the players below them, which
+ * is what the sorted set does for free.
+ */
+async function removeAllForUser(db: Database, userId: string) {
```
laverdet: I'm not sure we should reorder archived leaderboards. It would be better to show a placeholder tombstone entry for missing entries.

packages/xxscreeps/mods/meta/leaderboard/backend.ts
```diff
@@ -0,0 +1,95 @@
...
+};
+
+// `GET /api/leaderboard/find?mode=world&username=…[&season=2026-07]` — one player's standing. With
+// a season this answers "my rank" and the search box; without one the profile page reads every
+// season at once. Being unranked is an error response, which the client catches everywhere.
+hooks.register('route', {
+	path: '/api/leaderboard/find',
+
+	execute: makeValidatedQueryRoute(leaderboardFindSchema, async context => {
+		const { mode, season, username } = context.request.query;
```
laverdet, line 88: Does the client actually send null'd season entries? If the client doesn't send it I don't think we should expect it. Also for queries the client never send I think it's ok to respond with something different than the Screeps MMO host. I forged a request with null season and got back an empty list. I think it's ok for us to just 404.

packages/xxscreeps/backend/index.ts
```diff
@@ -110,13 +110,19 @@ export function bindTerrainRenderer<Type extends RoomObject>(object: Implementat
 	object.prototype[TerrainRender] = render;
 }
 
+interface ValidatedRouteOptions {
+	/**
+	 * Compile against the coercing validator, which rewrites values in place to the type the schema
+	 * asks for (e.g. "180" → 180) before `execute` sees them.
+	 */
+	coerceTypes?: boolean;
```
laverdet: We should consider just making this the default and removing the option.

packages/xxscreeps/mods/meta/leaderboard/model.ts
bastianh: I dropped the remove registration entirely rather than special-casing the current season, so a score is never rewritten after it lands.

The client needs an actual object to render against: lobby-world.html does LeaderboardList.users[item.user].username with no guard, and Angular's parser swallows the undefined lookup, so a missing entry renders as a nameless row with a dead profile link. list now fills in { badge: null, username: '[deleted]' } — checkUsername rejects brackets, so it can't collide with a real account.

packages/xxscreeps/mods/meta/leaderboard/backend.ts
bastianh: `Api.get` is `$http({method: 'GET', params})` on the default `$httpParamSerializer`, whose body starts `null === e || isUndefined(e) || (…)`, so both null and undefined params are dropped from the query string. The type is `season?: string` now.

The omitted-season form does need to keep answering with a list, though. `ProfileLoader` calls `leaderboard/find` with only `{username, mode}` for world and power, and unlike every lobby call site it has no `.catch` — an `{error}` body rejects in the Api service and fails the whole route resolve, so a 404 there would send every profile page to `profile-not-found.html`. I kept that branch and simplified the rest: unknown user or unknown mode is now just `Result not found`, which is the case the client genuinely never asks for.

packages/xxscreeps/backend/index.ts
bastianh: 👍🏼

PR #350, opened 2026-07-27
driver: decode the runtime source map with trace-mapping

misterwise:
`source-map-support` decodes the runtime bundle's map the first time it translates a frame from it, inside the isolate, charged to whichever player tick first reads `error.stack` — 38-41ms on current main against a bucket that refills at `kCPU: 100` per tick. And because `tickLimit` is `Math.min(config.runner.cpu.tickLimit, bucket)`, a player whose bucket has drained below the decode cost times out instead of paying it, is marked stale, and the recreated sandbox decodes again on its first stack.

The cost is the decoder, not the map. `source-map-support@0.5.21` resolves `source-map@0.6`, whose consumer materializes and sorts an object per mapping before it can answer the first query. This PR aliases `source-map` in the runtime bundle onto a small shim over `@jridgewell/trace-mapping` — the decoder webpack itself uses to build this bundle, deduped onto the copy already in the lockfile. `source-map-support` touches four members of the consumer — `new SourceMapConsumer`, `.sources`, `.sourcesContent`, `.originalPositionFor` — and duck-types the last, so the shim is exactly those four.

Measured in a real isolate via `isolate.cpuTime`, the clock `IsolatedSandbox.run` bills against `tickLimit`, on the bundle current main produces:

|  | initialize | first `error.stack` | heap at init | heap after decode |
|---|---|---|---|---|
| main | 13.8-14.3ms | 37.6-40.9ms | 5.3-5.4 MiB | 17.7-17.8 MiB |
| this PR | 13.2-14.7ms | 6.8-9.9ms | 5.2-5.3 MiB | 10.3-11.1 MiB |

Initialization and per-tick cost are unchanged; `source-map@0.6` also drops out of the bundle.

## Validation

- `npx xxscreeps test` — 645 passed; `eslint --max-warnings 0` clean on the changed files
- The two decoders agree on all 155,403 generated positions in the current bundle — every mapped segment, the columns either side of it, line starts, past end-of-line, and past end-of-file. Translated stacks are identical.
- Not covered by an in-repo test: `test/simulate.ts` pins `sandbox: 'unsafe'`, which never loads this module, so no in-repo test exercises the changed path.

### DIFF
--- a/.changeset/runtime-source-map-decoder.md
+++ b/.changeset/runtime-source-map-decoder.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Decode the runtime source map with `@jridgewell/trace-mapping` instead of `source-map@0.6`
--- a/packages/xxscreeps/driver/runtime/source-map-consumer.ts
+++ b/packages/xxscreeps/driver/runtime/source-map-consumer.ts
@@ -1,0 +1,26 @@
+import type { Needle, SourceMapInput } from '@jridgewell/trace-mapping';
+import { TraceMap, originalPositionFor } from '@jridgewell/trace-mapping';
+
+// `source-map-support` resolves `source-map@0.6`, whose consumer materializes and sorts an object
+// per mapping before it can answer the first query. For the runtime bundle that is most of a
+// player's CPU limit, charged to whichever tick first reads `error.stack`. Webpack aliases
+// `source-map` here instead, onto the decoder webpack already uses to build the bundle.
+export class SourceMapConsumer {
+	private readonly tracer;
+
+	constructor(map: SourceMapInput) {
+		this.tracer = new TraceMap(map);
+	}
+
+	get sources() {
+		return this.tracer.resolvedSources;
+	}
+
+	get sourcesContent() {
+		return this.tracer.sourcesContent;
+	}
+
+	originalPositionFor(needle: Needle) {
+		return originalPositionFor(this.tracer, needle);
+	}
+}
--- a/packages/xxscreeps/driver/sandbox/isolated/index.ts
+++ b/packages/xxscreeps/driver/sandbox/isolated/index.ts
@@ -27,6 +27,7 @@ const getRuntimeSource = runOnce(() => {
 		babel: [ Privates ],
 		alias: {
 			process: 'xxscreeps/driver/sandbox/isolated/process.js',
+			'source-map$': 'xxscreeps/driver/runtime/source-map-consumer.js',
 			'/xxscreeps:private-symbol': 'xxscreeps/driver/private/symbol/isolated-vm.js',
 			'xxscreeps/engine/schema/build/index.js': 'xxscreeps/engine/schema/build/runtime.js',
 			'xxscreeps/game/constants/index.js': import.meta.resolve('xxscreeps/game/constants/index.js'),
--- a/packages/xxscreeps/package.json
+++ b/packages/xxscreeps/package.json
@@ -23,6 +23,7 @@
     "@babel/core": "~7.29.7",
     "@babel/types": "^7.29.7",
     "@isolated-vm/experimental": "^0.0.10",
+    "@jridgewell/trace-mapping": "^0.3.31",
     "@loaderkit/resolve": "^1.0.6",
     "@screeps/launcher": "4.1.0",
     "@types/convert-source-map": "^2.0.3",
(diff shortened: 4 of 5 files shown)


laverdet:
1ccbea9ec7e9419b4bd746418d6fc102cb4bbf7a can help. That drops the source map to 500kb. It could probably get a lot smaller with other build optimizations too. It's just an area I haven't focused on.

misterwise:
You were right that there was more here — it just wasn't size. I chased the wrong variable, and I shouldn't have closed the door on yours.

`source-map-support@0.5.21` resolves `source-map@0.6`, whose consumer materializes and sorts an object per mapping before it can answer a single query. That, not the byte count, is the 33ms. Aliasing `source-map` onto a ~20-line shim over `@jridgewell/trace-mapping` — the decoder webpack already uses to build this bundle, and already in the lockfile via `@babel/core` and `terser` — measured on the bundle 1ccbea9e produces:

|  | initialize | first `error.stack` | isolate heap |
|---|---|---|---|
| today | 12.6-13.6ms | 34.8-37.7ms | 16.9-18.0 MiB |
| this PR's warmup | 47.1-49.1ms | 0.3-0.4ms | 17.4-18.1 MiB |
| trace-mapping alias | 12.6-16.2ms | 6.1-8.7ms | 9.9-10.3 MiB |

The alias is strictly better than today on every axis, including the ~7 MiB of isolate heap the warmup was going to cost you rather than save. `source-map@0.6` also stops being bundled, so the map drops 509,644 → 483,336 B. Adding the dependency is a three-line lockfile change; it dedupes onto the copy already resolved.

So I would rather withdraw the warmup than land both. It buys the last 6ms by paying 35ms of initialization in every isolate, unconditionally, and that is a bad trade once 6ms is what is left.

`source-map-support` touches four members of the consumer — `new SourceMapConsumer`, `.sources`, `.sourcesContent`, `.originalPositionFor` — and duck-types the last, so the shim is small and the alias sits beside the `path` and `buffer-from` ones already in that transform.

Worth flagging what I did not do. `source-map-support` is the only reason this bundle carries `path-browserify`, the `buffer-from: false` alias, and the `process` ProvidePlugin — nothing in the runtime imports `path` itself — on top of pinning `source-map@0.6`. Replacing it with a `prepareStackTrace` over `trace-mapping` would delete all four, and a CommonJS dependency with them. I did not propose that here because the frame-formatting edge cases — eval origins, constructor and method names — are the whole reason that library exists, and getting them subtly wrong degrades every player's stack trace. Say the word if you would rather have the bigger version.

## Validation

- `npx xxscreeps test` — 536 passed; `eslint --max-warnings 0` clean
- Measured by running the real runtime bundle in a real isolate and reading `isolate.cpuTime` deltas, the same clock `IsolatedSandbox.run` bills against `tickLimit`. Three isolates per configuration, twelve for the alias.
- The two decoders agree on all 161,601 generated positions in that bundle — every mapped segment, the columns either side, line starts, past end-of-line, and past end-of-file. Translated stacks match (`webpack:/module/packages/xxscreeps/driver/runtime/index.ts:86:25`), modulo one minified frame name that shifts because the bundle no longer contains `source-map`.
- Same caveat as before: `test/simulate.ts` pins `sandbox: 'unsafe'`, so no in-repo test loads this path.

Want me to repoint this PR at the alias, or close it and open a fresh one?

laverdet:
How is the "trace-mapping alias" heap measurement *lower* than today's baseline, which does no work?

Anyway I definitely don't want to burn 45ms on every initialization just in case the user examines a `.stack` property, that's obviously unreasonable.

misterwise:
The heap column was measured after the stack reads — post-decode in every row — and I should have said so in the table. At that point the baseline is not idle: it has just decoded the map with `source-map@0.6`, which retains a JS object per mapping, sorted into arrays. `trace-mapping` keeps the parsed segments in packed arrays instead. The difference is what each decoder retains, not the baseline doing work it doesn't do.

Re-measured on current main with heap sampled at both points:

|  | heap at init | heap after first `error.stack` |
|---|---|---|
| main | 5.3-5.4 MiB | 17.7-17.8 MiB |
| trace-mapping alias | 5.2-5.3 MiB | 10.3-11.1 MiB |

Identical before any stack read, as it should be — the alias does nothing at initialization.

Agreed on the warmup. I repointed this PR at the alias: rebased onto current main, dropped the warmup commit, rewrote the description. The numbers there are re-measured against the current bundle, since #357 changed its shape.

PR #355, opened 2026-08-01
controller: release the controlled room when downgrading to neutral

bastianh:
`release` tells a controlled room from a reserved one by reading `controller.level`, but the downgrade path decremented `room['#level']` before calling it. Going from level 1 to neutral therefore removed the room from the player's *reserved* rooms and left the controlled-rooms entry behind. The stale entry keeps the room in `/api/user/rooms` and `/api/user/overview` after it goes neutral, holds `world-status` at `normal`, and counts against the GCL cap in `claimController` — permanently costing the player a claim slot. The decrement was redundant anyway: both branches publish the new level through `updateRoomStatus`.

The regression test needs `simulate` to seed the per-user indices mods maintain. That is the processor worker's `initialize` step, so it moves into `initializeRoom` and both call it. The controller's `refreshRoom` registration moves from `driver.ts` to `processor.ts` along the way — it is a processor hook, and it was only reaching the harness because the test entry point imports the driver slot.

## Validation

`npm test` — 537 passed. Reverting the one-line processor change and re-running fails the new `downgrade to neutral` test on the post-tick room count (1 vs 0). `npm run lint` reports only three warnings, all in files this branch does not touch.


laverdet:
Mapping a hook has side effects, it closes the registration point permanently. We should only be doing that at "leaf" / entry modules. So I moved that to another file and made it a function maker instead. Some of the model details are getting a bit out of scope, these should generally be plain database functions.

PR #361, opened 2026-08-05
Decorations

bastianh:
Hey! 

First thanks for reviewing, tweaking and merging our merge requests  :) 

This branch is about decorations.. it includes basically the whole feature from inventory, catalog and rendering the decorations in the end.

This PR is not meant to be reviewed/merged at once .. I tried to shape the commits that they can basically merged one after another in a seperate PRs .. so this is meant as the big picture overview.

Maybe you want to have a short look over it or the commits and already add remarks about architectural decisions or something like that .. otherwise I could create seperate PR later.

Thanks :)


laverdet:
If you grep for `\b_id\b` you'll find they're (almost) all confined in backend implementations. This a client-only quirk which shouldn't be treated as any more canonical than a random GET query name. Parsing a request from the frontend client should only be done at the backend layer, and certainly never in the database model.

On room activations: what if you store these in a zset? Score = roomId (`parseRoomNameToId`), element = itemId. You can fetch all decorations in a room with zRange, or find which room a decoration is in with zScore. Then, all this back bending over where a decoration exists just goes away and is inherent to the data structure. After insertion you fetch the list of local room decorations again and remove conflicts in order of itemId.

If you move parsing of the client activation request out of `activate` and into the backend I think `activate`s spaghetti logic will more obvious. It seems like there are a lot of conditionals to shoehorn activation of two fundamentally different indices into the same function.

`...index !== undefined && { index },` storing a literal keyval key as a property of a value is *crazy*.

On `decoratedRoomsKey`, first this would go away entirely with the zset solution. But also derived data should live in scratch instead of db. Then on a service restart you derive the data again from the source of truth.

bastianh:
38886dde (catalog, packs, previews) — _id → id throughout the catalog: DecorationTheme/DecorationDefinition, the pack schema, pack.json, the README's pack docs, and every error message. The pack format no longer treats the Mongo-ism as canonical.

02b39d57 (own, place and wear) — the storage rework. Room placements now live in one zset per shard (model.ts:47), scored by parseRoomNameToId, member userId/itemId (composite because under grantAll the item id is the definition id, which users share). zRange at a room's score answers "what's here", zScore answers "where is it" — so the stored index field, the room hash field, and decoratedRoomsKey are all gone. deactivate asks the data structures instead of a remembered key. activate is split into typed entry points — activateInRoom, activateCreep, activateBadge — that take parsed props, never a raw payload; and after each write a settleConflicts sweep re-reads the index and resolves same-user conflicts in itemId order, so racing activations converge deterministically (the pre-check error for the client stays). The commit's tests carry an 11-line dispatch helper that the next slice replaces with the real backend glue.

f9be725b (serve) — backend.ts is now the only file that speaks _id: it strips it from the activation payload, dispatches on the definition's type, and puts it back on everything served (placementToWire, toClientDefinition, the themes route). listDecoratedRooms is one zRangeWithScores read and is now exact rather than a grows-forever superset, so the map test asserting a stale entry flipped accordingly.


it was suggested the zset member could be the bare itemId — it has to stay userId/itemId because grantAll items share ids across users, and the room query needs all users' members anyway.

packages/xxscreeps/mods/meta/decorations/backend.ts
```diff
@@ -0,0 +1,330 @@
...
+		if (userId === undefined) {
+			return { ok: 1, list: [] };
+		}
+		const items = await listForUser(context.db, userId);
+		return {
+			ok: 1,
+			list: items.map(item => ({
+				_id: item.id,
+				createdAt: new Date(item.createdAt).toISOString(),
+				...item.activatedAt !== undefined && { activatedAt: new Date(item.activatedAt).toISOString() },
```
laverdet: Isn't this just `activatedAt: item.activatedAt && new Date(item.activatedAt).toISOString()`

packages/xxscreeps/mods/meta/decorations/backend.ts
```diff
@@ -0,0 +1,330 @@
...
+	}),
+});
+
+hooks.register('roomSocket', async (shard, userId, roomName) => {
+	// Re-read only once something in the room changed. Creep decorations show up in every room, so
+	// this watches their channel too.
+	let stale = true;
+	const markStale = () => { stale = true; };
+	const [ unlistenRoom, unlistenGlobal ] = await Promise.all([
+		getRoomDecorationChannel(shard.db, shard.name, roomName).listen(markStale),
```
laverdet, line 356: For parallel resource acquisition like this you can use `acquireWith`. If one of these throws then the other is abandoned. So you'd acquire them into a `DisposableStack` (with defer) and then the returned effect becomes simply `disposableToEffect(disposable.move())`.

packages/xxscreeps/mods/meta/decorations/backend.ts
```diff
@@ -0,0 +1,330 @@
...
+});
+
+/** Whether any loaded decoration can stand in a room at all. A badge-only catalog leaves the map
+ * with nothing to show, so the per-room index reads below are skipped outright. */
+const hasRoomDecorations = Fn.some(catalog.definitions.values(), definition => isPlacedInRoom(definition.type));
+
+// Creep decorations are deliberately absent: they belong to a creep rather than to a room, so there
+// is no room for the map to draw them in.
+hooks.register('mapStats', async (context, { rooms, response, userIds }) => {
+	if (!hasRoomDecorations) {
```
laverdet: You could conditionally register the hook instead of checking for each invocation.

packages/xxscreeps/mods/meta/decorations/backend.ts
```diff
@@ -0,0 +1,330 @@
...
+		return;
+	}
+	// One read of the shard's zset tells which of the requested rooms hold anything — almost always
+	// few to none — instead of asking after every room one by one.
+	const decorated = await listDecoratedRooms(context.db, context.shard.name);
```
laverdet: I don't think this optimization make sense anymore. `listDecoratedRooms` fetches the entire zset and then `listForRoom` immediately turns around and requests the same information to get the decoration members.

I think the right pattern here would be a watcher which keeps the whole collection in the backend's memory by listening to the channels. I don't think you need to do that now though, just do the naive thing.

packages/xxscreeps/mods/meta/decorations/catalog.ts
```diff
@@ -0,0 +1,466 @@
...
+	theme: string;
+	/** 1–5. Drives the colour of the client's rarity indicator. */
+	rarity?: number;
+	groupDescription?: string;
+	props: Record<string, DecorationProp>;
+	layout?: DecorationLayout;
+	graphics?: DecorationGraphic[];
+	preview?: DecorationPreview;
+	/** Wall overlay texture, `wallLandscape` / `landscape` only. */
+	foregroundUrl?: string;
```
laverdet, line 106: It's an URL? Wouldn't this actually be a relative path fragment to a bundled asset?

packages/xxscreeps/mods/meta/decorations/backend.ts
```diff
@@ -0,0 +1,330 @@
...
+				}(),
+			};
+			assetCache.set(key, entry);
+			return entry;
+		}();
+		if (asset === undefined) {
+			return;
+		}
+		context.set('Cache-Control', 'public');
+		context.set('ETag', asset.etag);
```
laverdet: etag is actually the worst of the cache headers. We use it for some other assets because we don't control the URL, but in this case we do. So we should construct the URL with a bust on file modification time and send immutable cache headers.

packages/xxscreeps/mods/meta/decorations/catalog.ts
```diff
@@ -0,0 +1,466 @@
...
+		}
+		if (assetContentType(file.pathname) === undefined) {
+			throw new Error(`Asset '${value}' of decoration pack '${pack.name}' has an unsupported file type`);
+		}
+		try {
+			await fs.stat(file);
+		} catch (cause) {
+			throw new Error(`Asset '${value}' of decoration pack '${pack.name}' does not exist`, { cause });
+		}
+		const key = `${pack.name}/${decodeURIComponent(file.href.slice(directory.href.length))}`;
```
laverdet: We have `makeRelativeFragment` for this. Is the `decodeURIComponent` doing anything for us, if we end up passing it back into another URL constructor?

packages/xxscreeps/mods/meta/decorations/catalog.ts
```diff
@@ -0,0 +1,466 @@
...
+		...preview.original !== undefined && { original: await resolveAsset(preview.original) },
+		...preview['128x128'] !== undefined && { '128x128': await resolveAsset(preview['128x128']) },
+		...preview['256x256'] !== undefined && { '256x256': await resolveAsset(preview['256x256']) },
+	});
+
+	const definitions = await Fn.mapAwait(pack.decorations, async (definition): Promise<DecorationDefinition> => {
+		// The client and the generated previews both read these straight out as colours, so the
+		// format is checked here rather than wherever one of them trips over it.
+		for (const [ name, prop ] of Object.entries(definition.props)) {
+			if (prop.type === 'color' && prop.default !== undefined && !/^#[0-9a-fA-F]{6}$/.test(String(prop.default))) {
```
laverdet: This seems like validation that ajv should be doing for us. The prop declaration would be an `anyOf`.

packages/xxscreeps/mods/meta/decorations/pack/pack.json
```diff
@@ -0,0 +1,210 @@
...
+		"art/crate.svg": "<svg xmlns=\"http://www.w3.org/2000/svg\" viewBox=\"0 0 64 64\" width=\"64\" height=\"64\"><rect x=\"2\" y=\"6\" width=\"60\" height=\"52\" rx=\"3\" fill=\"#6b5334\" stroke=\"#3a2c1c\" stroke-width=\"4\"/><path d=\"M4 8 L60 56 M60 8 L4 56\" stroke=\"#8a6c46\" stroke-width=\"6\"/><path d=\"M2 20 H62 M2 44 H62\" stroke=\"#3a2c1c\" stroke-width=\"4\"/></svg>",
+		"art/crate-preview.svg": "<svg xmlns=\"http://www.w3.org/2000/svg\" viewBox=\"0 0 64 64\" width=\"64\" height=\"64\"><rect width=\"64\" height=\"64\" rx=\"8\" fill=\"#181818\"/><rect x=\"10\" y=\"14\" width=\"44\" height=\"36\" rx=\"2\" fill=\"#6b5334\" stroke=\"#3a2c1c\" stroke-width=\"3\"/><path d=\"M11 15 L53 49 M53 15 L11 49\" stroke=\"#8a6c46\" stroke-width=\"4\"/></svg>"
+	},
+	"decorations": [
+		{
+			"id": "xx-floor-plain",
```
laverdet: Each decoration has a unique id? In that case what you should have is an object (keyed by id), instead of an array of objects. For the author this feels very old school XML.

packages/xxscreeps/mods/meta/decorations/pack/pack.json
```diff
@@ -0,0 +1,210 @@
+{
+	"name": "xxscreeps",
+	"themes": [
+		{ "id": "xx-basic", "name": "Basics", "color": "#8f8f8f" },
+		{ "id": "xx-neon", "name": "Neon", "color": "#ff4fd8" }
+	],
+	"assets": {
+		"art/tag.svg": "<svg xmlns=\"http://www.w3.org/2000/svg\" viewBox=\"0 0 64 64\" width=\"64\" height=\"64\"><g fill=\"none\" stroke=\"#ffffff\" stroke-linecap=\"round\"><circle cx=\"32\" cy=\"32\" r=\"26\" stroke-width=\"5\"/><path d=\"M22 22 L42 42 M42 22 L22 42\" stroke-width=\"7\"/></g></svg>",
```
laverdet: These might as well just be files instead of payloads copy and pasted into a JSON blob.

packages/xxscreeps/mods/meta/decorations/pack/pack.json
```diff
@@ -0,0 +1,210 @@
+{
+	"name": "xxscreeps",
```
laverdet: Also this should probably just be yaml, which is the xxscreeps configuration language. Writing JSON of this size by hand is a big pain.

packages/xxscreeps/mods/meta/decorations/pack/pack.json
```diff
@@ -0,0 +1,210 @@
...
+	"themes": [
+		{ "id": "xx-basic", "name": "Basics", "color": "#8f8f8f" },
```
laverdet: Same thing here.. this is just an object shape wearing a tophat.

packages/xxscreeps/mods/meta/decorations/catalog.ts
```diff
@@ -0,0 +1,466 @@
...
+		washes.map(({ key, body }) => [ key, { kind: 'generated', body } ]));
+	const packNames = new Set<string>();
+
+	for (const source of sources) {
+		const pack = await loadPack(source);
+		if (packNames.has(pack.name)) {
+			throw new Error(`Duplicate decoration pack name '${pack.name}'`);
+		}
+		packNames.add(pack.name);
+		for (const theme of pack.themes) {
```
laverdet, line 426: You wouldn't need to check for duplicates if the declaration shape is an object.

packages/xxscreeps/mods/meta/decorations/config.ts
```diff
@@ -0,0 +1,45 @@
+export interface DecorationsSettings {
+	/**
+	 * Prefix put in front of decoration asset urls, which are otherwise rooted at `/`. Needed when
+	 * the backend does not sit at the root of the origin the client is served from: an origin of its
+	 * own, e.g. "https://screeps.example.com", or the path a proxy mounts it under, e.g.
+	 * "/(http://localhost:21025)" for the steamless client. Prepended verbatim, so it takes either.
+	 */
+	assetBaseUrl?: string;
```
laverdet: Does the client not accept relative URLs for the assets? If not, this should be a core backend configuration since its relevant to other endpoint. But I think in most cases we can use relative URLs.

packages/xxscreeps/mods/meta/decorations/index.ts
```diff
@@ -0,0 +1,9 @@
+import type { Manifest } from 'xxscreeps/config/mods.js';
+import * as types from 'xxscreeps/tsroot.js';
```
laverdet: This is only for mods which declare player-sandbox visible APIs.

packages/xxscreeps/mods/meta/decorations/model.ts
```diff
@@ -0,0 +1,433 @@
...
+
+export const grantAll = () => config.decorations?.grantAll ?? true;
+
+/**
+ * Announces that what is placed in a room changed, so open room sockets re-read it. Creep
+ * decorations show up in every room, so they get their own channel that all of them watch.
+ */
+const decorationChannel = (db: Database, name: string) => new Channel<DecorationUpdate>(db.pubsub, name);
+const roomChannelName = (shardName: string, room: string) => `decorations/${shardName}/${room}`;
+export const getRoomDecorationChannel = (db: Database, shardName: string, room: string) =>
```
laverdet, line 56: Is there a reason not to just have two functions returning `new Channel`? Instead of 4 functions doing whatever this is.

packages/xxscreeps/mods/meta/decorations/placement.ts
```diff
@@ -0,0 +1,187 @@
...
+/** Whether a placement is visible on the world map, which the `world` property decides. */
+export const isOnWorldMap = (placement: Placement) => placement.props.world === true;
+
+/** Property values share the placement's hash with its target, so they carry a prefix of their own. */
+const propFieldPrefix = 'prop/';
+
+/** Hash fields hold strings; the definition tells {@link decodeProps} what each one was. */
+export const encodeProps = (props: Record<string, PropValue>): Record<string, string> =>
+	Object.fromEntries(Object.entries(props).map(([ name, value ]) =>
+		[ `${propFieldPrefix}${name}`, typeof value === 'boolean' ? value ? '1' : '0' : String(value) ]));
```
laverdet: `String(typeof value === 'boolean' ? Number(value) : value)`

packages/xxscreeps/mods/meta/decorations/placement.ts
```diff
@@ -0,0 +1,187 @@
...
+
+const decodeProp = (prop: DecorationProp, value: string): PropValue =>
+	prop.type === 'boolean' ? value === '1' : prop.type === 'range' ? Number(value) : value;
```
laverdet: `switch (prop.type) { ... }`

packages/xxscreeps/mods/meta/decorations/model.ts
```diff
@@ -0,0 +1,433 @@
...
+	if (!await shard.data.sIsMember('rooms', room)) {
+		return { error: 'unknown room' };
+	} else if ((config.decorations?.requireRoomOwnership ?? true) && !await controlsRoom(shard, userId, room)) {
+		return { error: 'room not controlled' };
+	} else if (blocked(await listForRoom(db, shard.name, room), userId, itemId, definition)) {
+		return { error: 'already decorated' };
+	}
+	// Moving an item out of its old room has to happen before the new placement is indexed,
+	// otherwise a move within one room would drop the entry it just wrote.
+	await deactivate(db, userId, [ itemId ]);
```
laverdet, line 302: Is the comment about this true? Writing a decoration back into its own room should just be a no-op zset score touch right?

packages/xxscreeps/mods/meta/decorations/model.ts
```diff
@@ -0,0 +1,433 @@
...
+export async function deactivate(db: Database, userId: string, itemIds: Iterable<string>) {
+	await Fn.mapAwait(itemIds, async itemId => {
+		const fields = await db.data.hGetAll(activeKey(userId, itemId));
+		if (fields.activatedAt === undefined) {
+			return;
+		}
+		const member = indexMember(userId, itemId);
+		const channel = await async function() {
+			if (fields.shard !== undefined) {
+				const score = await db.data.zScore(shardIndexKey(fields.shard), member);
```
laverdet: `Promise.all` (the zscore will be answered before zrem).

packages/xxscreeps/mods/meta/decorations/model.ts
```diff
@@ -0,0 +1,433 @@
...
+ */
+export async function deactivate(db: Database, userId: string, itemIds: Iterable<string>) {
+	await Fn.mapAwait(itemIds, async itemId => {
+		const fields = await db.data.hGetAll(activeKey(userId, itemId));
+		if (fields.activatedAt === undefined) {
+			return;
+		}
+		const member = indexMember(userId, itemId);
+		const channel = await async function() {
+			if (fields.shard !== undefined) {
```
laverdet: This is a sneaky negated if (no-negated-condition). Anyway, for instances where the if & else are very different cases (like what is here) I prefer the else to be spelled out instead of using the early return.

packages/xxscreeps/mods/meta/decorations/model.ts
```diff
@@ -0,0 +1,433 @@
...
+				const score = await db.data.zScore(shardIndexKey(fields.shard), member);
+				await db.data.zRem(shardIndexKey(fields.shard), [ member ]);
+				return score === null ? undefined : roomChannelName(fields.shard, makeRoomNameFromId(score));
+			}
+			return await db.data.sRem(globalIndexKey, [ member ]) > 0 ? globalIndexKey : undefined;
```
laverdet: await is not needed

packages/xxscreeps/mods/meta/decorations/placement.ts
```diff
@@ -0,0 +1,187 @@
...
+ */
+function parseProp(name: string, prop: DecorationProp, value: unknown): ParsedProp | PlacementError {
+	switch (prop.type) {
+		case 'boolean': {
+			if (typeof value === 'boolean') {
+				return { value };
+			}
+			return value === 'true' || value === '1' ? { value: true } :
+				value === 'false' || value === '0' ? { value: false } :
+				{ error: `'${name}' is not a boolean` };
```
laverdet: We have reinvented exceptions. This is very golang, and also an artifact of the weird always 200 shape the client expects. I think we need a `throw new ClientError('error message');` for the backend to use. That's out of scope for now I'm just thinking aloud.

packages/xxscreeps/mods/meta/decorations/placement.ts
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ * The `active` payload of an activation request, checked against what the definition declares.
+ * Properties the client left out fall back to the definition's seed, so a placement always carries
+ * a complete set and later readers never have to consult the defaults again. Wire-only baggage —
+ * the `_id` the client sends back with an edited placement — is the backend's to strip before the
+ * payload gets here.
+ */
+export function parsePlacement(definition: DecorationDefinition, active: Record<string, unknown>): Placement | PlacementError {
+	const { shard, room, ...rest } = active;
```
laverdet: Can this live in backend.ts or does that mess up your tests? I don't like that this file is a mix of encoding for keyval and also for the client, it's confusing.

packages/xxscreeps/mods/meta/decorations/placement.ts
```diff
@@ -0,0 +1,187 @@
...
+	const props: Record<string, PropValue> = {};
+	for (const [ name, value ] of Object.entries(rest)) {
+		// `Object.hasOwn`: a request naming a prototype member — `toString`, `constructor` — would
+		// otherwise find Object.prototype's functions where a descriptor is expected.
```
laverdet: You could iterate over `definition.props` (trusted) instead of `rest` (untrusted) and you wouldn't need `Object.hasOwn`. It would also make the default value selection easier, which right now is its own loop.

laverdet (changes requested):

packages/xxscreeps/mods/meta/decorations/catalog.ts
bastianh: - As authored in the pack, foregroundUrl (and graphics[].url, preview.*, resources) is indeed usually a relative path fragment naming a bundled asset — resolveAsset at catalog.ts:320 checks it against the pack directory. It can also be an external URL or data URL (the isExternalUrl branch).
- After `loadPack`, the same field on the same interface always holds the resolved public URL that gets sent to the client verbatim.

The name itself is not freely choosable: `foregroundUrl/floorForegroundUrl/graphics[].url` are the official client's wire field names, and `toClientDefinition` spreads them straight through. So renaming the resolved side is off the table.

I'll look into optimizing it when I come to your later comments about the schema rework

packages/xxscreeps/mods/meta/decorations/catalog.ts
bastianh: The asset never goes back into a URL constructor on the serving side: the route is `/assets/decorations/:asset(.*)`, and koa-router 10 decodes the captured param with `safeDecodeURIComponent` before handing it over (layer.js:74). So `context.params.asset` arrives in decoded space, and the catalog key must live there too or a filename needing encoding (space, umlaut) would never match its own request: encoded key `art/my%20file.png` vs decoded param `art/my file.png`. The decode is load-bearing; dropping it would break exactly the filenames it exists for.

I'll make it:
```js
const key = `${pack.name}/${decodeURIComponent(makeRelativeFragment(directory, file).slice(2))}`;
```

packages/xxscreeps/mods/meta/decorations/config.ts
bastianh: the relative urls is the default when the config is not set.

I had problems with testing with the steamless-client .. and the official server uses s3 to actually serve the asset files from aws.. that was the first reason I made it configurable

packages/xxscreeps/mods/meta/decorations/model.ts
bastianh: Yes .. I extend the comment that it makes more clear why it is necessary.

the deactivate is for the cross-shard case, the old room's channel, and the hash del — and those have to happen before the re-index

packages/xxscreeps/mods/meta/decorations/placement.ts
bastianh: Happy to have claude look over it (in a seperate task) .)

bastianh:
Thanks for that extensive review! We went through all your comments/suggestions and  shaped it into 3 new commits to not rewrite what you already reviewed and keep the github-comments at the correct spots:

- pack format: yaml + object-keyed themes/decorations + file assets + export-map resolution (catalog.ts pack half, pack.yaml, art/, package.json, README pack docs, the pack-shape test changes)
- model/storage cleanups: channels, deactivate rewrite, listDecoratedRooms removal, comment fix (model.ts, its tests)
- backend/serving: cache-busted immutable assets, acquireWith, conditional mapStats, parsePlacement move + trusted iteration, backend.assetBaseUrl in core config (backend.ts, placement.ts, config.ts, config.schema.json)

bastianh:
hmm.. not sure why this test is failing .) should not be related to this pr

packages/xxscreeps/mods/meta/decorations/catalog.ts
laverdet: > It can also be an external URL or data URL

We shouldn't support external URLs in this manner. Usually for something like this you'll have a build process that uploads your assets to a bucket which you control and then rewrite the assets in the app.

laverdet:
If you want to break these out into separate PRs I can start a final review and merging. I could also do it on my end without a PR but GitHub is usually better about author attribution when there is an associated PR.

packages/xxscreeps/mods/meta/decorations/catalog.ts
bastianh: The client accepts everything for this field which can be used as an image src. On the official server this is just the aws cdn url. it uses crossorigin="anonymous", so it needs cors header to work.

I'll added a commit to remove our special treatment: 56f9bb27

bastianh:
> If you want to break these out into separate PRs I can start a final review and merging. I could also do it on my end without a PR but GitHub is usually better about author attribution when there is an associated PR.

Thanks I will do .) Most credits probably should go to claude and you corrected it more then me... at least in the end .. it was also some trial and error in the beginning ;) Especially when it crashed the renderer.

laverdet:
#365 #366

PR #366, opened 2026-08-08
decorations: ownership, placement and the backend routes

bastianh:
The rest of the decorations mod, on top of #365. Two commits, reviewable one at a time — I kept them separate because the first is pure account state and the second is pure wire surface, but they only make sense together: the model has no caller until the routes exist, so splitting them across two PRs would land a feature nothing can reach.

**`decorations: own, place and wear what the catalog offers`** — the account state. `model.ts` stores grants and placements in `db.data`; decorations are not `RoomObject`s, they cost no bytes in a room blob and the processor never touches them. Room placements live in one zset per shard scored by room id, so *where a decoration stands* is the data structure's own answer in both directions — a range at a room's score is what stands there, a member's score is where it stands. A room's channel fires when its placements change and one global channel carries the creep decorations, so an idle room costs nothing per tick. `placement.ts` holds the storage half: the props codec and which definitions collide. Each kind of placement — room, creep, badge — has a typed entry point, racing conflicts settle deterministically in itemId order, and deactivation drops a placement the same way everywhere, stranded ones included.

Placing checks room ownership through the controller model (#362). A badge decoration is worn rather than placed and answers the badge symbols hook (#364). `manage decoration` hands out grants when `grantAll` is off — with it on (the default) everybody owns the whole catalog and no grant is read at all, which is the natural setup for a private server.

**`decorations: serve the catalog, the inventory and what is placed`** — the routes. Assets are served immutable; the urls already carry a content cache-bust, so a browser may keep them forever. The inventory and themes routes replace the core stubs, and `activate` parses the request apart right here — the client's `_id` spelling is wire shape and nothing else, so it is stripped from what comes in and put back on what goes out, and it lives in this one file. The room socket and the world map re-read a room's placements when its channel fires. The client learns the whole feature from the version endpoint's feature flags and menu entries (#363).

Both commits build, pass the suite and lint on their own (623 and 645 tests). I also carried your catalog-slice edits forward into these files: `color` rather than `colour` throughout, and the one inline object type is now a named interface — `PlacementTarget`, which `Placement` extends, since it was the location half of that interface spelled out a second time.

### DIFF
--- a/.changeset/decorations-backend.md
+++ b/.changeset/decorations-backend.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Decorations reach the client: backend routes serve the catalog, the inventory and what is placed, pack assets are served immutable, and the room socket and world map pick up placement changes as they happen.
--- a/.changeset/decorations-model.md
+++ b/.changeset/decorations-model.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Decorations gain their account state: per-user grants and typed placement of what the catalog offers — in a room, on a creep, or worn as a badge — handed out with `manage decoration` when `grantAll` is off.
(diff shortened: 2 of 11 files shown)


laverdet:
Thanks. In the future please keep separate commits to two PRs. It's fine if you chain PRs on one and other, I can work through that. Otherwise I have to amend your commits and force push which makes attribution a lot harder to follow in github.

PR #367, opened 2026-08-13
email: hold an address pending confirmation, and confirm it from a mailed link

bastianh:
Sorry about the last two commits in decorations being in one pr.

This one is about email verification and update email, showing the "dirty email" in the client and stuff like that. Not sure what you have planed for email but I guess the sending would be done in a seperate mod? I guess verification still belongs into the server itself.

Again this is devided into several commits that can be pr seperatly, here in one pr as an overview.

An address typed at registration was believed on the spot: `create` associated the `email` provider
outright, so anyone could sign up under an address they don't own and take it away from whoever
does. This adds the other half — hold the address, mail a link, promote it when the user opens one —
behind `backend.autoVerifyEmail`, which stays on by default so a server with no way to send mail
keeps working exactly as it does today.

Two pieces are worth orienting on. Confirmation links are signed with the same key as session
tokens, so they survive a restart and work across replicas without shared storage; a nul-separated
purpose tag keeps the two kinds from ever being exchanged for one another, since no login payload
can contain one. And delivery is a `mailer` provider with room for exactly one implementation —
every server reaches its users differently — so xxscreeps mints and honours the link and a mod
carries it.

Turning `autoVerifyEmail` off needs `backend.publicUrl` set and a mod which sends mail; the backend
says so at startup if either is missing. Links are rooted at `publicUrl` rather than at the request
origin on purpose: that origin is the `Host` header, and a forged one would have us mail somebody a
genuine-looking link pointing elsewhere.

This is the whole chain as one PR to read in context. Every commit builds and passes the suite on
its own, so I can submit them separately instead — three of them (`report the account email`, `sign
tokens for purposes`, `hold an address pending`) don't depend on each other at all. Say which you'd
rather review.

Two things I left alone. `create` is still not transactional (its `// TODO: multi / exec` predates
this), so a lost race on the email index throws after the user record exists — the email reverse
index is claimed with `NX` now, which closes the confirmation race but not that one. And an address
held pending is not yet a sign-in identity: until it is confirmed the user signs in by username,
which is noted on the config option but changes what `POST /api/auth/signin` accepts right after
registration when the flag is off.

## Validation

`npm run build` and `npm test` clean on node 24 — 659 tests, 0 failures — at every one of the seven
commits, not just at the tip. `eslint --max-warnings 0` clean across every file touched. New coverage: token round-trip, purpose isolation, expiry and the
refusal that a login token is not a verification link (`backend/auth/test.ts`); pending-then-promote
including two accounts racing for one address (`engine/db/user/test.ts`); and the mailed link end to
end — sent to the pending address, rooted at `publicUrl`, its token decoding back to the user and
address, and opening it promoting them — plus the no-transport, no-`publicUrl` and refusal paths
(`backend/test.ts`).

Not covered: the `/api/auth/email/verify` and `/api/user/email/resend` routes themselves. Everything
they call is tested; the routing, the redirect target and the response shapes are not.

### DIFF
--- a/.changeset/auth-purpose-tokens.md
+++ b/.changeset/auth-purpose-tokens.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Sign tokens for purposes other than login, with their own lifetime, and never authenticate one.
--- a/.changeset/backend-mail.md
+++ b/.changeset/backend-mail.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Mail the confirmation link for a pending address, through a `mailer` a mod registers.
--- a/.changeset/email-verification-link.md
+++ b/.changeset/email-verification-link.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Confirm a pending address from a signed, expiring link; report an unconfirmed one as `emailDirty`.
--- a/.changeset/user-change-email.md
+++ b/.changeset/user-change-email.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Add `POST /api/user/email`, which the client's account page already calls to change an address.
--- a/.changeset/user-email-resend.md
+++ b/.changeset/user-email-resend.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Add `POST /api/user/email/resend`, which mails the confirmation link for a pending address again.
--- a/.changeset/user-info-email.md
+++ b/.changeset/user-info-email.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Report a user's email address on `/api/auth/me` from core, not only from the password mod.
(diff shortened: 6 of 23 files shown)


laverdet:
Before this, it would be a good idea to take a step back and make email a mod like steam and password. Maybe email is a pretty core feature but I really like the way mods break up the increasingly large scope of this project into independent project chunks. For the registration and login forms it's ok to cheat a bit and hardcode the `'email'` provider id instead of having a full-fledged mod hook for that one thing.

misterwise:
Context that landed after you opened this. #373 moved notification delivery behind a registered transport, and your #379 is where the shape is being settled: the proposal there is a hook any number of mods listen on, so a mailer and a chat bridge can both run.

I think these stack rather than compete. `mailer` as a single implementation reads right, since a server sends mail one way, and the many-listener part belongs a layer up where one notification fans out to mail, Discord and stdout. A notify-mail transport would listen on that hook and call your mailer.

Which leaves one thing worth settling before it lands: the drain runs in a cron service and `mods/backend/email` is loaded only by the backend, so a backend-tier mailer can't be reached from there.

Happy to build the notify-mail side onto whatever signature you land on.

PR #368, opened 2026-08-14
scripts: seal the world's outer boundary

bastianh:
A room is generated against whatever neighbours already exist and rolls ordinary exits toward the ones that don't, so the outermost rooms of a finished world open onto nothing — the map draws the exit, `describeExits` reports it, and a creep stepping onto that border tile is handed to a room that does not exist. `sealWorldBoundary` (`xxscreeps seal-world`, `--dry-run` to list) rebuilds each such room with its outward borders walled off; every inward border is inherited from the neighbour that already authored it, so no shared border moves and the room set stands.

Rebuilding re-rolls a room's terrain and objects, so this is a world-authoring step: run it once the world's extent is final and before players are placed — a boundary room holding a player's objects is refused. A rebuilt room takes the loadout of its mod-10 template type, which only shows on a world a bare `generate-room` authored off-template.

The parts that aren't obvious from the diff:

- The seal set is every side facing the void, not just the ones currently open. `buildRoom` re-rolls any border the caller doesn't author, so a side that was already walled off has to be authored again to stay that way — otherwise the pass opens fresh holes and never settles.
- A sealed highway border takes the deep lane mass and a minimum depth instead of the shallow corner one. The lane doesn't continue through it, so the mass has to close the corridor rather than flank an opening.
- `fillUnreachable` now returns early when no border tile is open. A room walled off on all four sides has no border to reach from, so every open tile read as unreachable and the fill left solid rock where the room's terrain belongs. Reachable today through a re-entered sector hole, not only through this command.
- `isSystemUser` is exported from `engine/processor/model.ts` rather than adding a seventh copy of the `length <= 2` predicate.

## Validation

`tsc -b` and the full suite on `b618e2ce`: 651 tests, 0 failed. `eslint --max-warnings 0 packages` reports only the three pre-existing `unbound-method` warnings in `driver/private/test.ts`, `mods/classic/combat/creep.ts` and `mods/mmo/wallstreet/market.ts`.

`scripts/test.ts` adds six tests: four over the boundary walk — including that an already-walled void side is carried along and that a fully walled room is left alone — and two shard-backed passes. Sealing a two-room world around a sector center keeps every `sectors` / `sectorControl` record intact and leaves nothing for a second pass; a world of one room comes back with ground rather than solid rock. Both shard-backed tests were checked against their production change reverted, and each fails.

### DIFF
--- a/.changeset/seal-world-boundary.md
+++ b/.changeset/seal-world-boundary.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Add `seal-world`: walls off the outer boundary of a generated world.
--- a/README.md
+++ b/README.md
@@ -202,6 +202,15 @@ options apply to the sector's normal rooms. Also available as
 `npx xxscreeps generate-sector W20N20`, taking the same flags as
 `generate-room`.
 
+`sealWorldBoundary` walls off the world's outer boundary, where the outermost
+rooms still carry the exits they rolled toward neighbors that were never
+generated. Each such room is rebuilt with its outward borders walled off and
+its inward ones inherited, which re-rolls its terrain and objects — so run it
+once the world's extent is final and before players are placed; a boundary room
+holding a player's objects is refused. Also available as
+`npx xxscreeps seal-world`, taking the `generate-room` flags plus `--dry-run`
+to list the rooms it would rebuild.
+
 ## Docker
 
 If you want to run xxscreeps in Docker, you can do this:
--- a/packages/xxscreeps/engine/processor/model.ts
+++ b/packages/xxscreeps/engine/processor/model.ts
@@ -234,7 +234,7 @@ export async function roomsDidFinalize(shard: Shard, roomsCount: number, time: n
 	}
 }
 
-const isSystemUser = (userId: string) => userId.length <= 2;
+export const isSystemUser = (userId: string) => userId.length <= 2;
 export async function updateUserRoomRelationships(shard: Shard, room: Room, previous?: ReturnType<typeof flushUsers>) {
 	const checkPlayers = (current: string[], previous?: string[]) => {
 		// Filter out NPCs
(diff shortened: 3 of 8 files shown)


laverdet:
I think it should just be impossible to generate an invalid map with the terrain editor. Doing this as a separate script step seems hacky

bastianh:
Currently the biggest unit the generator creates is a sector with open borders .. so if you want a bigger map you can run for example 
```
xs=(W10 W0 E0 E10)
ys=(N10 N0 S0 S10)

for y in "${ys[@]}"; do
  for x in "${xs[@]}"; do
    docker compose run --rm xxscreeps generate-sector "${x}${y}"
  done
done
```
for a 41x41 map with 1681 room.

I added this script to afterwards seal the edges.

I guess we can also add a script that combines both steps so you can just say generate a 4x4 map or something like that... but we still should have the possibility to regenerate a single sector and then to seal the edges again when it was a sector at the current map edge

laverdet:
What I'm saying is that a map with open borders is fundamentally broken, and it shouldn't be possible to generate one. The engine would maybe even crash if someone walks a creep over there. It's similar to Quake (the game) if you have a hole in your skybox it literally won't even let you save the map because the engine would crash if you tried to load it.

So if you generate a sector and that sector pushes the frontier of the world map then its borders should automatically be closed. If you then later generate a sector next to it then the generation would have the option to punch holes into the neighboring rooms. This process of punching a hole in a room's borders is something that we'll have to code regardless of whether or not we allow temporarily open world borders. So with that in mind I really think it makes sense to enforce closed world borders at each stage of terrain generation.

@misterwise

misterwise:
Yes, I think sector generation with openings that border on non-rooms doesn't make sense and is an oversight. It should be possible to update sector and single room generation to seal, and break the seals pretty easily if it grows later. This will also support a future 'xxscreeps generate' command by easily looping on sector generation. I'm happy to take a look at that.

bastianh:
Ok then I'll close this pr .)

PR #371, opened 2026-08-18
scripts: add an `xxscreeps export` command

misterwise:
Promotes `test/export.ts`, the manual payload writer, into a first-class `xxscreeps export` command in the `generate-sector` command shape: `--shard` selection defaulting to the first configured shard, a usage line, and a room-count summary. The engine is #353's `exportPayload`, unchanged.

## Validation

- Build, eslint, and the full test suite (645 passed).
- Live smoke: `xxscreeps import` (vanilla 121-room world), then `xxscreeps export world.json`. Spot-checked the output: 72 `@`, 141 `E` of which 27 `keeper: true`, 81 `M` with density.
- Round trip: `importPayload` of the exported file, saved into a fresh shard and re-exported, is byte-identical.


laverdet:
Since this is all in place can we drop the dependency on `@screeps/launcher` and point `xxscreeps import` to a payload of the default shard?

misterwise:
Yes, that will be next up!

PR #372, opened 2026-08-19
payload: cover keeper lairs and prebuilt extractors

misterwise:
A controller-less room's prebuilt objects had no codecs, so exporting one silently dropped them: the default shard loses all 32 keeper lairs and 9 extractors, and a generated source-keeper room exports 4 of its 9 objects. Lairs take the `K` marker. The extractor shares its mineral's tile, so it can't own a marker and rides the `M` metadata instead -- which is why `decode` may now return companion objects for one tile; the engine stamps the position on all of them and the id on the first. Only a neutral extractor is carried; a player-built one stays behind like every other owned structure.

## Validation

- Build, eslint, and the full test suite (647 passed), including new round-trip tests in the source and mineral mods.
- Exported the vanilla default shard: 72 `@`, 141 `E` (27 keeper), 81 `M` (9 with an extractor), 32 `K` -- every object except the bot spawns.

### DIFF
--- a/.changeset/payload-prebuilt-objects.md
+++ b/.changeset/payload-prebuilt-objects.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Payloads now carry keeper lairs and the prebuilt neutral extractor.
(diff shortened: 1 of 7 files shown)


misterwise:
One last thing to tidy before implementing import.

PR #375, opened 2026-08-19
decorations: gate the client's `inventory` feature on config

bastianh:
This PR introduces a new inventory switch which allows to disable the the client inventory feature flag. ( This removes the Inventory entry from the menu and applied decorations are not listed in the room view)

This is for example used on the season server. There are decorations configured in config that are automatically applied to every room on the server. This would be a later PR.

Open Question: Would we want to support server wide decorations together with a player inventory ? that might make it more complicated to merge the decorations in rooms... like would they get appended or same object decorations would be replaced ... but some types like graffity or creep decorations also stack. It would probably much simpler if server wide decorations would only work when the user can't set own decorations on top or replace them.
The mixing of server wide decorations with user decorations also has the disadvantage that the client lists the season decoration in its decoration section for every room .. this is only disabled when the inventory is also disabled

It also disables the grantAll default true value that simply grants all decorations to every user ... which was more a debug feature.

btw I created a xxscreeps based server decoration for testing and it looks like this: 

Screenshot 2026-08-19 at 13 33 28

### DIFF
--- a/.changeset/decorations-inventory.md
+++ b/.changeset/decorations-inventory.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Decorations are opt-in: `decorations.grantAll` and the new `decorations.inventory` both default off, so a server renders what is placed without handing anybody the catalog or the client's inventory section. Taking a decoration down stays served either way.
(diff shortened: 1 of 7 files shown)


laverdet:
I don't have a strong opinion on merging default decorations with user ones, or decorations in general. It seems like it wouldn't be tough to merge them, or flag some as internal only but I'm certainly not going to build it :)

PR #376, opened 2026-08-20
scripts: seal the world frontier during generation

misterwise:
Replaces #368. A map with open world borders should be impossible at every stage, so the invariant lives in generation itself rather than a repair command: a border facing a room the world doesn't have always seals, for single rooms and sectors alike. Generating beside a sealed room later punches the border open in place, re-decided at the measured seal rate.

A punch only removes wall. A highway regrows its deterministic wall field under the reopened exits and keeps a tile's wall only where the open room would still have one. Any other room opens the border span two tiles deep (the border tile and the tile behind it, the pair generation holds open for every exit) and carves a slot to its ground for each stretch the seal cuts off. A room whose walls aren't the generated field, such as an imported highway or a custom loadout at a highway coordinate, takes the wall-preserving punch instead; the mode is decided by regenerating the field under the room's current exits. Objects, swamp, and every other tile stay put, so a player-held room is punched like any other and one player at the frontier can't block growth.

Keeps the #368 terrain machinery by @bastianh (sealed highway borders close the corridor with a floored wall mass; a fully sealed room keeps its ground) and drops the seal-world command, which the invariant makes unnecessary.

## Validation

- `npx xxscreeps test`: 651/651, including new scripts/room-gen tests asserting a punch only ever removes wall, punched borders satisfy the exit invariants generation enforces (every open border tile is backed by open ground and reaches the room interior), a room that is not the highway its name claims keeps its terrain, objects and a player's controller survive a punch, and two adjacent sectors reconnect across their shared ring.
- `eslint --max-warnings 0 packages`: clean.

https://claude.ai/code/session_01GUuUkpE7mDp7vBdD6LUbJn

### DIFF
--- a/.changeset/frontier-seal.md
+++ b/.changeset/frontier-seal.md
@@ -1,0 +1,5 @@
+---
+"xxscreeps": patch
+---
+
+Generated rooms wall off borders facing ungenerated rooms; generating the neighbor reopens them.
--- a/README.md
+++ b/README.md
@@ -182,7 +182,12 @@ and a prebuilt extractor. `terrainType` picks one of 28 wall layouts and
 omitted. Like `npx xxscreeps import`, it is an offline operation — stop the
 server before running it so cached world state in the backend, processor, and
 runner workers doesn't go stale. Exits are read from any already-generated
-neighbor, so adjacent generated rooms connect.
+neighbor, so adjacent generated rooms connect. A border facing a room the
+world doesn't have is walled off instead — no room ever opens an exit onto
+nothing — and generating that neighbor later punches the seal open in place,
+re-deciding the border at the live world's rate. A punch only removes wall:
+every other tile, swamp patch, and object stays exactly where it was, so a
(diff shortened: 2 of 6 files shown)


misterwise:
A few things worth your eye before review:

- The new tests stand up local storage to exercise offline tooling, which the repo has so far shipped without tests (import, the earlier room-gen commands). Happy to drop them to keep that line, but the punch invariants (a punch only removes wall, every reopened exit reaches ground) felt worth pinning since they guard player-held rooms.
- The two-sector test generates 231 rooms and adds about 12 seconds to the suite. It is the only test covering the ring reopen across a sector boundary; I can shrink it to a hand-built ring if the cost is too much.
- The punch tests pin spans and terrain through `deterministicRandomForTesting()`, so any change to generation's `Math.random` consumption order will need a seed bump. The gapped-span test is additionally seed-hunted for its precondition and would need re-hunting.

